PR #25, opened 2016-02-08
Added Apache documentation, based on centos 7.1+ documentation

woutervb:
I Added the documentation on how to setup a working Apache configuration based on the centos 7.1 installation documentation.

The 2 problems I currently have with the documentation are:

1. cannot seem to cross-link between the documents e.g. :ref:\`header\` does not seem to work
1. I cannot test a https setup based on letsencrypt, as I have no free public ip adresses for a test host

The config file is copied from a working setup, tested with Firefox, Chrome and IE 11 browsers. No websockets issues problems experienced.

See https://github.com/mattermost/platform/issues/1812 as a reference.


it33:
Asking for community help to verify before merging, thanks for adding this!

https://forum.mattermost.org/t/help-wanted-test-new-apache-instructions/878

MarkMaldaba:
The instructions say ```assume this server has an IP address of "10.10.10.3"``` but then the Apache example is using ```10.10.10.2```.

MarkMaldaba:
(Actually, I notice this is also a discrepancy in the Nginx docs that these are based on.)

woutervb:
@MarkMaldaba I think you missed the fact that a multi-server is described.
On the server with ip 10.10.10.3 apache (or nginx) is installed which in turn will contact the mattermost service installed on a machine with ip 10.10.10.2. The mattermost service will use the database installed on the machine with ip 10.10.10.1.

So the "public" dns to reach the webinterface of mattermost should point to 10.10.10.3.

I hope this clarifies  your concern.

ryanleesipes:
I tested this, both with and without SSL, I received the following error:

mattermost_unreachable

Are there any settings I should set on the Mattermost install to accommodate this set up?

ryanleesipes:
For instance, do rate limits have any bearing on this?

When filled in, vary rate limiting by HTTP header field specified (e.g. when configuring NGINX set to "X-Real-IP", when configuring AmazonELB set to "X-Forwarded-For").

The above does not address an Apache proxy.

MarkMaldaba:
@woutervb - it isn't clear from the instructions that you are talking about a multi-server setup.  In my configuration, everything is on a single server and so all IP addresses were the same.

I think it would be clearer if the docs were written from the point-of-view of the simplest use-case (i.e. a single-server setup) with additional instructions for multi-server configurations.

Or else, be explicit: "This assumes that the system is configured across three separate servers: Apache is installed on a server with IP address 10.10.10.3, Mattermost on 10.10.10.2 and MySQL on 10.10.10.1.  If you are using a single-server setup, then all three IP addresses should be set to the same value."  Or something like that.

woutervb:
@it33 How would you like to see that I proceed. Merge this document in the existing rhel 7.1 document (and try to explain the setup). I guess that my initial approach to create a separate apache document does not work out as well as I wished, as I see that it confuses people about the setup (described in the other document).

MarkMaldaba:
What would make most sense would be to have a single RHEL 7.1 document, with instructions for both Apache and Nginx.  That will remove duplication and make things easier to find.

it33:
Thanks @woutervb, highly appreciated. How about combining Apache with RHEL as you propose? 

At some point we hopefully meet someone who understands Sphinx and can help enable the ability to "include" snippets of documentation in multiple places. Having Apache part of RHEL seems a good first step.

it33:
Also, there is some sort of markdown rendering used in our Sphinx setup and I am not deep on details, that is why Sphinx specific language might not work in files saved as .md. We need to figure out the right format to make things work (that might be just saving as .rst instead of .md, but someone would need to try a few things...)

nmadhok:
Agree with @ryanleesipes and I'm seeing the same issue:

screen shot 2016-02-25 at 5 21 03 am

screen shot 2016-02-25 at 5 21 26 am

colee222:
I am also having the same issue as nmadhok and ryanleesipes.

jurgenhaas:
Me too, on the console of Firefox and Chrome I can see that the Socket requests get a 301 response.

it33:
Thanks @nmadhok , @jurgenhaas , @colee222, appreciate the feedback. Can we have your help each posting [full repro steps to the troubleshooting forum?](https://forum.mattermost.org/t/how-to-use-the-troubleshooting-forum/150)

1. There are probably a lot more Apache users on the forum than in this docs repo. 
2. This is a [common error message on NGNIX](http://docs.mattermost.com/install/troubleshooting.html?highlight=unreachable#please-check-connection-mattermost-unreachable-if-issue-persists-ask-administrator-to-check-websocket-port) with many possible configuration issues, so details on your individual setups is important.

colee222:
Got it! I added the following to the apache2 conf "ProxyPassReverse wss://127.0.0.1/api/v1/websocket"

and re-enabled a2enmod proxy_wstunnel

r2d2leboss:
With this configuration, I'm having 301 for all web socket connections and the message "Please check connection, Mattermost unreachable. If issue persists, ask administrator to check WebSocket port."
More details here: http://forum.mattermost.org/t/problem-with-gitlab-mattermost-on-centos-6-with-scl-apache-2-4/1096/2
Maybe someone already got this issue and found a solution ?

mispdev:
Hi everybody,

I'm using Ubuntu 14.04.4 LTS with Apache 2.4.7 and Let's Encrypt SSL certificates and had similar issues.

I found a virtual host config that works for me - perhaps it also solves the problem for you? Just give it a try. I don't have multiple page loads anymore when switching a channel, and if I post something with Android client it immedeately shows up in web UI (and vice versa) - so websockets seem to work as expected.

In the following config replace `sub.domain.com`, `/opt/mattermost/mattermost/web`, `/etc/letsencrypt/live/sub.domain.com/` (or the like) with whatever fits your setup:
```
<VirtualHost *:443>

		ServerName sub.domain.com

		DocumentRoot /opt/mattermost/mattermost/web

		SSLEngine on
		SSLCertificateFile /etc/letsencrypt/live/sub.domain.com/cert.pem
		SSLCertificateChainFile /etc/letsencrypt/live/sub.domain.com/chain.pem
		SSLCertificateKeyFile /etc/letsencrypt/live/sub.domain.com/privkey.pem

		SSLProtocol             all -SSLv2 -SSLv3
		SSLCipherSuite          ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:kEDH+AESGCM:ECDHE-RSA-AES128-SHA256:$
		SSLHonorCipherOrder     on
		SSLCompression          off

		SSLOptions +StrictRequire

		<IfModule mod_headers.c>
				Header always set Strict-Transport-Security "max-age=15768000; preload"
		</IfModule>

		ProxyPreserveHost On
		ProxyRequests Off

		RewriteEngine On

		# The following two lines fix the problem that page is reloaded multiple times whenever switching the channel
		RequestHeader unset If-Modified-Since
		RequestHeader unset If-None-Match

		RewriteCond %{REQUEST_URI}  ^/api/v1/websocket    [NC,OR]
		RewriteCond %{HTTP:UPGRADE} ^WebSocket$           [NC,OR]
		RewriteCond %{HTTP:CONNECTION} ^Upgrade$          [NC]
		RewriteRule .* ws://127.0.0.1:8065%{REQUEST_URI}  [P,QSA,L]

		RewriteCond %{DOCUMENT_ROOT}/%{REQUEST_FILENAME} !-f
		RewriteRule .* http://127.0.0.1:8065%{REQUEST_URI} [P,QSA,L]
		RequestHeader set X-Forwarded-Proto "https"

		<Location /api/v1/websocket>
				Require all granted
				ProxyPassReverse ws://127.0.0.1:8065/api/v1/websocket
				ProxyPassReverseCookieDomain 127.0.0.1 sub.domain.com
		</Location>

		<Location />
				Require all granted
				ProxyPassReverse http://127.0.0.1:8065/
				ProxyPassReverseCookieDomain 127.0.0.1 sub.domain.com
		</Location>

		#LogLevel debug
		LogFormat "%{X-Forwarded-For}i %l %u %t \"%r\" %>s %b" common_forwarded
		ErrorLog  /var/log/apache2/sub.domain.com_error.log
		CustomLog /var/log/apache2/sub.domain.com_forwarded.log common_forwarded
		CustomLog /var/log/apache2/sub.domain.com_access.log combined env=!dontlog
		CustomLog /var/log/apache2/sub.domain.com.log combined

</VirtualHost>
```

I activated the following Apache modules:
```
	sudo a2enmod proxy
	sudo a2enmod proxy_http
	sudo a2enmod rewrite
	sudo a2enmod ssl
	sudo a2enmod proxy_wstunnel
```

Don't forget to restart Apache after the changes:
``` 
	sudo service apache2 restart
```

Notes:
 - The virtual host config is based on the template already posted here: https://github.com/mattermost/platform/issues/872
 - The config mentioned above seems to have a typo as it shows `ProxyPassReverse ws://127.0.0.1:8065/api/vi/websocket` where it should be `ProxyPassReverse ws://127.0.0.1:8065/api/v1/websocket` (note `vi` vs `v1`).
 - I added the config tweak to avoid multiple loads after channel switch mentioned here: https://forum.mattermost.org/t/solved-apache-proxy-is-reloading-every-page-twice/1276

Hope this helps someone out there or is at least an inspirational base for further experiments. Keep on the good work on Mattermost!

Cheers,
mispdev

esethna:
Closing this PR for now. Sounds like the best option is to combine the Apache with RHEL 7.1 documentation. If anyone is interested please feel free to submit new PR on this. Thanks!

PR #30, opened 2016-02-16
PLT-1799 Add slash commands documentation

jwilander:


esethna:
+1. FYI, I added slash commands to the index.rst

PR #44, opened 2016-02-25
Updating center channel color for themes

asaadmahmoodspin:
Hey @esethna have a look. I just changed the center channel text color to be a bit darker since I think it's a bit light. If you want you can create the screenshots again, even if you don't it's not that big of a deal since the only different is the center channel text color which is now just a bit darker.


asaadmahmoodspin:
Also made the link color a bit darker too.

esethna:
Merged, thanks Asaad

PR #46, opened 2016-02-27
Update account-settings.md regarding username change

esethna:


it33:
Just tested this, documentation doesn't match what's on `master`?

PR #47, opened 2016-02-27
Updated principles to FX Guidelines

it33:
If approved, propose removing the text from http://www.mattermost.org/design-principles/ and adding a redirect here instead


asaadmahmoodspin:
Looks fine.

PR #63, opened 2016-03-07
Updated systemd init daemon to wait for postgresql

raelga:
If postgresql is not up and ready, mattermost fails to start during boot with :  Failed to ping db err:pq: the database system is starting up.

### DIFF
--- a/source/install/prod-rhel-7.md
+++ b/source/install/prod-rhel-7.md
@@ -83,7 +83,7 @@
     ```
     [Unit]
     Description=Mattermost
-    After=syslog.target network.target
+    After=syslog.target network.target postgresql-9.4.service
 
     [Service]
     Type=simple


esethna:
Thanks for the PR @raelga! Adding @crspeller to help review this addition.

crspeller:
Thanks @raelga! Is there any way to make it postgresql version independent? So we don't have to update the docs with every postgresql version.

raelga:
I've looking about use wildcards in the service definition but found nothing. Another way to ensure it runs after postgresql without specifying the version is make the service run after multi-user target but is kind of a overkill. Anyways, I'm not an expert on the new systemd system, maybe some one knows better how to fix this issue that only happens when mattermost is faster than postgresql during the boot process, if the system is fast enough, this dependency issue is not a problem.

crspeller:
Ya we don't need that complexity. Thanks!

PR #109, opened 2016-04-23
Added cli team and admin creation

apheleia:
When setting up a new test machine, the database contains test accounts
after `make test` was run.
Due to that, Mattermost will not display the team creation screen on
`make run`, but rather the login screen.
In order for the developers to log in, they need to create a team and
an admin account first.

This change documents the steps on how to do that using the platform
cli.


apheleia:
As per chat with @coreyhulen in https://pre-release.mattermost.com/core/channels/developers on April 23rd

PR #129, opened 2016-05-05
Updated ESLint links

DavidLu1997:
added eslint link, and removed make check


it33:
@DavidLu1997 could we still list the directory, and add pre-formatting? Easier to read.

PR #135, opened 2016-05-06
Added forking instructions, fixed formatting

DavidLu1997:
Assuming anyone with direct commit access will not realistically need this guide.

@hmhealey please review :)

### DIFF
--- a/source/developer/developer-setup.md
+++ b/source/developer/developer-setup.md
@@ -4,7 +4,7 @@ Developer Machine Setup
 ### Mac OS X ###
 
 1. Download and set up Docker Toolbox
-	1. Follow the instructions at [`http://docs.docker.com/installation/mac/`](http://docs.docker.com/installation/mac/)
+	1. Follow the instructions at [http://docs.docker.com/installation/mac/](http://docs.docker.com/installation/mac/)
 	2. Start a new docker host  
 		`docker-machine create -d virtualbox dev`
 	2. Get the IP address of your docker host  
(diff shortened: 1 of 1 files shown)


it33:
+1 from IT, please pick a dev to review?

source/developer/developer-setup.md
```diff
@@ -4,44 +4,46 @@ Developer Machine Setup
 ### Mac OS X ###
 
 1. Download and set up Docker Toolbox
-	1. Follow the instructions at http://docs.docker.com/installation/mac/
```
hmhealey: You shouldn't need the backquotes around the URL here. Just having `[http://docs.docker.com/installation/mac/](http://docs.docker.com/installation/mac/)` should be fine. Same with the other URLs

source/developer/developer-setup.md
```diff
@@ -4,44 +4,46 @@ Developer Machine Setup
...
+	1. `cd ~/go`  
+	2. `mkdir -p src/github.com/<username>`  
+	3. `cd src/github.com/<username>`  
+	4. `git clone https://github.com/<username>/platform.git`  
+	5. `cd platform`
 
 5. Install godep with `go get github.com/tools/godep`
-5. Run unit tests on Mattermost using `make test` to make sure the installation was successful
-6. If the tests passed, you need to create a team and an admin account. You can choose a team name, email and password:
+6. Run unit tests on Mattermost using `make test` to make sure the installation was successful
```
hmhealey: You don't need to change it back, but just an FYI that Markdown automatically numbers lists like this. It's kind of annoying in a lot of cases, but it makes it easier to have long lists like this

hmhealey:
Just the one small thing with the links, otherwise it looks good

DavidLu1997:
Everything's fixed :)

PR #139, opened 2016-05-06
Added Makefile Documentation

DavidLu1997:
@hmhealey please review :)


source/developer/running-mattermost.md
```diff
@@ -0,0 +1,50 @@
...
+To clean the database:
+  - `make stop`
+  - `make clean-docker`
+  - `make run`
+  
+To run a test SMTP server: `sudo python -m smtpd -n -c DebuggingServer localhost:25`
+
+### Make Commands ###
+
+Here is a complete listing of all `make` commands:
```
hmhealey: I don't think we need to document all of these. Most of them are just there to eliminate some duplicated logic. I think we can remove the lines for setup-run-client-tests, cleanup-run-client-tests, .prebuild, prepare-enterprise, build* (they're all part of `make package`), and setup-mac (that's covered in the dev setup instructions)

hmhealey:
Looks good

PR #146, opened 2016-05-07
Change docker daemon for docker mattermost container

enahum:


it33:
@crspeller when you have time, could you please help review?

PR #168, opened 2016-05-18
Changed <username> to mattermost for go imports

DavidLu1997:
Otherwise, things such as:
`import "github.com/mattermost/platform/model"` fail.

@hmhealey please review


hmhealey:
Looks good

PR #178, opened 2016-05-20
Simplified Upgrading Instructions

DavidLu1997:
Using EDITION instead of team and enterprise to remove duplicated instructions.

@it33 how does this look?


it33:
Way too confusing, propose not doing this.

PR #195, opened 2016-05-29
Minor Changes to Attaching Files Documentation

DavidLu1997:
Changed wording to sound more natural.
Changed the description of "Get Public Link not visible" to not contact the system administrator, as it is usually disabled for a reason.

### DIFF
--- a/source/help/messaging/attaching-files.md
+++ b/source/help/messaging/attaching-files.md
@@ -15,7 +15,7 @@ Mattermost has a built in file previewer that is used to view media, download fi
 
 
 #### Sharing Public Links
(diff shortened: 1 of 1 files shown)


it33:
@DavidLu1997, "it has been disabled through the System Console" this is not necessarily true as it could have been changed via config.json. 

This PR seems more like a subjective stylistic tweak rather than an objective correction to formatting or facts. 

PRs impose a cost on others to review and merge. 

Can we have your help to focus on objective corrections? 

We want to set a tone in our PR cadence that we're working on things that matter most.

PR #198, opened 2016-05-29
PLT-2058 Debugging incoming web hook content

apheleia:
Tell the user that the request body is printed to the console for
easier debugging.

In the previous paragraphs, I only updated the numbering of the ordered list.

See also https://github.com/mattermost/platform/pull/3150

### DIFF
--- a/source/developer/webhooks-incoming.md
+++ b/source/developer/webhooks-incoming.md
@@ -33,8 +33,6 @@ Build Break - Project X - December 12, 2015 - 15:32 GMT +0
 
 ---
 
-Note: If the contents of incoming webhooks are cluttering your debug log, disable it in the `config/config.json` by setting `"EnableWebhookDebugging": false`
-
 ### Enabling Incoming Webhooks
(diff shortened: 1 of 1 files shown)


source/developer/webhooks-incoming.md
```diff
@@ -33,6 +33,8 @@ Build Break - Project X - December 12, 2015 - 15:32 GMT +0
 
 ---
 
+Note: If the contents of incoming webhooks are cluttering your debug log, disable it in the `config/config.json` by setting `"EnableWebhookDebugging": false`
```
esethna: Thanks @apheleia! I think the placement of this note might take a little too high priority in the document. I'd propose removing this note given that you've added similar info to the "Additional Notes" section below?

it33: Hi @apheleia, 

Perhaps it would be easier to add a new "Troubleshooting" section at the bottom of the document and note the Debugging functionality and flag option for webhooks in one place? 

Also, if you start the section we can add in more tips later.

apheleia:
Agreed, will do.

esethna:
Thanks @apheleia!

PR #199, opened 2016-05-29
godep is not required for the latest master

apheleia:
When running the commands with `godep` on the latest master, godep will
print the error `godep: No Godeps found (or in any parent directory)`.

This change explais that the same commands work if the developer omits
`godep` at the beginning of the command.

### DIFF
--- a/source/developer/developer-setup.md
+++ b/source/developer/developer-setup.md
@@ -33,20 +33,17 @@ Developer Machine Setup
 	4. `git clone https://github.com/<username>/platform.git`  
 	5. `cd platform`
 
-5. If you are running Mattermost version 3.0 or older, install godep with `go get github.com/tools/godep`
-6. Run unit tests on Mattermost using `make test` to make sure the installation was successful
-7. If the tests passed, you can run Mattermost using `make run`
-8. You need to create a team and admin account. You can choose a team name, email and password:
+5. Run unit tests on Mattermost using `make test` to make sure the installation was successful
+6. If the tests passed, you can run Mattermost using `make run`
+7. You need to create a team and admin account. You can choose a team name, email and password:
 	- Note: Make sure your team name does not contain any spaces
-    - Note: If you are running Mattermost version 3.0 or older, prepend all commands with `godep `.
-	  E.g. `godep go run mattermost.go  -create_team -team_name="name" -email="user@example.com"`
 	- `go run mattermost.go  -create_team -team_name="name" -email="user@example.com"`
 	- `go run mattermost.go  -create_user -team_name="name" -email="user@example.com" -password="mypassword"`
 	- `go run mattermost.go  -assign_role -team_name="name" -email="user@example.com" -role="system_admin"`
 
-9. Now you can log into Mattermost as the user you created in step 8
-10. You can stop Mattermost using `make stop`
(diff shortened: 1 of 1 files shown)


source/developer/developer-setup.md
```diff
@@ -37,11 +37,12 @@ Developer Machine Setup
 6. Run unit tests on Mattermost using `make test` to make sure the installation was successful
 7. If the tests passed, you need to create a team and an admin account. You can choose a team name, email and password:
 	- Note: Make sure your team name does not contain any spaces.
+    - Note: The latest version of mattermast does not rely on `godep` anymore. If you have the latest version, run the commands by omitting the `godep` at the beginning (e.g. `go run mattermost.go  -create_team -team_name="name" -email="user@example.com"`)
```
esethna: Thanks for catching this @apheleia! Some minor tweaks:

1. I think we want to keep the docs up to date with the latest version of Mattermost. So I'd actually propose removing the `godep` part in each command and then add a note that `godep` is required for versions prior to vx.x

2. There is also a minor typo: "mattermast" -> "Mattermost"

apheleia:
OK, will do. No problem. But which versions is `vx.x`?

esethna:
Thanks @apheleia. Can anyone from the community help review and share feedback on this change?

source/developer/developer-setup.md
```diff
@@ -33,17 +33,20 @@ Developer Machine Setup
 	4. `git clone https://github.com/<username>/platform.git`  
 	5. `cd platform`
 
-5. Install godep with `go get github.com/tools/godep`
+5. If you are running Mattermost version 3.0 or older, install godep with `go get github.com/tools/godep`
```
crspeller: No need to have this extra noise about old versions. This is a developer document. We don't support developing on old versions.

apheleia: OK, then I will remove all mentions of "old version"

source/developer/developer-setup.md
```diff
@@ -33,17 +33,20 @@ Developer Machine Setup
...
-
-8. Now you can run Mattermost using `make run`. Log-in as the admin account you created in step 6
-9. You can stop Mattermost using `make stop`
-10. If you want to setup for cross compilation (required for the `make package` and dependant targets) run:
+7. If the tests passed, you can run Mattermost using `make run`
+8. You need to create a team and admin account. You can choose a team name, email and password:
+	- Note: Make sure your team name does not contain any spaces
+    - Note: If you are running Mattermost version 3.0 or older, prepend all commands with `godep `.
+	  E.g. `godep go run mattermost.go  -create_team -team_name="name" -email="user@example.com"`
+	- `go run mattermost.go  -create_team -team_name="name" -email="user@example.com"`
```
apheleia: See #109 
As per chat with coreyhulen in https://pre-release.mattermost.com/core/channels/developers on April 23rd

apheleia: > When setting up a new test machine, the database contains test accounts
after make test was run.
Due to that, Mattermost will not display the team creation screen on
make run, but rather the login screen.
In order for the developers to log in, they need to create a team and
an admin account first.

esethna:
Thanks @apheleia, thanks @crspeller

PR #200, opened 2016-05-29
Add EE install instructions

it33:
- We started changing Ubuntu instructions into a mix of Team and Enterprise Edition and it got too confusing, so reversing those and adding explicit Enterprise Edition install guides.


it33:
Merging these now to help a customer

PR #201, opened 2016-05-29
Fixing dashes in links

it33:


it33:
Not needed because of : https://github.com/mattermost/docs/pull/202

PR #206, opened 2016-05-31
Adding restore instructions

crspeller:


esethna:
Thanks @crspeller, this is more clear for sure. Merging.

PR #212, opened 2016-06-03
Update sso-ldap.md

lfbrock:


esethna:
Thanks Lindsay. Added a commit to fix system console paths

PR #233, opened 2016-06-13
Moving recognition of contributors to one line

lindy65:


esethna:
Thanks @lindy65, can we please also remove the bullet point, so it would look like:

/platform
[lindy65](), [esethna]() ...

esethna:
Thanks @lindy65 !

PR #248, opened 2016-06-17
Fix wrong gitlab endpoints

pierreozoux:
lost 4h because of that, but happy that nobody will have to lose that time again :)
The rest of the doc was great, so thanks for your hard work!


esethna:
@pierreozoux thanks for the submission! The change looks good, but do you mind resolving the conflicts?

pierreozoux:
it seems fixed in the doc now, so I can close the PR

PR #259, opened 2016-06-28
Updating styles for h4

asaadmahmoodspin:
Have a look.


it33:
+1

Haven't tested but CSS change makes sense

esethna:
+1, thanks I checked with Asaad on the functionality

esethna:
I'll update the system console doc to use this heading style so it looks better

PR #269, opened 2016-06-29
Update to team creation setting

jasonblais:
Team creation enabled for System Administrators, even when the setting is set to `False`


esethna:
+1

PR #303, opened 2016-07-16
Fix formatting

it33:


esethna:
Tested this locally, I like the formatting in a table cell, it looks good on Sphinx and not as noisy as a formatted "note" box. @lfbrock @jasonblais if this looks good I will change the formatting in the whole doc.
image

lfbrock:
+1 looks good

esethna:
Thanks. Merging this in and will submit a PR for updating the whole doc to this format

PR #305, opened 2016-07-18
Note that themes per team is for v3.3

lfbrock:


jasonblais:
+1

esethna:
Do we want to open a new branch for 3.3 instead of having this in master?

lfbrock:
I think it's easier to just put the note in master (since themes is already incorrectly there), and then we can remove the available in v3.3 part in the v3.3 branch

esethna:
Okay, that's fine with me. +1

PR #311, opened 2016-07-20
Fix broken links in config-settings.rst

jasonblais:
I know we're in the process of upgrading this page, but noticed a few broken links


esethna:
+1, thanks Jason, merging this in and will test with the new formatting

PR #328, opened 2016-07-29
Flagged posts

lfbrock:


esethna:
+1, added two minor tweaks

jasonblais:
Will review once the flags are checked in and can confirm final tooltip + position of flags

esethna:
Flag position and tooltip remain unchanged

jasonblais:
Thanks! All looks good

+1

PR #331, opened 2016-08-01
Add 'mark messages read or unread' section

jasonblais:


esethna:
@jasonblais I know this isn't related to composing messaged but it seems like this could go in the messages table rather than its own section?

jasonblais:
2/5 too, but that makes sense.

Had a PR ready to submit for `/shortcuts`, will submit now with the latest change

lindy65:
+1

esethna:
+1, thanks!

PR #348, opened 2016-08-08
Being transparent on what we test for GitLab SSO

it33:
Existing docs are confusing people on what's supported, so need to be clear on what's tested.


lindy65:
+1

esethna:
+1

PR #350, opened 2016-08-10
Renaming CJK search so it's easier to find.

it33:


jasonblais:
Added missing period, otherwise +1

lindy65:
+1

Issue #85, opened 2016-04-02
[Help Wanted] How to turn on Options for Sphinx site


it33:
Can anyone help us turn on the "Options" for exporting to PDF and EPUB? 

Not sure if this requires custom work, help would be appreciated from anyone might be familiar with Sphinx

Example below of what we're looking for [from MongoDB documentation as an example.](https://docs.mongodb.org/manual/) 

![image](https://cloud.githubusercontent.com/assets/177788/14224541/e204c0c4-f854-11e5-9d2c-be43caf55af8.png)

terukizm:
Hi, @it33.

I think the "Options" is just linked by generated PDF and epub files.

In our Makefile, It is possible to generate.

# How to make

Mac OS X

## epub

```
$ make epub

$ open build/epub/Mattermost.epub
```

## PDF(with LaTeX)

Install TeX Live (for Mac):

```
$ brew install cask mactex
```

Open new Terminal, and type this command.
You need to wait for a loooooong time... (I wasted 2 hours)

```
$ sudo tlmgr update --self --all
```

Generate:

```
$ make latexpdf
(press 'r' is error-skip)

$ open build/latex/Mattermost.pdf
```

In other approach, I had tryed to generate PDF with [rst2pdf](https://github.com/rst2pdf/rst2pdf/commits/master).
But, It is not good.
(*.md file is not supported, and there were some problems with the library dependencies.)

It is easy to generate this documents.
But It is difficult to generate "beautiful" documents, I think.

I hope this will help!

shieldsjared:
Looking at the source of the generated page for this, it looks like it's purely custom by the folks at Mongo.  Are you looking for a quick-coolness factor, or are you looking to solve a problem.  If the later, what are the requirements and Im sure I can whip up something identical.

it33:
Hi @terukizm, 

Thanks for your comment, I'm sorry I missed it for some reason. It's very helpful. We'll work on converting the docs to .rst to support PDF based on your advice. 

Hi @shieldsjared, 

The key feature for us was to export to PDF, and to EPUB would be nice-to-have. We weren't aware this UI was custom, we though it was a setting that was flipped on. I'll update the title to emphasize this. 

Our understanding is PDF would only work with Sphinx's RST format, and we've been starting the process of migrating our .md files to .rst. After the PDF export is enabled we have to continue that work. 

Options might be: 

#### 1. Quick fix with link 

[...]

PR #354, opened 2016-08-11
Enabling sphinx options menu

shieldsjared:
Added a versions.html template which renders the options flyout menu on the side navbar.  On the options flyout, included a 'Report a Problem' link and an 'Edit on GitHub' link. Also added (what I believe is) a missing border to the right of the search text input box, missing due to the fixed positioning on the box.  Updated the options reveal to be the 'Mattermost Blue' in lieu of the base themes 'black', since the black didnt jive well w/ the rest of the page. mattermost/docs#85

Would like some input on the color of the flyout... Screenshot of the flyout in action:

optionsflyout


esethna:
Thanks @shieldsjared!! This is awesome, docs is miles beyond where it was before your contributions, much appreciated.

Reviewing I found one bug where the "Edit on Gitbhub" button in the options pane doesn't append the ".md" or ".rst" to the url so it results in a 404 (see screenshot). The "Edit on Github" button on the top right in the header works great.

image

Other than that we are discussing in the UX meeting tomorrow on the color of the options pane. Mattermost blue seems to stand out a lot, we might want a color that is less prominent. I will circle back shortly!

shieldsjared:
Yikes!  I'll see if I can figure that out. Seems like Sphinx puts a lot of information into the available context but I'm still trying to figure it all out. Worst case resolution would be to _not_ show the edit link whenever all variables needed to construct the URL aren't present. I have an idea of what the issue might be though :)

esethna:
Thanks @shieldsjared! We chatted in UX and think that maybe something between the default grey and black would be best. Can we try #333333? Open to other suggestions!

shieldsjared:
I'll post full screenshots before committing :) I should have time between flights tomorrow.

shieldsjared:
@esethna - I fixed the issue where the edit link _sometimes_ did not work as expected (I mirrored the error checking logic baked into the breadcrumbs template provided by the rtd theme).  Also, I updated the options theme to #333333 but then found that it was a such a close match to the body text color being used (#404040), so for consistency I made it #404040 to make it an exact match.  

optionsflyout 2

Also, I snuck in something (I can remove it if required): the border to the right of the search-box is being hidden because of the box-sizing and fixed positioning, so I added a right-border to the necessary element so that the right-border on the LH bar is consistent from top-to-bottom.

esethna:
Much appreciated, the changes look excellent! I will pass on for dev review and then get this merged!

enahum:
+1

PR #379, opened 2016-08-18
Adding note on HIPPA compliance

it33:


it33:
Answering customer question, PM please review even after the merge

Issue #368, opened 2016-08-16
[Help Wanted] Momentum scroll on mobile devices


esethna:
Navigate to docs.mattermost.com on a mobile device and notice that the website scroll is fixed to your finger movement.

Expected: momentum scroll observed on most other websites.

shieldsjared:
If somebody doesn't get to this before me, I'll knock it out. Just at a css attribute.

shieldsjared:
PR submitted.

esethna:
Resolved, thanks to @shieldsjared!

PR #381, opened 2016-08-18
Enable smooth-scrolling on (some) touch-enabled devices

shieldsjared:
Added the required css attributes to the body of the page to enable a smooth (aka elastic/momentum) scrolling on mobile devices, as requested by mattermost/docs#368.  Note that this is a feature that is only supported by certain touch-enabled devices/browsers!

### DIFF
--- a/source/theme.css
+++ b/source/theme.css
@@ -158,6 +158,10 @@ h1, h2, .rst-content .toctree-wrapper p.caption, h3, h4, h5, h6, legend {
 	width: 100%;
 	height: 100%;
 	background: transparent;
+	
+	/* May seem out of place, but it enables smooth/momentum/elastic scrolling on mobile devices */
+	overflow-y: scroll; /* has to be scroll, not auto */
+    -webkit-overflow-scrolling: touch;
 }
 
 .wy-side-nav-search input[type=text] {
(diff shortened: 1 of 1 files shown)


esethna:
Excellent, thanks @shieldsjared! Adding @asaadmahmood for dev review.

asaadmahmood:
@shieldsjared it works but it produces a scrollbar on the body, there already is a scrollbar on `.wy-nav-content-wrap`, so now the `.wy-nav-content-wrap` scrollbar appears weird and on the inside.

We also don't need to have `overflow: scroll` on all browser sizes, we can only make it scroll on lower browser widths for touch devices.

What we need to do is, we need to keep the code the same and just add:

.wy-nav-content-wrap {
    overflow-y: scroll;
    -webkit-overflow-scrolling: touch;
}

on possibly just the @media 768 media query.

shieldsjared:
Ah, thanks @asaadmahmood! I'll get it fixed tomorrow.

shieldsjared:
Now that I'm at my computer, I see what you're talking about... I don't know how I missed that!  Great catch!

shieldsjared:
@asaadmahmood - fixed and ready for review.  thanks for the input!

asaadmahmood:
+1

PR #384, opened 2016-08-18
added clarifications for single server install

pmccarthy01:


pmccarthy01:
looks like this could be done for ee-prod-rhel-6.rst too

crspeller:
+1

PR #386, opened 2016-08-19
Fix JS client link

hmhealey:
@esethna I don't know why it pointed to the WebClient before, but that wasn't correct. It should just point to the regular one because the WebClient has some stuff specific to our app in it.


esethna:
Thanks!

PR #391, opened 2016-08-21
UX guideline updates

it33:
1) Proposing "Error Message" > "Issue Message"

Per the design principles, it should be impossible for users to cause "errors" in Mattermost. The UI is not reflecting this (https://mattermost.atlassian.net/browse/PLT-3983). 

Propose we make a DISTINCTION between Error Messages (i.e. JavaScript errors and things that are bugs in our system) vs. Issue Messages, which is feedback to the user on how to adjust their input to get the outcome they want. 

We have a UX Defect right now that if you click into an email notification of a DM someone deleted you get this giant word "Error" that makes it seem like the system has failed. 

image

That giant "Error" word should only be used when the system is actually failing, not when it's operating as intended.

2) Adjusting headings to make it easier to group similar guidelines. 

Hopefully in another PR we can update the ordering to make grouping easier--input guidelines with other input guidelines, etc.

### DIFF
--- a/source/developer/fx-guidelines.rst
+++ b/source/developer/fx-guidelines.rst
@@ -243,24 +243,6 @@ Input: Hints
 
         Example 2: If a user enters invalid uppercase letters and spaces for a URL, show an issue message and also offer a correction, substituting dashes for spaces and lower case letters for uppercase letters, so the user can resubmit immediately with valid input.
 
-**Error messages:**
-
-    The word "error" in the Mattermost user interface should be reserved for only `bugs <https://docs.mattermost.com/process/terminology.html#bug>`_ in Mattermost software, such as encountering a JavaScript error. 
-    
-    Under the fast, obvious, forgiving design principle, it should be impossible for users to make an "error" in Mattermost. If they do something unintended the system should provide an "issue message" (see below). 
-
-
-**Issue messages:**
-
-    Issues messages help a user adjust their actions to produce a desired result. They different from "error messages" (see above) that indicate a `bug <https://docs.mattermost.com/process/terminology.html#bug>`_ with Mattermost software. 
-
-    Issue messages should appear immediately below input fields and offer clear and concise information about why an input cannot be accepted.
-
-    Issue messages should aesthetically appear helpful, and not punishing.
-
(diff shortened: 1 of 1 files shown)


esethna:
+1

lfbrock:
I thought about this some more, and I disagree. 

I think it's fixing the wrong issue. The guideline should be to show a helpful title describing the problem instead of showing the word "Error". 

Changing "error message" to "issue message" in our documentation just seems confusing, where else has anyone heard of an "issue message"? It's not clear what that is. And it really doesn't tell you how to solve the problem of having the "Error" title.

yangchen1:
I think I agree with points brought up by Ian and Lindsay.

I do think that having a giant "Error" misrepresents somewhat that it's the system's fault, but on Lindsay's point, the purpose of the message is to be descriptive and letting the user know what's going on, and this doesn't follow the guidelines.

I see where you are coming from, Ian - "Issue message" could be used to distinguish user slips/mistakes from actual system errors. But from the POV of people reading the document, "issue message" can be confusing, as Lindsay pointed out. I think what we could do instead to distinguish it for the purpose of this PR/docs is maybe say "System Error Messages" or something, and say "Error Messages" for user ones. And definitely mentioning in the docs that instead of actually saying "Error" in the platform, saying some helpful title such as "User does not exist" or something in your example should be the norm. :)

it33:
@lfbrock, @yangchen1 agree with your feedback--I think what I'm trying to say is we should never show the word "Error" to a user unless something unexpected happened with the System. 

In my mind there are: 

- Bad error messages (the ones that say "Error" when the user reaches an unsurprising state (clicking on a link to a message that was deleted). 
- Mediocre error messages (not embarrassing to the design team, but room for improvement). 
- Good error messages (concise, informative, high quality error messages). 

This is the "voice of the product" and we need to come across as a professional product. 

Here are some examples of giving user feedback in error messages from Microsoft and Google: 

- [Windows UI Error Guidelines](https://msdn.microsoft.com/en-us/library/windows/desktop/ms679325(v=vs.85).aspx) 
- [Google UI Error Guidelines](https://material.google.com/patterns/errors.html#) 

We need to knock out something similar. 

Could I ask PM's help to: 

1. Open a Jira ticket for UX/PM to figure this out? 
2. Review this ticket again for the other minor changes and let me know feedback? 

There's a bunch of minor reformatting I did to help make UX guidelines easier to navigate. Also, I've proposed minimum resolution in [another PR](https://github.com/mattermost/docs/pull/443) but it's going to conflict with this one so I added it here.

it33:
It's been a month, feedback shared seems addressed so merging and we can make additional PRs to fix anything missed, 

UX Guidelines probably low risk...

PR #396, opened 2016-08-23
Change 'Amazon EC2 administrator' to 'Amazon AWS administrator'

jwilander:
EC2 is a specific service AWS provides and isn't related to S3 (another service AWS provides).


esethna:
+1 Thanks!

PR #419, opened 2016-08-29
WIP: More mindsets

it33:
- Come on PMs, chime in :)

### DIFF
--- a/source/process/mindsets.rst
+++ b/source/process/mindsets.rst
@@ -28,3 +28,17 @@ When you rush to get something done quickly, it can actually increase the time a
 
 Rushing means a higher chance of missing things that need to be done, and the cost of doing them later is significantly higher because you have to re-create your orignal setup to add on the work. 
 
+Assumption or Emotion 
+~~~~~
+
+There are only two reasons why people disagree, either: 
+
+1. There is an **emotion** in the way (and probing for this typically solves ii)
+2. People have different underlying **assumptions**
+
+When you disagree, try ask questions until you find where your underlying assumptions differ, then: 
+
+a. Try to find a way to replace assumptions with facts
+b. See if you can agree on a decision after understanding each other's assumptions 
+
+People here are generally very smart, and it's worthwhile to understand where they're coming from. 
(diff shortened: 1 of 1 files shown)


lfbrock:
I mean I like the general idea, but I don't agree with saying there are **only** two reasons people disagree, since it seems very simplified/untrue. First of all, I think that's assuming both people have access to the same information set, and that both people don't have any errors in their logic/reasoning.  

If we include those assumptions, then I think there's still something missing around priorities/values. Different people can have different priorities which can then lead to disagreements, and that is not covered by assumptions and emotions.

yangchen1:
I agree with you Lindsay. The way I read Ian's edit though, what you say about access to information set and errors in logic/reasoning is under the 2nd reason of people having different assumptions for me. Priorities/values _could_ be under emotions, but I think it's a stretch. :)

it33:
Thanks @lfbrock, @yangchen1, agree with your feedback and adjustments made. 

Where I've found this mindset useful is to create awareness of emotions caught up in a decision, so they don't persist as hidden variables. I find it useful: 

1. As a self-check to make sure I'm not sub-consciously digging into a position just to be right. 
2. As a reminder no one is ever "wrong". There's just different assumptions (including priorities, values, logic, etc.) and emotions going on. This can turn seemingly adversarial discussions into cooperative discussions. 

I've updated the PR, let me know your thoughts?

it33:
[Adding "Mini-boss, End-boss"](https://github.com/mattermost/docs/pull/419/files#diff-ac3dec54896b0eadd3f2b7c43bbf2393R49)

it33:
Adding "Likes & Wishes"

lfbrock:
Alright, I still disagree with priorities being counted as an assumption because it does not fit the definition of the word in any conceivable way, but it's close enough.

it33:
Hmm... I might be missing something. In my mind, there are facts and assumptions, and priorities can be either, depending on whether they are objective (customer needs X feature in order to deploy) or subjective (quality level of a component needs to be higher). That said, I wouldn't say the way I perceive the world is normal...

it33:
@lfbrock [added priorities to emotion & assumption](https://github.com/mattermost/docs/pull/419/files#diff-ac3dec54896b0eadd3f2b7c43bbf2393R31). 0/5 open to feedback.

it33:
Removing WIP prefix as it seems like we have enough to consider a merge.

lfbrock:
+1

esethna:
+1, closing to resubmit due to conflicts

PR #431, opened 2016-08-31
Manually updating channel handle/URL when renaming a channel

jasonblais:


it33:
>You may also change the channel URL (also known as the channel handle) when renaming a channel. **The URL is not automatically updated to prevent unintentionally breaking existing links and integrations.**

Seems a little expert mode, it might be more important to be clear that changing a channel URL it will break existing links (will it break webhooks and permalinks too?)

I don't think anyone expects the channel URL to change when the UI name changes. It would be like updating the site name and expecting the entire server URL to change.

jasonblais:
Thanks, updated

esethna:
+1

lfbrock:
+1

PR #433, opened 2016-08-31
Delete slashCommandsTable.PNG

yangchen1:

PR #435, opened 2016-08-31
New Slash Commands table.

yangchen1:
Updated to include new slash commands.


esethna:
Thanks Yang, can you please combine the commits to delete the table and add the new one into one pull request? ie make the changes on the same branch?

jasonblais:
Actually, I think it'll override the image, so https://github.com/mattermost/docs/pull/433/ is not needed.

I'm not sure where/how the image is generated, but it would be nice if the shortcuts were in alphabetical order. 

Also, `/shortcuts` is missing from the list.

lfbrock:
Sorry @jason I don't think alphabetical order makes sense, because it would make sense to group some like online/offline/away together, and collapse/expand

esethna:
@yangchen1 @jasonblais, looks like this PR is not replacing the table but adding a new one since the extension is not capitalized. @yangchen1 can you please tweak this branch to delete the old table and add the new one with corrected capitalization.

image

esethna:
+1

lfbrock:
+1

PR #442, opened 2016-09-01
Proposed tweak to "Judgement Day" definition

it33:
So it's a little more neutral, what to include instead of what to cut


lfbrock:
+1

jasonblais:
+1

PR #443, opened 2016-09-01
Proposing minimum display resolution

it33:
- As of Jan 2016, it seems 96% of browsers are 1280 x 800 or larger, 99% display 16.7M colors

image


it33:
Moving this to another PR, since there's a merge conflict.

PR #446, opened 2016-09-02
Fixing quotes

coreyhulen:


esethna:
Thanks Corey, that seems to be the only place in the doc using those quotes.

PR #456, opened 2016-09-06
Documenting a wss issue I encountered with resolution

dun3:
After tripple-checking my nginx installation, I found out that I set the "Site URL" setting instead of leaving it empty. Documenting this here by extending the steps to follow with a warning that the site url setting will impact wss functionality.


shieldsjared:
@dun3 My team actually found that we _had_ to have the setting entered (we're behind a proxy) in order for wss  and email links to work.  The issue we found was that if you have a trailling '/' at the end of the URL, it would cause the error you received.  Given our experience, I'm not sure I'd recommend this statement "Most likely you want to leave that entry blank." in your proposal.  Leaving that entry blank should not be mutually exclusive to the use of web-sockets. (e.g. you get one or the other).

I recently reported a bug about the trailing '/' issue and it was resolved in 3.4.  Check out [the Jira ticket](https://mattermost.atlassian.net/browse/PLT-4047) for details.  Might resolve your issue as well, if you haven't upgraded to 3.4.  I'm curious to see the outcome.

esethna:
Closing this PR due to inactivity. Please mention me here if you'd like it re-opened.

PR #458, opened 2016-09-06
Note for troubleshooting outbound connections

it33:


it33:
Merge to address customer issue

PR #461, opened 2016-09-06
Heroku documentation

lfbrock:
Open to feedback on whether we need to doc the heroku settings, since they are described well in the app install itself (see screenshot for example). I included them, but not sure if it's helpful, or if it would be better to keep the docs short since the goal is a quick + easy install. 0/5
screenshot 2016-09-06 15 56 47

### DIFF
--- a/source/install/heroku.rst
+++ b/source/install/heroku.rst
@@ -1,0 +1,31 @@
+Mattermost Preview using Heroku
+===============================
+
+Quickly set up Mattermost in Preview Mode to explore product functionality using Heroku.
+
+This configuration is not recommended for use in production, as it currently does not support email notifications. Also note that for persistent file storage AmazonS3 must be used, as local storage will be deleted periodically when the Heroku process restarts.
+
+Heroku Installation
+--------------------
+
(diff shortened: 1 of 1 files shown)


source/install/heroku.rst
```diff
@@ -0,0 +1,31 @@
...
+Heroku Installation
+--------------------
+
+1. Sign up for a free account on `Heroku <https://www.heroku.com/>`_
+
+2. Click the |HerokuDeploy|_ button
+
+.. |HerokuDeploy| image:: https://www.herokucdn.com/deploy/button.svg
+.. _HerokuDeploy: https://heroku.com/deploy?template=https://github.com/mattermost/mattermost-heroku
+3. (Optional) Adjust the settings in Heroku
```
esethna: Maybe just mention that these fields autocomplete but if you'd like to customize your setup then you can optionally edit these fields.

esethna:
+1 other than my comment above, I think it's fine to detail the optional settings, the doc is not long

lfbrock:
@esethna, updated the line to be  "(Optional) Customize the default settings in Heroku"

jasonblais:
+1

PR #469, opened 2016-09-09
Update known issues in search docs

jasonblais:
As reported here: https://mattermost.atlassian.net/browse/PLT-4060


yangchen1:
+1

lfbrock:
@jasonblais, can you clarify this a little more? It seems like the issue is that the dash is ignored?

esethna:
+1

PR #474, opened 2016-09-09
Update emoji image

lfbrock:


esethna:
@lfbrock sorry the file extension needs to be capitalized or else github treats it as a different image rather than a replacement of the original

lfbrock:
@esethna thanks, should be fixed now

jasonblais:
+1

PR #492, opened 2016-09-15
Update Zapier Redirect URL for Global Activation

esethna:
Old Redirect URL doesn't work anymore for connecting new accounts since we are going through the global activation process. Note: Old accounts authorized with that URL are not disconnected.


esethna:
Pushing this in.

esethna:
```
test

PR #497, opened 2016-09-16
Fix a typo, and fix title underline lengths.

JeffSchering:
The build throws warnings when the title underline in an .rst file is too short. The title underline should be at least as long as the title text.


shieldsjared:
+1

PR #512, opened 2016-09-20
Add error message guidelines to UX guidelines

yangchen1:
Added error message guidelines and examples. Changed some wording of "issue message" to "error message" because the term "issue message" is not a term that is used and thus it's difficult to find. The guideline still specifically states that the actual message itself shouldn't say "error."

See previous discussion here: #391


lfbrock:
+1, I also added some changes to fix grammar since there were some old sentences in the doc that did not make sense

source/developer/fx-guidelines.rst
```diff
@@ -434,10 +434,47 @@ If what the users need to perform is obvious, we should make concious decisions
 ---------------------------
 
 |
+Feedback: Error Messages
+---------------------------
+
+On occasion, Mattermost users may encounter a problem that prevents them or the Mattermost system from successfully completing a task. The unexpected user behaviour or system response should be communicated to the user through an error message, and should follow the design principles of forgiving and obvious.
+ 
+Error messages should:
```
esethna: There actually needs to be a line break before the first bullet or they dont render properly

image

source/developer/fx-guidelines.rst
```diff
@@ -434,10 +434,47 @@ If what the users need to perform is obvious, we should make concious decisions
...
+- State the encountered problem with the component in the title (e.g. Team URL Not Found).
+- Describe very briefly why the error happened under the title.
+- When possible, have a link to direct users to help complete their original task or return to the previous state.
+- Give recommendations/solution(s) for the next course of action.
+- Be consistent with the Mattermost voice and look professional.
+- Be avoided when possible (by avoiding the error condition).
+
+Error message should not:
```
esethna: Same here

esethna:
+1 other than the formatting fix required, then it can be merged, thanks!

Issue #506, opened 2016-09-19
[Help Wanted] CSS to fix line wraps in titles for left-hand sidebar


it33:
After the [re-org PR](https://github.com/mattermost/docs/pull/502) goes in, propose having someone fix CSS for line wraps in left-hand sidebar.

Right now, formatting is all broken: 

![image](https://cloud.githubusercontent.com/assets/177788/18622436/e2670b66-7de3-11e6-8254-bc9e8b5c708f.png)

esethna:
Closing as resolved, thanks @shieldsjared!

PR #514, opened 2016-09-20
506 - Corrections to Security Policies TOC (fixes LHS)

shieldsjared:
Corrects issues cited in #506 

Removed anchor tags from the end of the headers, and replaced TOC named-anchor links w/ the permalink names generated by Sphinx.  The LHS menu was corrupt because of the inline anchors (HTML) being added to the headers w/in the markdown file.

LHS now looks as it should.

image

### DIFF
--- a/source/process/security.md
+++ b/source/process/security.md
@@ -2,12 +2,12 @@
 
 This document summarizes the internal security policies at Mattermost, Inc. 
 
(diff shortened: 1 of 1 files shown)


it33:
Thanks @shieldsjared!!

PR #529, opened 2016-09-27
Sidebar improvements

asaadmahmood:
PLT-3974 and PLT-3975. To have the collapse working on click on an item in the sidebar and also improving the styling of the items in the sidebar.


it33:
@shieldsjared Not urgent, wondering if you could help review when you had a chance? 

You're probably one of the more knowledgeable people in the community on this.

asaadmahmood:
What the JS does is basically it allows users to toggle the sidebar element even if it is open once a user clicks on it.

asaadmahmood:
Also @shieldsjared this is based on master, and we have a caption currently on master, I'm not sure what you mean there.

shieldsjared:
Hey @asaadmahmood 

The `.caption` class is only added to the LHS table of contents when the `:caption:` option is included in the `:toctree:` directive.  Before the docs were split into guides, the root `index.rst` had captions applied to each `:toctree:` (which influenced the look-and-feel of the LHS).

Commit [253a39b](https://github.com/mattermost/docs/commit/253a39b875af6e2c11bea2feb1eb7b0924c19e0b#diff-17ccd0fa0e8d8212130258dcc1d84d19) refactored the TOC, resulting in the removal of all TOC captions.

Now that the caption directive is no-longer in use (you can search the code repository for "caption" to confirm), the LHS is not generating any elements that include the `.caption` class.  Since there are no elements being generated with the `.caption` class, the jQuery binding `$('.wy-menu-vertical .caption')` will never result in any elements... and therefore the handler is never being called.

asaadmahmood:
Okay I think it's fine, I'll remove that code since it's no longer required.

shieldsjared:
+1 

Yeah I'd remove that whole block. Without seeing the two Jira tickets, I can't comment on the changes applicability...just solely on the code changes in the commit.

esethna:
Thanks for the review @shieldsjared and for the changes @asaadmahmood. Merging this in.

Issue #548, opened 2016-10-01
[Help Wanted] Reorganize and correct Desktop App info table


it33:
https://docs.mattermost.com/help/apps/desktop-guide.html#id2

![image](https://cloud.githubusercontent.com/assets/177788/19017024/98dea934-87e0-11e6-8c80-1eeba41b5555.png)

Propose: 

1. Name the columns Windows, Mac and Linux (rather than having "OS X", which is less obvious)

2. Re-order rows so we have all the Yes, Yes, Yes at the top, then Yes, X, X, then [BLANK], YES, X. so it's easier to see what's in Windows (most popular), 

3. Re-order the settings in the section below so it matches the table. 

4. Update table to match the description below, example: In the description [Start App on Logon](https://docs.mattermost.com/help/apps/desktop-guide.html#start-app-on-login-windows-linux-only) is only available on Windows and Linux, but the table shows it's available on Mac, this doesn't appear to be correct. 

Here is Setting Page on Mac: 
![image](https://cloud.githubusercontent.com/assets/177788/19017048/02f0cbbc-87e2-11e6-93c3-b721cf464ab0.png)

ClemMakesApps:
Looks like this issue should be closed since #549 merged 😄

jasonblais:
Thanks for the note @ClemMakesApps! Closing the issue,

PR #549, opened 2016-10-01
Update desktop guide

jasonblais:
As suggested in #548 

1. Name the columns Windows, Mac and Linux (rather than having "OS X", which is less obvious)
2. Re-order rows so we have all the Yes, Yes, Yes at the top, then Yes, X, X, then [BLANK], YES, X. so it's easier to see what's in Windows (most popular),
3. Re-order the settings in the section below so it matches the table.
4. Fix a typo in table, removing "Start App on Login" from Mac


esethna:
+1

PR #567, opened 2016-10-04
yum + sudo

gabx:
1- yum is now depreciated on RedHat & Fedora. Please use dnf instead
2- use the hash tag command prompt to make the reader know the command has to be run as root.


esethna:
Closing as it is replaced by #600

Issue #489, opened 2016-09-13
[Help Wanted] Create install guide variant with MySQL commands for Ubuntu 16.04, RHEL6 and RHEL7


it33:
While MySQL is more popular than Postgres, the Mattermost install guides default to Postgres. 

This ticket is to create a variant of [the Ubuntu, RHEL6 and RHEL7 install guides](https://docs.mattermost.com/#install-guides) that offer step-by-step instructions for MySQL. 

Please post-fix the names of the new files with `mysql`, so for `prod-rhel-6` please call it `prod-rhel-6-mysql`.

jasonblais:
You can see an example provided in [Ubuntu 14.04 guides](https://docs.mattermost.com/install/prod-ubuntu-1404.html#set-up-mysql-database-server)

JeffSchering:
I've added the MySQL instructions to the Ubuntu 16.04, RHEL 6.6, and RHEL 7.1 install guides. See PR #742

JeffSchering:
The MySQL instructions are included in the new install guides for Ubuntu and RHEL.

PR #573, opened 2016-10-05
Added MySQL database instructions

JeffSchering:
I installed Mattermost on a vanilla Ubuntu 14.04 VM and it works fine. I did not set up the Nginx proxy though. Also, I could not get the Upstart daemon working; I don't know if that's my fault or the instructions. I assume it's a problem on my end.

Also:
- changed from using 'ubuntu' user to creating and using 'mattermost' user. (The user 'unbuntu' does not exist on a vanilla install of 14.04)
- future-proofed by removing 'currently v3.3.0' from step 4 in the Set up Mattermost Server section.
- changed the Mattermost file extract command to 'sudo tar -xvzf *.gz' which is easier to type

### DIFF
--- a/source/install/prod-ubuntu.rst
+++ b/source/install/prod-ubuntu.rst
@@ -135,31 +135,30 @@ Set up MySQL Database Server
 Set up Mattermost Server
 ------------------------
 
-For the purposes of this guide we will assume this server has an IP
-   address of ``10.10.10.2``
+For the purposes of this guide we will assume this server has an IP address of ``10.10.10.2``
 
-2. Create a mattermost user and group
+1. Create a mattermost user and group
 
     - ``sudo adduser --system --group mattermost``
 
-3. Change to the mattermost home directory.
+2. Change to the mattermost home directory.
 
     ``cd /home/mattermost``
 
-4. Download `any version of the Mattermost Server <https://docs.mattermost.com/administration/upgrade.html#version-archive>`_ by typing:
+3. Download `the latest version of the Mattermost Server <https://docs.mattermost.com/administration/upgrade.html#version-archive>`_ by typing:
 
    -  ``wget https://releases.mattermost.com/X.X.X/mattermost-X.X.X-linux-amd64.tar.gz``
    -  Where ``vX.X.X`` is the latest version.
 
-5. Extract the Mattermost Server files by typing:
+4. Extract the Mattermost Server files by typing:
 
    -  ``sudo tar -xvzf *.gz``
 
-6. Change the user and group of the extracted files to mattermost
+5. Change the user and group of the extracted files to mattermost
 
    - ``sudo chown -R mattermost:mattermost mattermost/``
 
-7. Create the storage directory for files. We assume you will have
+6. Create the storage directory for files. We assume you will have
    attached a large drive for storage of images and files. For this
    setup we will assume the directory is located at
    ``/mattermost/data``.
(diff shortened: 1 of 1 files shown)


shieldsjared:
#489

JeffSchering:
Thanks @shieldsjared! I wasn't aware of that issue when I added in the MySQL instructions. From that issue, it looks like separate docs are being requested; one for PostgreSQL and one for MySQL. That would make two very long docs almost identical except for the database type.

I also came across this doc, [Production Enterprise Install on Ubuntu 14.04 LTS](https://docs.mattermost.com/install/ee-prod-ubuntu.html), which doesn't seem to be in the ToC and at first glance looks pretty much identical to [Production Install on Ubuntu 14.04 LTS](https://docs.mattermost.com/install/prod-ubuntu.html). If the Enterprise Install doc is to remain, then that would be four separate docs that contain mostly identical information. This can be problematic when it comes to updating. For example, if changes are required in the NGINX server SSL section, it might only end up being updated on one of the four pages, unintentionally leaving the other three in an incorrect state and leading to confusion and uncertainty among those who use the docs.

Ideally, the docs should be split up and made more granular so that each section gets its own page.

In any case, what should be done with this PR? Reject it until things get sorted out? Or merge it in anyway so that at least the MySQL info is there and then deal with the organization of the docs?

it33:
@JeffSchering highly appreciate the PR, near term, 

I've added labels to have this PR reviewed, and will defer to the reviewer's decision. 

Long term, there are options inside of our documentation framework, Sphinx, to create the options for including either MySQL or PostgreSQL instructions and either Team Edition or Enterprise Edition instructions into concise install guides.

esethna:
Thanks @JeffSchering. I think it's better to get this info in for now. Regarding the organization, RST offers the option to use an "include", to copy sections across multiple different documents. Could we try to use this command to copy over the `Set up database server` section to the ee-prod-ubuntu.rst doc? Here is an example of the [syntax](http://docutils.sourceforge.net/docs/ref/rst/directives.html#including-an-external-document-fragment) from the [docker-ebs install guide](https://github.com/mattermost/docs/edit/master/source/install/docker-ebs.rst):

image

We can use the `start-after` and `end-before` flags to include only the database section.

JeffSchering:
One problem with this solution is that some future editor of this doc would have no idea that the MySQL section is used elsewhere, and might inadvertently make changes that are not appropriate for those other docs.

Maybe we can do this:
1. Merge this PR
2. I'll create a separate file for the MySQL stuff, then update the 4 Ubuntu install files to include the MySQL bit, and submit a PR for that.
3. At some point in the future the 4 Ubuntu install files (and others like it) can be chunked into single-topic files. Then the install docs would end up being a bunch of include statements and not much more. The chunked files could have comments in them about where they are used and precautions to take when modifying them.

esethna:
@JeffSchering Cool, that makes sense.

Is anyone from the community able to help test this PR? I want to make sure we get a thorough dev review since this changes some other things in addition to adding the MySQL instructions.

source/install/prod-ubuntu.rst
```diff
@@ -95,48 +100,88 @@ Set up Database Server
...
    -  ``vi config.json``
-   -  replace ``DriverName": "mysql"`` with ``DriverName": "postgres"``
-   -  replace
-      ``"DataSource": "mmuser:mostest@tcp(dockerhost:3306)/mattermost_test?charset=utf8mb4,utf8"``
-      with
-      ``"DataSource": "postgres://mmuser:mmuser_password@10.10.10.1:5432/mattermost?sslmode=disable&connect_timeout=10"``
-   -  Optionally you may continue to edit configuration settings in
+   - If you are using PostgreSQL:
+      1. Set ``DriverName":`` to ``"postgres"``
+      2. Set ``"DataSource:"`` to the following value: ``"postgres://mmuser:mmuser_password@10.10.10.1:5432/mattermost?sslmode=disable&connect_timeout=10"``
```
enahum: at the top of this section we say: **For the purposes of this guide we will assume this server has an IP
     address of ``10.10.10.2``**

but here we are using `10.10.10.1` which is never mentioned

JeffSchering: Can you verify?

The address of the Mattermost server is 10.10.10.2 and the address of the database server is 10.10.10.1.

The command `postgres://mmuser:mmuser_password@10.10.10.1:5432/mattermost?sslmode=disable&connect_timeout=10` is connecting to the database server, so 10.10.10.1 should be correct.

enahum: yeap, sorry my bad

esethna:
Thanks @JeffSchering! Merged!

PR #575, opened 2016-10-06
Fix JS style-guide syntax error

StraylightSky:
Strings section for javascript was using double quotes instead
of back tick.


esethna:
Thanks @StraylightSky! Merging.

PR #580, opened 2016-10-08
Fix the URL of Docker installation for Ubuntu

yuya-oc:
https://docs.docker.com/installation/ubuntulinux/ is not available.
https://docs.docker.com/engine/installation/linux/ubuntulinux/ seems new instruction.


esethna:
+1, confirmed with HH

PR #581, opened 2016-10-08
Remove flicker when navigating to links internal to the current page

shieldsjared:
Currently when selecting a link in the LHS that is internal to the currently displayed page, there is a noticeable "flicker" due to a recentering function.  This was previously used to accommodate the page content floating underneath of the header.  Since that issue has been fixed, the recentering is unecessary.  Removing this removes the "flicker".

To visualize the issue:
1. Navigate to the [Upgrade Guide](https://docs.mattermost.com/administration/upgrade.html)
2. Within the left-hand-sidebar, select a subsection of the document, such as **Upgrade Enterprise Edition**

Notice the flicker on the screen as the page scrolls, and then recenters (again).

Removing the javascript removes the flicker.

### DIFF
--- a/source/myscript.js
+++ b/source/myscript.js
@@ -8,18 +8,5 @@ $(document).ready(function(){
 		var sidebarScrollPosition = $('.wy-menu-vertical li.current>a').offset().top;
 		$('.wy-side-scroll').scrollTop(sidebarScrollPosition-120);
 	}
-
-	var centerScroll = function(position){
-		setTimeout(function(){
-			var centerScrollPosition = $('.wy-nav-content-wrap').scrollTop();
-			$('.wy-nav-content-wrap').scrollTop(centerScrollPosition-position);
-		},100);
-	};
-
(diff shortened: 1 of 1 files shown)


esethna:
Thanks Jared, asking @asaadmahmood with help for dev review.

asaadmahmood:
That's going to make the header appear above the heading for some users I think, which was initially the reason why I added this. We can still comment it out and test it for different users, if the issue comes back, we can add it back again.

shieldsjared:
@asaadmahmood - correct it was to reposition the header when the content of the page was overplayed by the top navigation bar. In one of my prior PRs I corrected that issue. You'll see that the top of the scroll bar on the article content is now visible (previously the top was hidden by the top navigation bar, hence the reason to offset the scroll.  Do you still want me to convert it to commented out code, given that?

asaadmahmood:
@shieldsjared If you're sure that the issue is fixed, then it's fine. We can always trace the file back if we want.

esethna:
Thanks guys, I will merge this in.

PR #582, opened 2016-10-08
Added search box to the homepage

shieldsjared:
Added a search box to the homepage, at the request of @lfbrock.

image

I'm not entirely sold on the "looks", but it's functional.  Any suggestions @asaadmahmood @it33 @esethna?  Feel free to make direct modifications or let me know if there's something I can change.

### DIFF
--- a/source/templates/index.html
+++ b/source/templates/index.html
@@ -21,6 +21,31 @@
             margin: .67em 0
         }
 
+        .landing .landing-search {
+            overflow: hidden;
+            padding:20px;
+        }
+
+        .landing .landing-search input[type="text"] {
+            width: 640px;
+            border-style: none;
+            padding: 8px 12px;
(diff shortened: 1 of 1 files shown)


it33:
Great work, thanks Jared! On my phone right now, just curious, wondering how much effort it would be to place the cursor in the input box, maybe have the text in the background? 

Sort of like how the message input box works in Mattermost. Not sure if that is more an Asaad question.

Overall really wonderful, makes the old UI look like it's really missing something :)

shieldsjared:
@it33 - Set the input field to autofocus, left the placeholder text, and squashed the commits.

it33:
Changes look straight forward so going to commit, @asaadmahmood, @esethna open to new PRs if you'd like to adjust the look,

PR #588, opened 2016-10-10
Simplifying production Docker install instructions

it33:


it33:
Merged minor update to Production Docker install instructions in an attempt to simplify.

PR #597, opened 2016-10-14
Added developer setup troubleshooting section

hmhealey:
I also removed the section of the setup that asked users to run `make test` when doing initial setup since that can take 10-20 minutes on some machines and it would unintentionally create the first system admin account so the user would be unable to log in without recreating their database.

### DIFF
--- a/source/developer/developer-setup.md
+++ b/source/developer/developer-setup.md
@@ -1,6 +1,8 @@
 Developer Machine Setup
 -----------------------------
 
+If you run into any issues getting your environment set up, please check the Troubleshooting section at the bottom for common solutions.
+
 ### Mac OS X ###
 
 1. Download and set up the latest version of VirtualBox. [https://www.virtualbox.org/wiki/Downloads](https://www.virtualbox.org/wiki/Downloads)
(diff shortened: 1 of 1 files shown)


source/developer/developer-setup.md
```diff
@@ -174,34 +172,56 @@ Any issues? Please let us know on our forums at: http://forum.mattermost.org
...
+
+#### I can't log into Mattermost because I don't have an account
+
+You can create an account using the following command:
+```
+go run mattermost.go -create_user -email="user@example.com" -password="mypassword"
+```
+Optionally, you can make that account a System Admin with the following command:
+```
+go run mattermost.go -assign_role -email="user@example.com" -role="system_admin"
```
jwilander: I think you need to have the role as `system_admin system_user` since assigning roles will overwrite all the user's previous roles

source/developer/developer-setup.md
```diff
@@ -220,4 +240,17 @@ Use `make run-fullmap` to produce JavaScript that you can look at in a debugger.
...
+Since running every single unit test takes a lot of time while making changes, you can run a subset of the serverside unit tests by using the following:
+```
+go test -v -run='<test name or regex>' ./<package containing test>
+```
+For example, if you wanted to run `TestPostUpdate` in `api/post_test.go`, you would run the following:
+```
+go test -v -run='TestPostUpdate' ./api
+```
+
+Alternatively, if you're writing clientside code, you can run only the clientside unit tests by using `make test-client`.
```
jwilander: Minor but `clientside` should be `client-side`

source/developer/developer-setup.md
hmhealey: Ah right. I remember we discussed that, but I forgot to update it here

hmhealey:
@jwilander Fixed both of those

lfbrock:
+1

esethna:
+1 thanks HH!

PR #603, opened 2016-10-18
update release schedule

reach3r:
update release schedule to the bi-monthly cycle, as is already written here: https://docs.mattermost.com/process/release-process.html#release-process


esethna:
+1, thanks!

PR #604, opened 2016-10-18
Favorite channels documentation

lfbrock:


esethna:
+1, added a commit to fix the images as they weren't appearing.

jasonblais:
We might want to scale the images down a bit. Would assume there to be some RST command that allows for it

image

lfbrock:
@eric - actually I had the images written so they would appear on the generated HTML pages with an appropriate size; I tested on my local instance and I believe the previous text should work on the actual site. 

The problem is there's no good way to set image size for the markdown docs other than using HTML, and the images in the generated site are in an `_image` folder and not an `image` folder, so they can't appear in both GitHub and the website using the same code.

lfbrock:
@jasonblais it's an old doc so it's not RST.

esethna:
Okay reverted the image syntax. +1

PR #605, opened 2016-10-18
Update config-settings for rate limiting

jasonblais:
There's a followup Jira ticket to add an explanation of how the rate limiting settings affect the Mattermost instance https://mattermost.atlassian.net/browse/PLT-4406

Discussed with Christopher, and we felt that would be nice to add, particularly with our performance testing recently.


esethna:
+1

esethna:
@jasonblais Is that coming in separately or should we mark this PR as WIP?

jasonblais:
It'll be separate, not part of this PR

PR #606, opened 2016-10-20
Increase minimum go version in Ubuntu instructions

grundleborg:
Increase from 1.5.1 to 1.6 to be consistent with MacOS instructions, as
platform no longer works with go 1.5.

Now I've done this, I'm wondering if it would be even better just to say the "latest" version, or to declare the versions needed at the top of the document before the platform/distro specific instructions begin?

### DIFF
--- a/source/developer/developer-setup.md
+++ b/source/developer/developer-setup.md
@@ -48,7 +48,7 @@ Any issues? Please let us know on our forums at: https://forum.mattermost.org/
 		- `127.0.0.1 dockerhost`
 3. Install build essentials
 	1. `apt-get install build-essential`
-4. Download Go 1.5.1 from [http://golang.org/dl/](http://golang.org/dl/)
+4. Download Go 1.6 from [http://golang.org/dl/](http://golang.org/dl/)
 5. Set up your Go workspace and add Go to the PATH
 	1. `mkdir ~/go`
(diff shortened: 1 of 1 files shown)


esethna:
@grundleborg would we be okay to say the following?

```
4. Download the lastest version of Go from http://golang.org/dl/
  - Go 1.5 and earlier is not supported.
```

grundleborg:
That would probably be the ideal thing to do, yes. But I can't say for sure whether there is a reason that elsewhere in the docs 1.6 is recommended (see the macOS section on that page), or whether that was just the latest at the time. You'd probably need to check with someone who knows the project better than me to be sure :)

esethna:
It's likely that the doc was just not updated. Adding @jwilander to confirm:

1. Can we just say "latest version" throughout the doc?
2. Is Go 1.5 and earlier no longer supported throughout the doc?

jwilander:
1. Let's be specific with a version because we usually like to be a major version behind and that also lets us test the newer Go versions before switching to them.
2. I believe vendoring was introduced in 1.6, so yes we would no longer support 1.5

jwilander:
Throwing it to PM review so @esethna can ask for changes if needed

esethna:
Thanks JW. I've added a note on not supporting Go 1.5.

PR #539, opened 2016-09-29
Draft of instructions for setting up Mattermost server using TLS

crspeller:

PR #608, opened 2016-10-20
Rebase and Tweak TLS setup instructions

esethna:
Already dev approved in https://github.com/mattermost/docs/pull/539


esethna:
Thanks, updated

PR #619, opened 2016-10-27
Noting double SSO is not supported

it33:


it33:
Merging to answer user question

PR #620, opened 2016-10-27
PLT-4320 Fix table of contents links in install guides

lfbrock:
On install pages (eg https://docs.mattermost.com/install/prod-rhel-6.html), if you click the links "Set up Database Server", "Set up Mattermost Server", etc, the page currently does not jump. 

This should fix the issue.


esethna:
+1 Thanks LB!

PR #628, opened 2016-10-30
Team names now at least 2 characters long instead of 4

jasonblais:
Also fixes a bug in `../settings/team-settings.md` as team name should only be at most 15 characters long.


lfbrock:
+1

esethna:
+1

PR #635, opened 2016-11-01
[Doc] letsencrypt-auto certonly need to be configured against nginx

gmorel:
screenshot from 2016-11-01 11 42 43
Otherwise it is displaying an error since trying to use Apache

screenshot from 2016-11-01 11 44 22


esethna:
Thanks @gmorel! Merging

PR #645, opened 2016-11-04
Renamed Video and Audio calling, adding more info

it33:


esethna:
Combined in another commit to the changelog. Will go through PM review when the changelog PR gets merged.

Issue #631, opened 2016-10-31
[Help Wanted] Help fix LEFT SIDE menu button on docs.mattermost.com on mobile web


it33:
Repros: 

- Go to docs.mattermost.com on mobile web and click the LEFT SIDE menu button 

![image](https://cloud.githubusercontent.com/assets/177788/19844421/c7fc03fe-9ee9-11e6-88e2-f8d4e0932cc7.png)

Observed: 

- Blank sidebar appears : 

![image](https://cloud.githubusercontent.com/assets/177788/19844376/3ef2df1a-9ee9-11e6-9edd-6b6c77b3301d.png)

Expected: 

- LEFT SIDE menu button should either not appear on docs.mattermost.com on mobile web, or it should redirect to https://docs.mattermost.com/overview/index.html

shieldsjared:
@it33 - this issue has been resolved

jasonblais:
Closed via #664, thanks @shieldsjared!

PR #664, opened 2016-11-16
Mobile / Fix LHS Menu Button

shieldsjared:
Github Issue #631 

Removed the sphinx-generated top navigation bar from the mobile view entirely.  It only provided access to the table of contents, which on a mobile view, was not responsive.  In the mobile view, relying on the use of the TOC tree and use of the breadcrumbs for navigation.  Will need to enable a means to access the search box w/in the mobile view from any page in a later PR....

### DIFF
--- a/source/theme.css
+++ b/source/theme.css
@@ -274,6 +274,9 @@ header .header__links li a {
 	header .links__icon {
 		display: block;
 	}
+	.wy-nav-top {
+		display: none;
+	}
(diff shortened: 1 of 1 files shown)


esethna:
Thanks Jared! @asaadmahmood please take a look when you have a chance?

asaadmahmood:
+1, looks fine to me.

Issue #630, opened 2016-10-31
[Help Wanted] Add search bar to mobile web experience for docs


it33:
Everyone loves the new search experience for docs on desktop: 

![image](https://cloud.githubusercontent.com/assets/177788/19844345/01d82e3c-9ee9-11e6-9569-fa41f166855f.png)

It's a little more difficult to use on mobile:

![image](https://cloud.githubusercontent.com/assets/177788/19844333/e6aa79c6-9ee8-11e6-80c1-536b4751b438.png)

jasonblais:
Closed via #665, thanks @shieldsjared!

PR #665, opened 2016-11-17
Added a search box to the nav header for the mobile view.

shieldsjared:
Github Issue: #630 

** Recommend merging PR #664  before this one otherwise the UX will look FUNKY **

Added a magnifying glass icon to the top header, and centered the mattermost logo/header.  When clicking the magnifying glass, a searchbox appears.  Also, added safety catches in case the hamburger menu is already open (and vice versa).

image

image

### DIFF
--- a/source/myscript.js
+++ b/source/myscript.js
@@ -1,7 +1,17 @@
 $(document).ready(function(){
 
 	$('header .links__icon').on('click', function (){
-		$(this).next().slideToggle();
+		$("header .header__searchbar").hide();
+		$("header .header__links").slideToggle();
+	});
+
+	$('header .search__icon').on('click', function (){
+		$("header .header__links").hide();
+		$("header .header__searchbar").slideToggle(400, function() {
+			if($(this).is(':visible')) {
+				$(this).find("input[type=text]").focus();
+			}
+		});
 	});
 
 	if($('.wy-menu-vertical li.current>a').length){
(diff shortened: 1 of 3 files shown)


asaadmahmood:
If we can remove the dark border for the search bar on focus, that would be good.
screen shot 2016-11-18 at 11 11 08 pm

+1 apart from that, looks good.

it33:
@asaadmahmood actually, I'm wondering if the design is better with the dark border since it's more clear the box has focus--what's the alternative otherwise? 

Just to check, when you hit the search icon, is the cursor in the search box right away? If so, I think we're good to go. 

@asaadmahmood?

asaadmahmood:
@shieldsjared had a talk with @it33, we're going to remove the focus from the mobile experience, or if we want we can just have it slightly darker than the default border, the reason for that being that the element is already in focus on the screen on mobile, and there's a hint towards the focus when the keyboard pops up, which makes it quite apparent. Thus we don't really need that dark border on mobile for the focus state. Let me know if that sounds good.

shieldsjared:
I'm good with whatever :) I'll submit the update tomorrow.

shieldsjared:
Ok, @it33 and @asaadmahmood.  I removed the black border (wasnt there intentionally, just had to add some additional styles to the textbox to match the standard-view search box styling).  Also, added a call to ensure that when you press the magnifying glass, focus is automatically set on the textbox.

(commits squashed)

it33:
I think Asaad's already done the dev review for this, given he's generated the output and noticed the black outline, merging as we seem to be blocking on the wrong PRs...

image

PR #674, opened 2016-11-22
Fixed a typo

JeffSchering:
A missing LF was causing the code block to not display correctly in Step 7 of _Set up NGINX Server_


esethna:
Thanks!

PR #677, opened 2016-11-23
Fixing typo in config-settings.rst

akrasian:
Entering "tls" in the config file will cause running mattermost to fail with the message "Invalid value for webserver connection security.".
However, "TLS" works and will correctly configure mattermost to run with HTTPS.


esethna:
+1

PR #678, opened 2016-11-23
Update formatting-text to rst

lfbrock:
Update to rst, and fixed incorrect list for syntax highlighting.

I think the new image should appear as follows on the actual docs site, not sure why they look so large on GitHub:

screenshot 2016-11-23 17 58 18


esethna:
+1, I will check the images after merge

PR #679, opened 2016-11-24
Correct meeting timezones.

grundleborg:
This refers to all meetings/events in "San Francisco time", as discussed on pre-release, which is when they actually are, not UTC-7 / UTC-8 as used in the documentation.

### DIFF
--- a/source/process/release-process.md
+++ b/source/process/release-process.md
@@ -12,7 +12,7 @@ Recommended pre-reading:
 ## Release Timeline
 
 Notes: 
-- All cut-off dates are based on 10am PST (UTC-07/08) on the day stated. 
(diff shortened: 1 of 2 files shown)


it33:
@jasonblais are you approving without the timezone link change or with it? 

I'm okay either way, just not sure if it's clear to @grundleborg 

Here are the options: 

1. http://everytimezone.com/ (compares SF timezone with other timezones) 
2. https://www.timeanddate.com/worldclock/usa/san-francisco (shows time in SF only) 

image

jasonblais:
Sorry, I had missed your note. 

@grundleborg Let's go with http://everytimezone.com/ as it includes other timezones and is easier to compare against your own.

Otherwise looks good to me.

grundleborg:
@it33 changed the timezone links as you suggested, thanks.

PR #680, opened 2016-11-24
Release process updates based on feedback from v3.5

esethna:
May need tweaks after platform discussion on 11/25


esethna:
Updated based on feedback from platform meeting

lfbrock:
Thanks @esethna, it looks good!

Can we also: 
1. Update the bullet points to be consistent with ending in period vs not ending in period. 
2. Update the point about adding download links to the download page to clarify which links should be added there, and to test the links and make sure they work

esethna:
Updated

PR #681, opened 2016-11-26
Fix urls for GitLab configuration

quentinus95:
Fix urls for GitLab configuration.


jasonblais:
Thanks @Coac for the review! :+1:

PR #682, opened 2016-11-28
Propose no exclamation points in errors

it33:
- North American-centric expression of excitement might not translate well across global cultures.

Reaction to this message: 
image


yangchen1:
Fixed

esethna:
+1, I fixed minor typo

PR #683, opened 2016-11-28
Increasing depth of Core Team Handbook

it33:


it33:
Small tweak will just merge this, don't think it'll conflict with anything.

PR #684, opened 2016-11-28
Increasing page depth to 2

it33:
Our docs are growing, proposing increasing page depth to make things easier to find

I pushed a change to Core Team Handbook and it seems like a great improvement: https://docs.mattermost.com/guides/core.html

### DIFF
--- a/source/guides/administrator.rst
+++ b/source/guides/administrator.rst
@@ -7,7 +7,7 @@ Install Guides
 ==============
 
 .. toctree::
-   :maxdepth: 1
+   :maxdepth: 2
    :glob:
 
    /install/requirements*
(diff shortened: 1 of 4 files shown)


it33:
- Small tweak to TOC depth to make things easier to find

it33:
Example of depth = 2 (easy to find things that would otherwise be buried) 

image

Example of depth = 1 (hard to find things) 

image

esethna:
+1 on these changes as long as it does not expand the default sidebar nesting level when you click into the integration guide, user guide etc. Based on the Core Team Handbook section it looks like it just changes the landing page, so I think we can push these changes then I will verify after merge.

Default nesting level of the guides in the sidebar shouldn't change:

image

PR #689, opened 2016-11-30
Update and rename messaging-basics.md to messaging-basics.rst

lfbrock:


jasonblais:
We should come up with a guideline on whether to capitalize "Shift", "Alt", and other keyboard buttons in our docs. We're really inconsistent right now.

image

I looked into this previously, and there isn't really a standard as far as I know. We however use capital letters in `shortcuts` and keyboard shortcuts docs (+ in a few other places), so propose using capitals.

Up for discussion though.

source/help/getting-started/messaging-basics.rst
```diff
@@ -0,0 +1,45 @@
+Messaging Basics
+================
+
+--------------
+
+**Write messages** using the text input box at the bottom of Mattermost.
+Press **ENTER** to send a message. Use **Shift+ENTER** to create a new
```
jasonblais: We should come up with a guideline on whether to capitalize "Shift", "Alt", and other keyboard buttons in our docs. We're really inconsistent right now.

image

I looked into this previously, and there isn't really a standard as far as I know. We however use capital letters in `shortcuts` and keyboard shortcuts docs (+ in a few other places), so propose using capitals.

Up for discussion though.

jasonblais: Another thing is whether to have spaces between the keyboard button labels, e.g. `SHIFT+ENTER` or `SHIFT + ENTER`.

Again, we're pretty inconsistent.

enahum:
Also changes in the text should be included in platform too

lfbrock:
The only text change was from Shift+Enter to SHIFT+ENTER, but not sure if it's worth updating now in platform because we probably want someone to search for other similar instances in the product help text and make those changes as well.

jasonblais:
@wget Offered to search for other similar instances in the platform repo earlier

wget:
@jasonblais Yep on my todo actually. And wrt. to what we do in others FOSS projects, those have to be uncapitalized

Alt+Shift+H+Enter for example. And these keys have to be translated. Alt+Maj+H+Entrée in French for example.

jasonblais:
Chatted with @wget on the standard in most FOSS projects when it comes to keyboard buttons. I'll bring this up in PM meeting to discuss.

Doesn't block this PR so I'll merge this.

If any changes are needed, I'll add it to our docs style guides, and do a pass across both the platform and docs repos.

PR #692, opened 2016-12-01
Clean up API documentation

jwilander:
There was a bunch of misplaced and redundant information in the documentation for our APIs. I've removed the unrelated information and made the links to api.mattermost.com a lot more prominent since that's where developers should be getting the majority of their information about our APIs.

I also added the API docs to our developer's guide because they belong there as much as they belong in the integration guide (I left it there as well).

Not sure if you want to put the information I removed about webhooks and slash commands somewhere else (they already have pages of their own) but it doesn't really belong with our API documentation.

Everything removed in source/developer/web-service.md already exists on api.mattermost.com.

Feedback welcome.


source/developer/web-service.md
```diff
@@ -1,178 +0,0 @@
-# Web Service API Overview
```
esethna: Is removing this page going to break links? Should we just add a tomb stone linking to api.mattermost.com and remove it from the docs sidebar, rather then entirely deleting it?

jwilander: Sure, makes sense. How do I do that?

lfbrock:
+1

PR #707, opened 2016-12-05
Adding intro & TOC to production install guides

it33:


it33:
Verified TOC changes went through properly

PR #710, opened 2016-12-05
Updating TOCs missed

it33:


it33:
Merging now and will verify after merge

PR #712, opened 2016-12-05
Removing TOC from Docker preview install

it33:
Breaks formatting because of includes


it33:
Merging to fix bug and will verify this works

PR #726, opened 2016-12-09
Renaming file to remove sg_ prefix

JeffSchering:


JeffSchering:
Good catch @esethna. Thanks!

PR #731, opened 2016-12-12
Updated subsection underlines

JeffSchering:
... to conform to the underlines that Pandoc produces when converting files.


lfbrock:
+1

PR #732, opened 2016-12-13
Adding credits to core team handbook

it33:
- Adding credits since we're pulling from these different sources pretty frequently. 
- Also adjusting headings to match style guide


yangchen1:
+1.

PR #739, opened 2016-12-15
Help section about math formulas

nils-werner:
The formatting section had no content about math formulas. This PR adds one, including a screenshot.


source/help/messaging/formatting-text.rst
```diff
@@ -235,3 +235,19 @@ Create a table by placing a dashed line under the header row and separating the
...
+
+.. code::
+
+  ```latex
+  X_k = \sum_{n=0}^{2N-1} x_n \cos \left[\frac{\pi}{N} \left(n+\frac{1}{2}+\frac{N}{2}\right) \left(k+\frac{1}{2}\right) \right]
+  ```
+
+Renders as:
+
+.. image:: ../../images/markdownMath.PNG
```
esethna: PNG should be lowercase to get the image to work correctly (since the file extension is in lowercase)

source/help/messaging/formatting-text.rst
```diff
@@ -235,3 +235,19 @@ Create a table by placing a dashed line under the header row and separating the
 Renders as:
 
 .. image:: ../../images/markdownTable1.PNG
+
+
+Math Formulas
```
esethna: Propose moving this as a sub section of `Code Block` under the `Synax Highlighting` section.

source/help/messaging/formatting-text.rst
nils-werner: Ah, didn't notice that, sorry.

source/help/messaging/formatting-text.rst
nils-werner: I disagree. The fact that it is done using a `Code Block` is just an implementation detail. None of the features associated with a `Code Block` are available in math mode: syntax highlighting, themes, or the 4-space indented codeblock syntax.

Also, think of it like this: A user tries to search the docs for how to typeset a formula. The fact that it is done using a `Code Block` is most likely not known to the user. So putting the feature under `Code Block` actually obscures the information by putting it under a seemingly unrelated headline.

The way it is formulated now improves discoverability while highlighting the detail that is technically a `Code Block`.

esethna: Thanks for the explaination, that makes sense, I'm okay with keeping it separate.

source/help/messaging/formatting-text.rst
esethna: Thanks!

lfbrock:
+1

PR #740, opened 2016-12-15
Add a note to Site URL setting

jasonblais:
A community member was blocked on their deployment because they appended a team name to the site URL.

As a result, no one was able to create an account on the system

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -43,6 +43,8 @@ Site URL
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The URL, including port number and protocol, from which users will access Mattermost. Leave blank to automatically configure based on incoming traffic.
 
+.. note:: Do not append a team name to the end of the site URL.
+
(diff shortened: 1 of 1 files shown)


source/administration/config-settings.rst
```diff
@@ -43,6 +43,8 @@ Site URL
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The URL, including port number and protocol, from which users will access Mattermost. Leave blank to automatically configure based on incoming traffic.
```
esethna: Should we give an example like `https://domain.com`?

lfbrock: Sorry @JeffSchering not sure what you mean by a URL is more than just the protocol+domain+port - I think it is the URL to the root page of the site? Do you mean Site URL doesn't make it clear that team shouldn't be included?

JeffSchering: Yes, at least to me, *Site URL* makes it sound as if you can use something like *https://mattermost.example.com:1234/something/custom* because it's a URL too.

esethna:
Tweaked based on feedback and provided examples

lfbrock:
+1

PR #744, opened 2016-12-16
Update and convert channel settings to rst

lfbrock:
One thing that still needs to be updated in this doc is "dropdown" or "drop-down", wasn't 100% sure what was correct.


JeffSchering:
As a compound adjective, it would be *drop-down*. If used as a noun, then *dropdown*.

I'd be tempted to to leave it out altogether because it looks more like a pop-up than a drop-down; drop-downs typically have a down-arrow that you click. So in this case, you could use *channel menu* instead of *channel drop-down menu*.

lfbrock:
Thanks @JeffSchering, updated it to drop-down - I left it in there for now because technically there is an arrow beside the channel name, but I'm open to taking it out if other people think it's clearer without "drop-down" there. 
screenshot 2016-12-19 13 10 24

source/help/settings/channel-settings.rst
```diff
@@ -0,0 +1,59 @@
+Channel Settings
+================
+
+--------------
+
+Notification preferences, channel header, channel purpose, and channel
+name are customizable for each channel. To access these settings, click
+the channel name at the top of the page to open the channel drop-down
```
esethna: I think I agree with taking out drop-down when we refer to the menus. I think of these as drop-downs, not so much the channel menu or the main menu.
image

source/help/settings/channel-settings.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Notification preferences can be modified for each channel you belong to.
+
+Send Desktop Notifications
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, the desktop notification preference assigned in *Account
+Settings* is used for all channels. To customize the desktop
+notification preference for each channel, click the channel name at the
+top of the page to access the drop-down menu, then click
```
esethna: > channel menu

source/help/settings/channel-settings.rst
```diff
@@ -0,0 +1,59 @@
...
+**Notification Preferences > Send Desktop Notifications**.
+
+Mark Channel Unread
+~~~~~~~~~~~~~~~~~~~
+
+By default, channel names are bolded for all new messages in a channel. 
+To only bold the channel name when you are mentioned, open the channel drop-down menu and click
```
esethna: drop-down

source/help/settings/channel-settings.rst
```diff
@@ -0,0 +1,59 @@
...
+
+By default, channel names are bolded for all new messages in a channel. 
+To only bold the channel name when you are mentioned, open the channel drop-down menu and click 
+**Notification Preferences > Mark Channel Unread > Only for mention**.
+
+
+Channel Header
+--------------
+
+Select **Edit Channel Header** to change the text that appears next to
```
esethna: "Select **Edit Channel Header** to change" > "Select **Edit Channel Header** from the channel menu to change..."

source/help/settings/channel-settings.rst
```diff
@@ -0,0 +1,59 @@
...
+the channel name at the top of the screen. It can be used to summarize
+the channel topic or provide links to frequently accessed documents. Any
+channel member can edit this setting, unless the System Administrator
+has `restricted the
+permissions <https://docs.mattermost.com/administration/config-settings.html#enable-public-channel-management-permissions-for>`__.
+
+Channel Purpose
+---------------
+
+Select **Edit Channel Purpose** to change the text that appears in the
```
esethna: Same comment as above

source/help/settings/channel-settings.rst
```diff
@@ -0,0 +1,59 @@
...
+**Notification Preferences > Mark Channel Unread > Only for mention**.
+
+
+Channel Header
+--------------
+
+Select **Edit Channel Header** to change the text that appears next to
+the channel name at the top of the screen. It can be used to summarize
+the channel topic or provide links to frequently accessed documents. Any
+channel member can edit this setting, unless the System Administrator
```
esethna: System Administrator > System Admin

source/help/settings/channel-settings.rst
```diff
@@ -0,0 +1,59 @@
...
+has `restricted the
+permissions <https://docs.mattermost.com/administration/config-settings.html#enable-public-channel-management-permissions-for>`__.
+
+Channel Purpose
+---------------
+
+Select **Edit Channel Purpose** to change the text that appears in the
+*More…* menu for channels. The channel purpose is usually a short
+description that helps others decide whether to join the channel. Any
+channel member can edit this setting, unless the System Administrator
```
esethna: System Administrator > System Admin

source/help/settings/channel-settings.rst
```diff
@@ -0,0 +1,59 @@
...
+has `restricted the
+permissions <https://docs.mattermost.com/administration/config-settings.html#enable-public-channel-management-permissions-for>`__.
+
+Channel Name
+------------
+
+Select **Rename Channel** to change the channel name or URL. Any channel
```
esethna: Same comment as above

source/help/settings/channel-settings.rst
```diff
@@ -0,0 +1,59 @@
...
+member can edit this setting, unless the System Administrator has
```
esethna: System Administrator > System Admin

source/help/settings/channel-settings.rst
esethna: Also, we should maybe say "handle" instead of URL (and then explain it in a sentence) since that how it appears in the UI.

lfbrock: Weren't we going to rename that so it doesn't say channel handle anymore, since that's not very clear? I feel like URL is clearer, or we should add help text to handle since it's really confusing...

source/help/settings/channel-settings.rst
lfbrock: Do we really need to change this to System Admin? 

If so it might be best to make a separate issue for this, as I think there are many places in the docs and in the product help text that use Administrator.

source/help/settings/channel-settings.rst
lfbrock: Updated to remove drop-down

lfbrock:
@esethna - made the changes except the System Admin one, I don't think there's a guideline for that? And to me System Admin and System Administrator are interchangeable. But if it's something we actually want to do we should add it to guidelines and make a separate ticket since there are many places in the docs and product help text that use Administrator.

esethna:
Thanks LB. We decided to refer to them as "Admins" in the actual product. But yes, we should add this to the guidelines for docs and then check other places

yangchen1:
+1 :)

PR #748, opened 2016-12-19
Adding option to upgrade to EE in migration guide

it33:


it33:
Merging since this is needed to answer a question

PR #750, opened 2016-12-19
Add restoring ./data to migration guide

lfbrock:


source/administration/migrating.rst
```diff
@@ -17,10 +17,13 @@ The following instructions migrate Mattermost from one server to another by back
     1. Backup the database from the SOURCE Mattermost server and restore it in place of the database to which the DESTINATION server is connected
 5. Migrate ``config.json`` from SOURCE to DESTINATION  
     1. Copy of ``config.json`` file from SOURCE deployment to DESTINATION 
-6. Start the DESTINATION deployment  
+6. Migrate ``./data`` from SOURCE to DESTINATION
```
hmhealey: Maybe just add something that says this step is only necessary if you have local file storage enabled (`FileSettings.DirverName` is set to `local`)? This isn't necessary if you're using S3

PR #762, opened 2017-01-04
Update WebRTC image

jasonblais:


lfbrock:
Should we use someone who has a less blurry picture when enlarged instead? Maybe like Lindy?

jasonblais:
@lfbrock good point, we definitely should. Updated

PR #768, opened 2017-01-08
Update SAML troubleshooting doc + move sso-saml.md to rst

jasonblais:


source/deployment/sso-saml.rst
```diff
@@ -0,0 +1,113 @@
...
+
+`Configure SAML with Okta <http://docs.mattermost.com/deployment/sso-saml-okta.html>`__
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+`Configure SAML with Microsoft ADFS <http://docs.mattermost.com/deployment/sso-saml-adfs.html>`__
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Pre-installation
+~~~~~~~~~~~~~~~~
+
```
JeffSchering: This section of the doc is confusing. The links look like titles instead of links.

image

Shouldn't the Pre-installation section come before the Okta and ADFS install sections?

Also, if using the monolith style, that content be included in-line.

source/deployment/sso-saml.rst
```diff
@@ -0,0 +1,113 @@
+SAML Single-Sign-On (E20)
+-------------------------
+
+Overview
+~~~~~~~~
+
```
JeffSchering: The page title should have = as the underline character, and the next level section should have - as the underline character.

source/deployment/sso-saml.rst
jasonblais: Hmm, that's a good point

jasonblais:
@JeffSchering How does this looks like?

source/deployment/sso-saml.rst
jasonblais: As for monolith style, I'd be worried that the docs page would get too long.

For instance the [ADFS docs alone is long](https://docs.mattermost.com/deployment/sso-saml-adfs.html), and we're planning to add docs for OneLogin soon, in addition to Okta and ADFS.

source/deployment/sso-saml.rst
```diff
@@ -48,21 +40,21 @@ installed. If not, run - ``apt-get install libxmlsec1-openssl`` on
 Ubuntu - ``yum install xmlsec1-openssl`` on RHEL
```
JeffSchering: How about something like:

    Mattermost officially supports Okta and Microsoft ADFS as the identity providers (IDPs), 
    but you can also configure SAML for a custom IDP. (note absence of link to install doc, 
    and use of 'can' instead of 'may')

    - To set up SAML with Okta, see :doc:`sso-saml-okta`
    - To set up SAML with Microsoft ADFS, see :doc:`sso-saml-adfs`

Also, after looking at the Okta and ADFS doc, I see that each has its own Pre-installation section, so you can probably get rid of the Pre-installation section here. It's a duplication of effort, and can easily get out of sync with the Okta and ADFS SAML config docs.

jasonblais: If someone deploys SAML with another IDP like OneLogin, they wouldn't likely read through the Okta and ADFS guides.

Hence, for those admins, it is useful to include the pre-installation note here too, as otherwise it might be missed. Similarly with the troubleshooting section,

source/deployment/sso-saml.rst
```diff
@@ -0,0 +1,110 @@
+SAML Single-Sign-On (E20)
```
lfbrock: I think these might be "Single Sign-On", without the first hyphen

source/deployment/sso-saml.rst
```diff
@@ -0,0 +1,110 @@
...
+configuration process, they can set an existing account to System
+Administrator using `a commandline
+tool <http://docs.mattermost.com/deployment/on-boarding.html#creating-system-administrator-account-from-commandline>`__.
+
+2. Received error message: ``An account with that username already exists. Please contact your Administrator.``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This usually means an existing account has another authentication method enabled. If the user wants to use the existing account for SAML authentication, they should sign in using that method (such as email and password), then change their sign-in method to SAML via **Account Settings > Security > Sign-in method**.
+     
+This error message can also be received if the ``Username Attribute`` of their SAML credentials doesn't match the username of their Mattermost account. If so, the user can update the attribute at their identity provider (for instance, back to the old value if it had been previously updated).
```
lfbrock: Is this section accurate? 

If I have a different username in Mattermost, and then switch to SAML, I think it updates my username to be the saml username. Also I'm not sure if users can update the attribute at their identity provider (an admin may need to do that?).

jasonblais: It is -- the username is updated only if not already taken (confirmed with EN).

Reworded the section that an admin can update it (I wrote System Admin as they probably could update it would know the person who can)

lfbrock: Isn't that case covered by the first paragraph? Or is it possible to have a SAML system with two accounts with the same username?

jasonblais:
Chatted with @lfbrock  and removed the part about updating the username attribute in the IDP. It's not likely a preferred solution for an IT admin given SAML credentials are used for other services than just Mattermost.

lfbrock:
+1

PR #720, opened 2016-12-08
Update statistics.md and convert to RST

jasonblais:

PR #769, opened 2017-01-08
Convert statistics.md to statistics.rst and update for 3.6

jasonblais:
Addresses all feedback from https://github.com/mattermost/docs/pull/720


source/administration/config-settings.rst
```diff
@@ -2235,3 +2235,19 @@ Standard setting for OAuth to determine the scope of information shared with OAu
...
 +----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+
+________
+
+Analytics Settings
+```````````````````````````
+*Available in Enterprise Edition E10 and higher*
+
+Maximum Users for Statistics
+~~~~~~~~~~~~~~~~~~~~~~~~~
```
JeffSchering: This underline is too short for the title. The Sphinx processor will ignore it, although the GitHub .rst processor does not.

source/administration/statistics.rst
```diff
@@ -0,0 +1,70 @@
...
+**Total Teams:** The total number of teams created on your system.
+
+**Total Posts:** The total number of posts made in all the teams on your system, including deleted posts and posts made using automation.
+
+**Total Channels:** The total number of public channels and private groups created in all the teams on your system, including deleted channels. Doesn't include direct message channels.
+
+**Total Posts (graph):** The total number of posts made on a certain day in all the teams on your system, including deleted posts and posts made using automation.
+
+**Active Users With Posts (graph):** Users who made a post on a certain day in all the teams on your system, including system messages posted from the user's account.
+
```
JeffSchering: Would these be better as a definition list? The .rst markup for that looks like this:

    Total Users
      The total number of active accounts created on your system. Excludes inactive accounts.

    Total Teams
      The total number of teams created on your system.
    
    and so on

You wouldn't need the bold markers because the output to html is a \<dl\> element.

jasonblais: I tried it out and I think it makes sense here, since it really is a short list of definitions.

Should we add it to the doc style guides as well?

JeffSchering: Ok, I'll add it to Issue https://github.com/mattermost/docs/issues/772

jasonblais: +1

source/administration/statistics.rst
```diff
@@ -0,0 +1,93 @@
...
+    The number of active connections currently on your master database.
+
+Replica DB Conns
+    The number of active connections currently on one or more of `your read replica databases <https://docs.mattermost.com/deployment/cluster.html#database-configuration>`_.
+
+Channel Types
+    This chart displays the number of public channels and private groups in a visual format, including channels that may have been archived.
+
+Posts, Files and Hashtags
+    This chart displays the number of posts containing files, hashtags or text-only. Posts containing both files and hashtags are counted in both categories, and deleted posts are included.
```
lfbrock: "only text" instead of "text-only"

source/administration/statistics.rst
```diff
@@ -0,0 +1,93 @@
...
+Team statistics
+---------------
+
+Team Statistics are viewable under **System Console** > **Teams** > **Statistics**.
+
+Total Users
+    The total number of active accounts who are members of this team. Excludes inactive accounts.
+
+Public Channels
+    The number of public channels created in this team. Doesn't include archived channels.
```
lfbrock: Should we call this "deleted" channels since that's what we call it in UI?

lfbrock:
+1

PR #790, opened 2017-01-16
Update desktop troubleshooting section

jasonblais:


source/help/apps/desktop-guide.rst
```diff
@@ -4,7 +4,7 @@ Using Mattermost Desktop App
 
 Mattermost desktop applications are available for Windows, Mac and Linux operating systems. 
 
-You can `download the apps directly from our download page <https://about.mattermost.com/downloads/>`_ and visit our `installation guides <https://docs.mattermost.com/install/desktop.html>`_ for help during set up.
+You can `download the apps directly from our download page <https://about.mattermost.com/downloads/>`_ and visit our `installation guides <https://docs.mattermost.com/install/desktop.html>`_ for help during set up and for troubleshooting tips.
```
JeffSchering: Is the url https://docs.mattermost.com/install/desktop.html correct? I don't see any troubleshooting info on that page.

Is the URL https://docs.mattermost.com/help/apps/desktop-guide.html?

source/install/desktop.rst
```diff
@@ -71,3 +71,20 @@ To download the unofficial, community-driven .deb packages for the desktop appli
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The app data might be corrupted - remove all the files in `C:\Users...\AppData\Local\mattermost`, then try re-installing the app.
+
+Desktop App window is black and doesn't load the page
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you are using a special video driver such as Optimus, try disabling it to see if the problem is resolved.
+
+Otherwise, try disabling GPU hardware acceleration by using the `--disable-gpu <http://peter.sh/experiments/chromium-command-line-switches/#disable-gpu>`_ Chromium command line switch.
```
JeffSchering: FWIW: I get a black app window from time-to-time, and I found that CTRL+R fixes it.

jasonblais:
@JeffSchering Would you have any suggestions for re-organizing? Would be happy to do that

JeffSchering:
@jasonblais The first thing to do should be to make each set of instructions stand alone. When someone scans the page looking to add a team, they get this:

1. Click Add new team next to the right of Team Management section.
2. Enter Name and a valid URL, which begins with either http:// or https://.
3. Click Add

But then they have no idea what the starting point is because they scanned past the sentence "Team Management appears in the Settings page, available from the File menu under Settings." up above.

So the first step should be:

1. Click **File > Settings** to open the *Settings* page

Also the list of options and descriptions might be best as a definition list, and maybe the table should be in its own section called "Option Availability" or something like that, and maybe remove location references such as "below is a list" for future Accessibility compliance.

source/help/apps/desktop-guide.rst
jasonblais: Yeah, this PR moves the section to https://docs.mattermost.com/install/desktop.html

jasonblais:
I pushed a few tweaks @JeffSchering, let me know how that looks

source/help/apps/desktop-guide.rst
```diff
@@ -40,108 +40,95 @@ Editing servers and teams
...
+| Leave app running in notification center when application window is closed    |                           | Yes                       | Yes                       |
++-------------------------------------------------------------------------------+---------------------------+---------------------------+---------------------------+
 
 Menu Bar
 ---------------------------------------------------------------------
 
 The desktop app contains a menu bar with additional features and shortcuts to streamline your experience. 
 
-If the menu bar is hidden, you may use the ALT key to display the menu. To have the menu displayed at all times, go to the Settings page and uncheck the **Hide menu bar** setting.
+If the menu bar is hidden, you may use the ALT key to display the menu. To have the menu displayed at all times, go to the Settings page by pressing CTRL/CMD+COMMA and uncheck the **Hide menu bar** setting.
```
JeffSchering: The menu bar info should be at the beginning of the doc because its basically the starting point of user operations.

source/help/apps/desktop-guide.rst
```diff
@@ -4,32 +4,32 @@ Using Mattermost Desktop App
...
 
 Team Management
 ---------------------------------------------------------------------
 
 You can connect to multiple Mattermost servers from a single interface on the desktop app.
 
-Team Management appears in the Settings page, available from the `File` menu under `Settings`.
+The Team Management section allows you to add, edit and remove servers and teams.
 
 Adding servers and teams
```
JeffSchering: We have to be careful because it's easy for people to think that they can create a new team by following this procedure, when really all they're doing is adding an existing team to a list.

But then it's even more misleading because this doesn't actually add a team to the list. It adds a server and a name for the tab. The tab might have the same name as a team, but that would be a coincidence. You can have more than one team under the tab.

Maybe we need to revisit this doc outside the limits of a PR?

image

source/help/apps/desktop-guide.rst
```diff
@@ -4,32 +4,32 @@ Using Mattermost Desktop App
...
 
 To add a new server on the Mattermost instance, 
 
-1. Click **Add new team** next to the right of Team Management section.
-2. Enter **Name** and a valid **URL**, which begins with either ``http://`` or ``https://``.
-3. Click **Add**.
+1. Go to **File** > **Settings** or press CTRL/CMD+COMMA
+2. Click **Add new team** next to the right of Team Management section.
+3. Enter **Name** and a valid **URL**, which begins with either ``http://`` or ``https://``.
+4. Click **Add**.
```
JeffSchering: 3. In the **Name** field, enter the name that you want for the tab.
4. In the **URL** field, enter the complete URL of the server that you want to connect to. For example, `https://pre-release.mattermost.com/`
5. Click **Add**.

The UI has sample text for the fields, so it's not needed in the instructions.

source/help/apps/desktop-guide.rst
```diff
@@ -4,32 +4,32 @@ Using Mattermost Desktop App
...
 
 You can now access all teams you have joined in the server once you have saved the settings, and each will appear as a separate tab in the app.
```
JeffSchering: Actually, each server appears as a separate tab. You can create a tab for each server, but you get the same team selection page regardless. There's nothing stopping you from opening *TeamA* in the *TeamB* tab and *TeamB* in the *TeamA* tab. Or from opening *TeamA* in the *My Own Server* tab.

source/help/apps/desktop-guide.rst
jasonblais: Yeah, agree, it's a little misleading.

With the release of team sidebar feature, we can now easily display ALL teams in the same server tab, so we'll be renaming "team management" to "server management".

Previously you had to have a separate tab for each of your teams, if you wanted to quickly switch between them, or immediately see whether one of the teams has a mention or an unread. That's no longer the case as the team sidebar covers those.

I was planning to revisit the doc prior to the next Desktop App release (which is in February), and I'll make sure to address the point you raised here.

source/help/apps/desktop-guide.rst
```diff
@@ -4,32 +4,32 @@ Using Mattermost Desktop App
...
+The Team Management section allows you to add, edit and remove servers and teams.
 
 Adding servers and teams
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 To add a new server on the Mattermost instance, 
 
-1. Click **Add new team** next to the right of Team Management section.
-2. Enter **Name** and a valid **URL**, which begins with either ``http://`` or ``https://``.
-3. Click **Add**.
```
jasonblais: The thing with the current desktop release is that the top menu bar might be hidden for users, depending on the "Hide Menu Bar" option.

So I think it's better to give both in this case.

This "Hide Menu Bar" option is planned to be removed in the next desktop release as well -- once done, I can update this step to remove the "CTRL/CMD+COMMA" part

source/help/apps/desktop-guide.rst
```diff
@@ -4,32 +4,32 @@ Using Mattermost Desktop App
...
+1. Go to **File** > **Settings** or press CTRL/CMD+COMMA
```
jasonblais: Would that confuse the "Mattermost Main Menu"? 

0/5, maybe we should call it 'menu bar'?

source/help/apps/desktop-guide.rst
jasonblais: > The UI has sample text for the fields, so it's not needed in the instructions.

So should I add the example for step 4?

source/help/apps/desktop-guide.rst
jasonblais: That's correct -- see comment in https://github.com/mattermost/docs/pull/790#discussion_r97192597

source/help/apps/desktop-guide.rst
```diff
@@ -40,108 +40,95 @@ Editing servers and teams
...
++-------------------------------------------------------------------------------+---------------------------+---------------------------+---------------------------+
+| Toggle window visibility when clicking on the tray icon                       | Yes                       |                           |                           |
++-------------------------------------------------------------------------------+---------------------------+---------------------------+---------------------------+
+| Show icon on menu bar                                                         |                           | Yes                       | Yes                       |
++-------------------------------------------------------------------------------+---------------------------+---------------------------+---------------------------+
+| Leave app running in notification center when application window is closed    |                           | Yes                       | Yes                       |
++-------------------------------------------------------------------------------+---------------------------+---------------------------+---------------------------+
```
jasonblais: The table makes it easier for a new user to skim through the app options for their platform. That's what its original intention was at least

source/help/apps/desktop-guide.rst
jasonblais: Actually, the first time a user installs the app, they see the Settings page and they need to add a new server (or team).

Below that, they see the app options.

So the menu bar functionality is really only useful afterwards.

jasonblais:
Thanks Jeff! Left note to some of your feedback

source/help/apps/desktop-guide.rst
JeffSchering: ok

source/help/apps/desktop-guide.rst
JeffSchering: ok

source/help/apps/desktop-guide.rst
JeffSchering: Come to think of it, the list of menu items is superfluous too. It doesn't give any info that can't be seen in the UI. The idea is to have the UI as intuitive as possible, and have as much help info in the UI as possible, and keep the written help as small as possible, with only information that cannot be seen anywhere else. In this particular case, there's nothing really special about the menu items that warrants inclusion in the written help. Almost all of them are standard app items that any user is already familiar with.

JeffSchering: > If the menu bar is hidden, you may use the ALT key ...

Use of the word *may* can be ambiguous, especially to people who are new to English. It can mean *might*, or *can*, or *have permission to*. It's best to use either *might* or *can* instead of *may*. In this particular case though, you could probably get rid of "you may" altogether, so that you have:

>If the menu bar is hidden, use the ALT key ...

source/help/apps/desktop-guide.rst
JeffSchering: With the App Options as a definition list, it seems pretty easy to skim through for platform-specific options, and the table seems to be duplication of effort and add no new information to the page.

source/help/apps/desktop-guide.rst
jasonblais: I don't think it hurts to include a short description of what they do. I'll update the text here though

source/help/apps/desktop-guide.rst
jasonblais: If you want to quickly skim through Mac-specific options, how long would it take to find them on the definition list?

If it's easy, then yeah I guess the table is unnecessary.

The original intent of the table was to have a quick reference at the top, followed by more information after the table if needed.

source/help/apps/desktop-guide.rst
JeffSchering: True... Best support is no support. But it's also true that _The best help is no help_.

If the info is in the UI, then you don't need it in the help. And if people aren't sure about what one of the menu options does, then when they refer to the docs they should get more info than what's in the UI. For example, _Undo (CTRL+Z): Reverses previous action_ makes looking at the docs a waste of time, because it has no additional information. The description is basically the menu item reworded. Also, there's no one under the age of 80 who doesn't already know what Edit > Undo does.

Similarly for the other menu items, which are all pretty standard computer interface items. The only one that might be special is _Settings (CTRL+COMMA): Opens app settings where you can manage your servers and configure desktop app settings_ and it's easier to click that item and see what you get than it is to leave the app, go to the docs, and start looking around for what it means.

Crisp, tight docs without superfluous info are easier and cheaper to translate, easier to search through (because the search results are more relevant as they don't have superfluous info), and easier to read and use.

source/help/apps/desktop-guide.rst
```diff
@@ -40,108 +41,95 @@ Editing servers and teams
...
 
 Menu Bar
 ---------------------------------------------------------------------
 
 The desktop app contains a menu bar with additional features and shortcuts to streamline your experience. 
 
-If the menu bar is hidden, you may use the ALT key to display the menu. To have the menu displayed at all times, go to the Settings page and uncheck the **Hide menu bar** setting.
+If the menu bar is hidden, use the ALT key to display the menu. To have the menu displayed at all times, go to the Settings page by pressing CTRL/CMD+COMMA and uncheck the **Hide menu bar** setting.
 
 Below is a list of menu options with the corresponding keyboard shortcuts. For Mac, replace CTRL by CMD unless otherwise specified.
```
JeffSchering: The description for one of the keyboard shortcuts doesn't seem correct:
>Redo (CTRL+SHIFT+Z; CTRL+Y): Redoes the most recent action

CTRL+SHIFT+Z undoes the last undo. It doesn't redo the last action. For example, if the most recent action was a paste, then it doesn't paste again.

jasonblais: True, it's more like `redoes a previously undone action`.

I'm not sure how to write that clearly, having 'redoes' and 'undone' in the same sentence feels complicated.

JeffSchering: Would either of "Reverts the last Undo" or "Reverses the effect of the last Undo" work?

jasonblais:
Should be ready for review :)

source/help/apps/desktop-guide.rst
```diff
@@ -40,163 +41,62 @@ Editing servers and teams
...
+4. Click **Save**.
 
 Removing servers and teams
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To remove a server from your desktop app environment, click on **Remove** next to the server or a team you want removed.
+To remove a server from your desktop app environment, 
+
+1. On the menu bar, go to **File** > **Settings** or press CTRL/CMD+COMMA
+2. Click **Save** click on **Remove** next to the server or a team you want removed.
```
esethna: Don't you need to click remove, then click save?

source/help/apps/desktop-guide.rst
```diff
@@ -4,32 +4,33 @@ Using Mattermost Desktop App
...
 To add a new server on the Mattermost instance, 
 
-1. Click **Add new team** next to the right of Team Management section.
-2. Enter **Name** and a valid **URL**, which begins with either ``http://`` or ``https://``.
-3. Click **Add**.
+1. On the menu bar, go to **File** > **Settings** or press CTRL/CMD+COMMA
+2. Click **Add new team** in the Team Management section.
+3. In the **Name** field, enter the name that you want for the tab. 
+4. In the **URL** field, enter the complete URL of the server that you want to connect to. Must begin with either ``http://`` or ``https://``.
+5. Click **Add**.
```
esethna: Need to click save after?

source/help/apps/desktop-guide.rst
```diff
@@ -40,163 +41,62 @@ Editing servers and teams
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-If your server is hosted on ``https://``, insecure content, images and scripts with ``http://`` are not rendered by default. This option allows such content to be rendered. If you do enable it, please be mindful of potential security risks shared over ``http://`` protocols.
-
-This setting is on by default.
-
-Note: Enabling mixed content will disable YouTube video previews `due to an issue in the underlying technology <https://github.com/electron/electron/issues/2749>`_ used by the Mattermost Desktop app.
+    If your server is hosted on ``https://``, insecure content, images and scripts with ``http://`` are not rendered by default. This option allows such content to be rendered. If you do enable it, please be mindful of potential security risks shared over ``http://`` protocols.
+    
+    This setting is on by default.
```
esethna: Propose: on > enabled

Same in other settings

source/help/apps/desktop-guide.rst
```diff
@@ -40,163 +41,62 @@ Editing servers and teams
...
 
 Hide menu bar (Windows, Linux only)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Hides the menu bar. When enabled, pressing ALT will toggle whether the menu bar is shown or hidden.
 
-Hides the menu bar. When enabled, pressing ALT will toggle whether the menu bar is shown or hidden.
-
-This setting is off by default.
+    This setting is off by default.
```
esethna: Propose: off > disabled

Same in other settings

source/install/desktop.rst
```diff
@@ -71,3 +69,24 @@ To download the unofficial, community-driven .deb packages for the desktop appli
...
+
+Troubleshooting
+--------------------------------------------------
+
+Possible solutions to issues encountered when using the Desktop App.
+
+"Installation has failed" dialog
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The app data might be corrupted - remove all the files in `C:\Users...\AppData\Local\mattermost`, then try re-installing the app.
```
esethna: Propose: remove > delete

source/install/desktop.rst
```diff
@@ -71,3 +69,24 @@ To download the unofficial, community-driven .deb packages for the desktop appli
...
+
+Desktop App window is black and doesn't load the page
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Firt try to clear cache and reload the app from **View** > **Clear Cache and Reload** or by pressing CTRL/CMD+SHIFT+R.
```
esethna: firt > first

source/install/desktop.rst
```diff
@@ -71,3 +69,24 @@ To download the unofficial, community-driven .deb packages for the desktop appli
...
+
+Next, quit the app and re-start it to see if the issue clears.
```
esethna: I think we use "restart" as standard

source/install/desktop.rst
esethna: Maybe use bullet points here or indent?

source/help/apps/desktop-guide.rst
```diff
@@ -40,163 +42,63 @@ Editing servers and teams
...
 
-1. To the right of the team or server you want update, click **Edit**.
-2. Edit **Name** and/or **URL**.
-3. Click **Save**.
+1. On the menu bar, go to **File** > **Settings** or press CTRL/CMD+COMMA
+2. Next to the team or server you want update, click **Edit**.
+3. Edit **Name** and/or **URL**.
+4. Click **Save**.
 
 Removing servers and teams
```
esethna: These section titles should be in title case

source/install/desktop.rst
```diff
@@ -71,3 +69,23 @@ To download the unofficial, community-driven .deb packages for the desktop appli
 2. Open a terminal and execute the command ``sudo dpkg -i mattermost-desktop-<VERSION>-<ARCH>.deb``
 
 3. Open **Dash** (located at top left corner) and input ``mattermost``, then click the Mattermost icon. The Mattermost desktop should open automatically.
+
+Troubleshooting
+--------------------------------------------------
+
+Possible solutions to issues encountered when using the Desktop App.
+
+"Installation has failed" dialog
```
esethna: It seems like the headings are giving a lot of visual priority to errors. Do these need to be section headings or can we just use the same indented formatting as the setting names? 

image

esethna: It seems like the headings are giving a lot of visual priority to errors. Do these need to be section headings or can we just use the same indented formatting as the setting names? 

image

jasonblais: @JeffSchering thoughts? 

I used headings to follow our other docs, but not sure if we should add a principle about this in our style guides

JeffSchering: I think if you want the labels to show up in the ToC or be automatically included via the `.. contents::` directive, then you have no choice but to use headings. Otherwise, a table would probably be the best way, but tables in reST are really difficult to get right, so a definition list would suffice.

jasonblais: Cool, I changed this to a definition list.

We might want to visit other docs that use headers for troubleshooting section for consistency (and maybe add to doc guidelines?)

jasonblais:
Updated, feel  free to merge @JeffSchering if it looks okay

PR #792, opened 2017-01-16
Update changelog for v3.6.1

lfbrock:
(Wait until the final v3.6.1 is cut before merging)

### DIFF
--- a/source/administration/changelog.md
+++ b/source/administration/changelog.md
@@ -4,8 +4,6 @@ This changelog summarizes updates to [Mattermost Team Edition](http://www.matter
 
 ## Release v3.6.1  
 
-Release date: 2017-01-16
-
 ### Notes on Patch Release
 
  - **v3.6.1, released 2017-01-17**
--- a/source/administration/changelog.md
+++ b/source/administration/changelog.md
@@ -2,9 +2,16 @@
 
 This changelog summarizes updates to [Mattermost Team Edition](http://www.mattermost.org/), an open source team messaging solution released bi-monthly under an MIT license, and [Mattermost Enterprise Edition](https://about.mattermost.com/pricing/), a commercial upgrade offering enterprise messaging for large organizations.
 
-## Release v3.6.0  
+## Release v3.6.1  
 
 Release date: 2017-01-16
+
+### Notes on Patch Release
+
+ - **v3.6.1, released 2017-01-17**
+   - Fixed a performance regression when sending many notifications at once, upgrading only necessary if you notice a large delay when sending `@all` or `@channel` notifications in channels with a lot of users
+ - **v3.6.0, released 2017-01-16**
+   - Original 3.6 release.
 
 ### Security Update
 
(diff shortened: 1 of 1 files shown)


source/administration/changelog.md
```diff
@@ -2,9 +2,14 @@
...
 This changelog summarizes updates to [Mattermost Team Edition](http://www.mattermost.org/), an open source team messaging solution released bi-monthly under an MIT license, and [Mattermost Enterprise Edition](https://about.mattermost.com/pricing/), a commercial upgrade offering enterprise messaging for large organizations.
 
-## Release v3.6.0  
+## Release v3.6.1  
 
-Release date: 2017-01-16
+### Notes on Patch Release
+
+ - **v3.6.1, released 2017-01-17**
+   - Fixed a performance regression when sending many notifications at once, upgrading only necessary if you notice a large delay when sending `@all` or `@channel` notifications in channels with a lot of users
```
jasonblais: I thought the second part, `upgrading only necessary...`, was still referring to the fix itself.

Maybe we could break it into two bullet points so it's more obvious 3.6.1 upgrade is only necessary for the case you described:

```
 - **v3.6.1, released 2017-01-17**
   - Fixed a performance regression when sending many notifications at once
   - Note: Upgrading to Mattermost v3.6.1 only necessary if you notice a large delay when sending `@all` or `@channel` notifications in channels with a lot of users
```

esethna:
+1, after the final date of release is updated

source/administration/changelog.md
```diff
@@ -2,9 +2,17 @@
...
 
-## Release v3.6.0  
+## Release v3.6.1  
 
-Release date: 2017-01-16
+### Notes on Patch Release
+
+ - **v3.6.1, released 2017-01-17**
+   - Fixed a performance regression when sending many notifications at once (for example, when `@all` or `@channel` is used in a channel with many users)
+   - Fixed an issue where the config flag for the CLI was not backwards compatible
```
esethna: Should "config" be formatted? Propose `--config`

source/administration/changelog.md
```diff
@@ -2,9 +2,17 @@
...
+   - Fixed an upgrade issue for MySQL databases using certain character sets
```
esethna: Do we want to say which character sets?

source/administration/changelog.md
```diff
@@ -2,9 +2,17 @@
 
 This changelog summarizes updates to [Mattermost Team Edition](http://www.mattermost.org/), an open source team messaging solution released bi-monthly under an MIT license, and [Mattermost Enterprise Edition](https://about.mattermost.com/pricing/), a commercial upgrade offering enterprise messaging for large organizations.
 
-## Release v3.6.0  
+## Release v3.6.1  
 
-Release date: 2017-01-16
+### Notes on Patch Release
+
+ - **v3.6.1, released 2017-01-17**
```
esethna: date needs to be updated

source/administration/changelog.md
lfbrock: There's different alternatives for the config flag, including `--config` (new) and `-config` (old). I think it's appropriate to just reference it as the "config flag", unless we want to reword the whole sentence.

PR #796, opened 2017-01-18
Spec review guidelines

lfbrock:
Open to feedback. 

The spacing is a little weird with the subpoints (there's a break after them, see screenshot), but not sure if there's a way to format nicer. 

screenshot 2017-01-18 15 15 58

### DIFF
--- a/source/process/spec-process.rst
+++ b/source/process/spec-process.rst
@@ -30,13 +30,13 @@ This is a rough guideline for drafting and reviewing specs. It’s meant to be u
     - Pair reviews
     
       - Iterate on designs as needed
-      - Early check in with developers to identify any potential issues
+      - Early meetings with developers to discuss the design, and identify any potential issues
       
 3. **Share with team.**
 
     - Post draft with @channel in Spec Review channel, asking to review and add comments
     
-4. **Spec review.**
+4. **Initial spec review.**
 
     - Schedule a meeting with interested people to discuss comments and any other questions about the spec, try to include at least one member from each of the following teams:
     
--- a/source/process/spec-process.rst
+++ b/source/process/spec-process.rst
@@ -1,0 +1,46 @@
+Spec Review Process
+===================
+
+This is a rough guideline for the process for drafting and reviewing specs. It’s meant to be used as a starting point, but not all specs will follow this exact process - please use your judgment, as it’s something that will vary depending on the situation.
+
+1. Brainstorm / research.
+    - Research topic
+    - Research customer requirements
+    - Pair design brainstorm
+    - User workflows
+    - Consider options / vision for UI
+2. Initial draft.
+    - Goal / “Why"
+
+      - Link to customer data / feedback
+      - Link back to company objectives
+    - User Stories
+    - Rough and Quick Wireframes / Mock ups
+
+      - Flows of screens
+      - Focus should be on the mock ups, with text that supports design
+
+    - Questions
+    - “Not in scope”
+    - Success Metrics
+    - Pair reviews
+
+      - Iterate on designs as needed
+      - Early check in with developers to identify any potential issues
+3. Share with team.
+    - Post draft with @channel in Spec Review channel, asking to review and add comments.
+4. Spec review.
+    - Schedule a meeting with interested people to discuss comments and any other questions about the spec:
+
+      - UX Design
+      - Product Manager
+      - Developer
+      - Test
+5. Update draft.
+    - Update based on feedback. Update wireframes / mock ups.
+6. Review again.
+    - Post updated draft in Spec Review channel, asking the people who previously commented for another review.
+    - Share with interested customers for feedback
+7. Repeat reviews/updates.
+    - Repeat as needed until final draft is signed off on.**
+8. **Break into tickets with dev.**
--- a/source/process/spec-process.rst
+++ b/source/process/spec-process.rst
@@ -1,46 +1,61 @@
 Spec Review Process
 ===================
 
-This is a rough guideline for the process for drafting and reviewing specs. It’s meant to be used as a starting point, but not all specs will follow this exact process - please use your judgment, as it’s something that will vary depending on the situation.
+This is a rough guideline for drafting and reviewing specs. It’s meant to be used as a starting point, and not all specs will follow this exact process - please rely on your judgment, as not all steps may apply.
 
-1. Brainstorm / research.
+1. **Brainstorm / research.**
+
     - Research topic
     - Research customer requirements
     - Pair design brainstorm
     - User workflows
     - Consider options / vision for UI
-2. Initial draft.
-    - Goal / “Why"
+    
+2. **Initial draft.**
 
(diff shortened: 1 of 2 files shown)


source/process/spec-process.rst
```diff
@@ -0,0 +1,46 @@
...
+
+      - Link to customer data / feedback
+      - Link back to company objectives
+    - User Stories
+    - Rough and Quick Wireframes / Mock ups
+
+      - Flows of screens
+      - Focus should be on the mock ups, with text that supports design
+
+    - Questions
```
jasonblais:
```
  - Questions
  - “Not in scope”
  - Success Metrics
```

might not be clear for someone reading the doc for the first time

source/process/spec-process.rst
```diff
@@ -0,0 +1,46 @@
+Spec Review Process
+===================
+
+This is a rough guideline for the process for drafting and reviewing specs. It’s meant to be used as a starting point, but not all specs will follow this exact process - please use your judgment, as it’s something that will vary depending on the situation.
```
jasonblais: Maybe 

```
This is a rough guideline for drafting and reviewing specs. It’s meant to be used as a starting point, and not all specs will follow this exact process - please rely on your judgment, as not all steps may apply.
```

 - `for the process` seems unnecessary
 - `but` --> `and`
 - shortening the ending a little

source/process/spec-process.rst
```diff
@@ -0,0 +1,46 @@
...
+      - Early check in with developers to identify any potential issues
+3. Share with team.
+    - Post draft with @channel in Spec Review channel, asking to review and add comments.
+4. Spec review.
+    - Schedule a meeting with interested people to discuss comments and any other questions about the spec:
+
+      - UX Design
+      - Product Manager
+      - Developer
+      - Test
```
jasonblais: you mean QA/tester?

source/process/spec-process.rst
```diff
@@ -0,0 +1,46 @@
...
+    - Pair reviews
+
+      - Iterate on designs as needed
+      - Early check in with developers to identify any potential issues
+3. Share with team.
+    - Post draft with @channel in Spec Review channel, asking to review and add comments.
+4. Spec review.
+    - Schedule a meeting with interested people to discuss comments and any other questions about the spec:
+
+      - UX Design
```
jasonblais: UX Designer

source/process/spec-process.rst
```diff
@@ -0,0 +1,46 @@
...
+      - Product Manager
+      - Developer
+      - Test
+5. Update draft.
+    - Update based on feedback. Update wireframes / mock ups.
+6. Review again.
+    - Post updated draft in Spec Review channel, asking the people who previously commented for another review.
+    - Share with interested customers for feedback
+7. Repeat reviews/updates.
+    - Repeat as needed until final draft is signed off on.**
```
jasonblais: propose we have at least one person sign off from each of the groups (maybe a PM and a dev to start with)

source/process/spec-process.rst
lfbrock: I was using "UX Design" and "Test" more as someone from the UX Design / test teams as opposed to individuals, but we should be consistent on the list.

source/process/spec-process.rst
lfbrock: Yeah I think the idea is that it should be signed off on by all groups, but I can update to make that more explicit

source/process/spec-process.rst
```diff
@@ -0,0 +1,46 @@
+Spec Review Process
+===================
+
+This is a rough guideline for the process for drafting and reviewing specs. It’s meant to be used as a starting point, but not all specs will follow this exact process - please use your judgment, as it’s something that will vary depending on the situation.
+
+1. Brainstorm / research.
+    - Research topic
```
lfbrock: I don't think that's correct, at least in my local version it only bolds the point? Which is I think the formatting we wanted for lists?

But maybe @JeffSchering can comment on what the formatting should be, I'm 0/5. 

screenshot 2017-01-23 15 28 52

source/process/spec-process.rst
lfbrock: Updated to explain a bit more, left success metrics because I think that's pretty standard? 

This is for the core team handbook, so the audience would be mainly new hire product managers / ux designers.

source/process/spec-process.rst
```diff
@@ -0,0 +1,61 @@
...
+    - Consider options / vision for UI
+    
+2. **Initial draft.**
+
+    - Goal / “Why"
+    
+      - Link to customer data / feedback
+      - Link back to company objectives
+      
+    - User Stories
```
esethna: Can we expand on what this means?

source/process/spec-process.rst
```diff
@@ -0,0 +1,61 @@
...
+    - What is “Not in scope” for the spec (things that are purposefully left out, to be addressed at another time)
+    - Success Metrics
+    - Pair reviews
+    
+      - Iterate on designs as needed
+      - Early check in with developers to identify any potential issues
+      
+3. **Share with team.**
+
+    - Post draft with @channel in Spec Review channel, asking to review and add comments
```
esethna: Maybe link to the channel?

source/process/spec-process.rst
```diff
@@ -0,0 +1,61 @@
...
+    
+4. **Spec review.**
```
esethna: Maybe call this "Initial Review"

source/process/spec-process.rst
```diff
@@ -0,0 +1,61 @@
...
+      - UX Design
+      - PM
+      - Dev
+      - Test
+      
+5. **Update draft.**
+
+    - Update based on feedback, including wireframes / mock ups
+    
+6. **Review again.**
```
esethna: Then maybe call this "Second Review"

source/process/spec-process.rst
```diff
@@ -0,0 +1,61 @@
...
+
+    - Post updated draft in Spec Review channel, asking the people who previously commented for another review
+    - Share with interested customers for feedback
+    
+7. **Repeat reviews/updates.**
+
+    - Repeat as needed until final draft is signed off on by UX Design, PM, Dev, Test
+    
+8. **Break into tickets with dev.**
```
esethna: Maybe add a subpoint here on who owns this? ie does dev do this or PM with dev help?

source/process/spec-process.rst
lfbrock: Going to leave as is, it's a pretty well know / well defined term

source/process/spec-process.rst
lfbrock: I'm going to leave as is, because I think we're still figuring this out. Likely the dev will break it out into tickets, but I want PM involved as well to make sure they understand the tickets, that the tickets are a reasonable size, and that larger features are being worked on in a way that will make it easy to spin up a test server at the earliest stage possible, so we avoid having features "go dark".

PR #807, opened 2017-01-21
Single-sign-on --> Single sign-on typo fix

jasonblais:


jasonblais:
This probably just needs one reviewer to confirm the correct use of `single sign-on` (SSO)

JeffSchering:
Single sign-on seems to be the industry norm.

PR #809, opened 2017-01-24
Update nodejs install version for Ubuntu.

grundleborg:
The 5.x releases give a warning about being deprecated and no longer
getting security fixes. Current master works fine with 6.x, which is
still maintained.


jasonblais:
+1

Issue #785, opened 2017-01-13
Document updating system with HA


it33:
Lots of questions on how Mattermost applies update in HA mode. Propose we add something like the below to https://docs.mattermost.com/deployment/cluster.html#id13

This needs someone to draft the step-by-step instructions for the Update Guide at the bottom, and we definitely need tech review here. 

There are some open items to work on (Corey can provide more detail): 

1. Ability to add and remove nodes without requiring rolling cluster 
2. Ability to have code support multiple database schemas 
3. HTTP/2 between servers to hold websocket open when doing inter-server communication 

------

# High Availability (Beta) 

*Available in Enterprise Edition E20.*

High Availability support enables a Mattermost system to maintain service during outages and hardware failures through the use of redundant infrastructure. 

## Overview 

High availability in Mattermost consists of running redundant Mattermost application servers, redundant database servers and redundant load balancers so that the failure of any one of these components does not interrupt operation of the system. 

#### 1) Redundancy at anticipated scale. 

Redundant components (application server, database and load balancer) need to be sized and configured to carry the **full anticipated load of the system** in the event of an outage of other components. If this requirement is not met, an outage of one component may overload its backups, causing a system outage. 

#### 2) Continuous single-server updates 

A high availability Mattermost system can apply most configuration changes and dot release security updates without interrupting service by following a continuous single server update procedure: 

1. **Disconnect redundant servers** - Configure load balancers to consolidate requests to a single application server capable of supporting anticipated traffic during update procedure (depending on resources available, you may choose to run updates during low traffic periods). 
2. **Apply single server update** - Apply update to single server by updating `config.json` and running a command to load the new configuration into the application server, which continues to operate without down time. 
3. **Apply update to redundant servers** - Apply the same updates to the redundant servers. 
4. **Reconnect redundant servers** - Configure load balancers to once again balance traffic across all application servers. 

Step-by-step instructions for these procedures are available: 
- [Update system configuration changes while operating continuously](XXXXXX)
- [Update system with security patch dot release while operating continuously](XXXXXX) 

**Exception:** A brief period of maintenance downtime is required if you are changing any of the following: 

- **Configuration settings requiring server restart** (Default Server Language, Rate Limiting, Webserver Mode, Database or High Availability) - typically update and restart time is less than 5 seconds 
- **Database schema** - typically update and restart time is less than 30 seconds 

Step-by-step instructions for these procedures are available: 

1. [Update with database schema changes requiring maintenance downtime](XXXXXX) 
[...]

PR #825, opened 2017-01-30
Added instructions for updating HA clusters without service interruption

JeffSchering:
This PR implements Issue [Document updating system with HA #785](https://github.com/mattermost/docs/issues/785)


source/deployment/cluster.rst
```diff
@@ -1,116 +1,116 @@
...
-- Support for high availability configurations of the Mattermost application server is available in Mattermost Enterprise Edition E20 version 3.3 and higher. 
+.. contents::
+  :backlinks: top
+  :local:
+  
+Continuous Operation During Server Updates
+------------------------------------------
 
-This documentation provides a deployment guide, configuration and compatibility details, as well as troubleshooting advice.
+To ensure uninterrupted service during updates, you must make sure that the components are properly sized and that you follow the correct sequence for updating the system components.
```
it33: "ensure" is incorrect, should be something like "enable"--the step here is a necessary, but not sufficient condition.

source/deployment/cluster.rst
```diff
@@ -1,116 +1,116 @@
-High Availability (E20) - Beta
+High Availability Cluster (E20)
 ===============================
 
-*Available in Enterprise Edition E20.*
+The high availability cluster feature enables a Mattermost system to maintain service during outages and hardware failures through the use of redundant infrastructure. It is available only in Enterprise Edition E20.
```
it33: Let's keep *Available in Enterprise Edition E20* as a separate clear line below the title as the first item. 

@JeffSchering can we add something to style guide that Enterprise Edition features should have a note like this in documentation? We should eventually phase out the (E10), (E20) next to the title, as it doesn't work "(E10)" should technically be "(E10, E20)" and that breaks if we one day add a new SKU.

source/deployment/cluster.rst
```diff
@@ -1,116 +1,116 @@
...
 
+Each application server, database, and load balancer must be sized and configured to carry the full load of the system. If this requirement is not met, an outage of one component can result in an overload of the remaining components, causing a complete system outage.
 
-- `Deployment Guide <https://docs.mattermost.com/deployment/cluster.html#id1>`_ - **Instructions to set up and maintain high availability on your Mattermost servers.**
- - `Initial Setup Guide for High Availability <https://docs.mattermost.com/deployment/cluster.html#id2>`_
- - `Adding a Server to the Cluster <https://docs.mattermost.com/deployment/cluster.html#id4>`_
- - `Removing a Server from the Cluster <https://docs.mattermost.com/deployment/cluster.html#id9>`_
+You can apply most configuration changes and dot release security updates without interrupting service, provided that you update the system components in the correct sequence. Exceptions to this rule are changes to configuration settings that require a server restart, and updates that involve a change to the database schema. Downtime for a server restart is around 5 seconds, and for a database schema update, downtime is around 30 seconds.
```
it33: Too much text, we need documentation highly **scannable** 

@JeffSchering can we add something to style guide where a) we have "scannable" as a high level objective we're going for, and b) as a guideline, we don't have more than 2-3 sentences without a heading? IT admins today are talking to each other over SnapChat, attention spans aren't as long as they used to be. 

The points we need to make are: 

1. the two conditions needed for continuous operation, and 
2. there's some reasonable exceptions that do require downtime

Proposing we

AVOID BLOCKS OF TEXT: 

The three paragraphs in succession are too difficult to scan: 

image

LOOK WE'RE GOING FOR: 

Please try more headings, the following isn't a perfect example, but it has the rhythm: 

image

source/deployment/cluster.rst
```diff
@@ -134,75 +134,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
+8. Repeat the upgrade procedure for the server that was left running.
+
+Updating server with service interruption
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A service interruption is required when the update includes a change to the database schema or when a change to ``config.json`` requires a server restart, such as when making the following changes:
+
+  * Default Server Language
+  * Rate Limiting
+  * Webserver Wode
```
it33: Mode?

source/deployment/cluster.rst
```diff
@@ -134,75 +134,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
------
+Update Guide
+------------
+
+Updating configuration without service interruption
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A service interruption is not required for most configuration updates. See `Updating server with service interruption`_ for a list of configuration updates that require a service interruption.
+
+You can apply updates during a period of low load, but if your HA cluster is sized correctly, you can do it at any time. The system downtime is brief, and depends on the number of Mattermost servers in your cluster. Note that you are not restarting the machines, only the Mattermost server applications. A Mattermost server restart generally takes about 5 seconds.
```
it33: 1) Please check with dev, but I'm not sure this is true: 

>if your HA cluster is sized correctly, you can do it at any time

If you have 5 servers in HA and the system is correctly sized to handle failure of 2 servers, you still need 3 servers to handle throughput--which means you can't run continuous while updating. 

Our update only works when a single server can handle 100% of anticipated load during the update process. 

2) Please double-check with dev, not sure why we say this: 

>The system downtime is brief, and depends on the number of Mattermost servers in your cluster.

My understanding is there should be no downtime.

source/deployment/cluster.rst
```diff
@@ -134,75 +134,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
+4. Shut down Mattermost on all but one server.
+5. Reload the configuration file on the server that is still running. Go to **System Console > Configuration** and click **Reload Configuration from Disk**
+6. Start the other servers. 
+
+Updating server without service interruption
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A service interruption is not required for security patch dot releases of the Mattermost server.
+
+You can apply updates during a period of low load, but if your HA cluster is sized correctly, you can do it at any time. The system downtime is brief, and depends on the number of Mattermost servers in your cluster. Note that you are not restarting the machines, only the Mattermost server applications. A Mattermost server restart generally takes about 5 seconds.
```
it33: See above for feedback on this paragraph

source/deployment/cluster.rst
```diff
@@ -134,75 +134,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
+1. Review the upgrade procedure in the *Upgrade Enterprise Edition* section of :doc:`../administration/upgrade`.
+2. Make a backup of your existing ``config.json`` file.
+3. Set your proxy to move all new requests to a single server. If you are using NGINX and it's configured with an upstream backend section in ``/etc/nginx/sites-available/mattermost`` then comment out all but the one server that you intend to update first, and reload NGINX.
+4. Shut down Mattermost on each server except the one that you are updating first.
+5. Upgrade each Mattermost instance that is shut down.
+6. On each server, replace the new ``config.json`` file with your backed-up copy.
+7. Start the Mattermost servers.
+8. Repeat the upgrade procedure for the server that was left running.
+
+Updating server with service interruption
```
it33: Propose renaming this title and two above it with something more focused on "continuous operation", which should be the theme of this document 

Proposed titles: 

- Updating configuration changes while operating continuously
- Updating server version while operating continuously
- Updating system with configuration change requiring scheduled maintenance downtime

@JeffSchering also, propose we add to Style Guide: 

1. an objective of accurately framing documentation in the context of what is to be achieved, 
2. examples of terminology in the context of achievement 

Examples of terminology in the context of achievement: 

- Use "with continuous operation" to be clear on the desired state and avoid "without service interruption" as a double negative definition.  
- Use "with scheduled maintenance downtime" instead of "with service interruption", to make it clear the system should run under IT best practices. 

We should also be clear terminology in the context of achievement should always be accurate and should avoid unintentionally overselling a feature, 

Please see if we can use the [example above on what HA offers](https://github.com/mattermost/docs/pull/825/files#r98596393) as hyperbolic illustration of a correction?

source/deployment/cluster.rst
```diff
@@ -134,75 +134,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
+
+  * Default Server Language
+  * Rate Limiting
+  * Webserver Wode
+  * Database
+  * High Availability
+
+If the update includes a change to the database schema, the database is upgraded by the first server that starts.
+
+Apply updates during a period of low load. The system downtime is brief, and depends on the number of Mattermost servers in your cluster. Note that you are not restarting the machines, only the Mattermost server applications.
```
it33: Please check with @crspeller to see if there is an alternate procedure to make the change with less downtime by consolidating load to one server, having only a brief outage to update it, then going back to full HA configuration, similar to the sections above where we operate continuously. 

It's more work, but it's less downtime, so propose offering it as an alternative procedure.

source/deployment/cluster.rst
```diff
@@ -1,116 +1,116 @@
...
+  :backlinks: top
+  :local:
+  
+Continuous Operation During Server Updates
+------------------------------------------
 
-This documentation provides a deployment guide, configuration and compatibility details, as well as troubleshooting advice.
+To ensure uninterrupted service during updates, you must make sure that the components are properly sized and that you follow the correct sequence for updating the system components.
 
+Each application server, database, and load balancer must be sized and configured to carry the full load of the system. If this requirement is not met, an outage of one component can result in an overload of the remaining components, causing a complete system outage.
```
coreyhulen: Maybe "Upon failure the remaining application servers, database servers..."

source/deployment/cluster.rst
coreyhulen: Mabye "and for a database schema update, downtime can be up to 30 seconds."

source/deployment/cluster.rst
```diff
@@ -1,116 +1,116 @@
...
- - `Database <https://docs.mattermost.com/deployment/cluster.html#database-configuration>`_ - Size and deploy a multi-database configuration for high availability and scaling.
+The following outline describes the procedure for updating a high availability system without a service interruption.
 
-- `Troubleshooting <https://docs.mattermost.com/deployment/cluster.html#id14>`_ - **Advice on troubleshooting your high availability setup.**
-
------
+1. **Disconnect redundant servers** - Configure load balancers to consolidate requests to a single application server capable of supporting anticipated traffic during the update process. Depending on the resources available, you might choose to run updates during low traffic periods.
+2. **Apply single server update** - Apply the update to a single server by updating ``config.json`` and running a command to load the new configuration into the application server, which continues to operate without down time.
+3. **Apply update to redundant servers** - Apply the same updates to the redundant servers.
+4. **Reconnect redundant servers** - Configure load balancers to once again balance traffic across all application servers.
```
coreyhulen: Maybe also mention a rolling configuration update.

source/deployment/cluster.rst
```diff
@@ -134,75 +134,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
 
 File Storage Configuration
-----------------------------------------
+^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. note:: File storage is assumed to be shared between all the machines utilizing services such as NAS or Amazon S3. If ``"DriverName": "local"`` is used then the directory at ``"FileSettings":`` ``"Directory": "./data/"`` is expected to be a NAS location mapped as a local directory, otherwise high availability will not function correctly and may corrupt your file storage. If you’re using Amazon S3 for file storage then no other configuration is required.
+.. note::
+  1. File storage is assumed to be shared between all the machines that are using services such as NAS or Amazon S3.
+  2. If ``"DriverName": "local"`` is used then the directory at ``"FileSettings":`` ``"Directory": "./data/"`` is expected to be a NAS location mapped as a local directory, otherwise high availability will not function correctly and may corrupt your file storage.
+  3. If you’re using Amazon S3 for file storage then no other configuration is required.
```
coreyhulen: Maybe "Amazon S3, or Minio for file storage..."

source/deployment/cluster.rst
```diff
@@ -134,75 +134,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
+If you’re using the Compliance Reports feature in Enterprise Edition E20, you need to configure the  ``"ComplianceSettings":`` ``"Directory": "./data/",`` to share between all machines or the reports will only be available from the System Console on the local Mattermost server.
 
 Migrating to NAS or S3 from local storage is beyond the scope of this document.
 
 Database Configuration
-------------------------------------
-Scaling the database can be accomplished by utilizing the read-replica feature. The Mattermost server can be set up to use one "master" database and up to 8 read replica databases. Mattermost distributes read requests across all databases, and sends write requests to the master database, and those changes are then sent to update the read replicas. 
+^^^^^^^^^^^^^^^^^^^^^^
+
+Use the read-replica feature to scale the database. The Mattermost server can be set up to use one "master" database and up to 8 read replica databases. Mattermost distributes read requests across all databases, and sends write requests to the master database, and those changes are then sent to update the read replicas.
```
coreyhulen: "and up to 8 read replica" is a limit of Aroua (I think).  Maybe "and multiple read replica"

source/deployment/cluster.rst
```diff
@@ -134,75 +134,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
+2. Go to **System Console > Database** and click **Recycle Database Connections** to take down existing database connections and set up new connections in the multi-database configuration.
 
-Manual failover for master database  
-`````````````````````````````````````````````````````````````````
-If the need arises to switch from the current master database--for example, if it is running out of disk space, or requires maintenance updates, or for other reasons--the Mattermost server can switch to using one of its read replicas as a master database by updating ``DataSource`` in ``config.json``. The following procedure can be used to apply the settings without shutting down the Mattermost server: 
+While the connection settings are changing, there might be a brief moment when writes to the master database are unsuccessful. The process waits for all existing connections to finish and starts serving new requests with the new connections. End users attempting to send messages while the switch is happening will have an experience similar to losing connection to the Mattermost server.
```
coreyhulen: instead of " will have an experience" maybe " may have an experience"

source/deployment/cluster.rst
```diff
@@ -134,75 +134,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
-- **Server is down**: If an inter-node communication fails to send a message it will attempt again in 15 seconds.  If the second attempt fails, the server is assumed to be down. An error message is written to the logs and the System Console will show a status of red for that server.
+- **Server version mismatch**: Mattermost will still attempt the inter-node communication, but the System Console will show a red status for the server since the high availability feature assumes the same version of Mattermost is installed on each server in the cluster. It is recommended to use the `latest version of Mattermost <https://www.mattermost.org/download/>`_ on all servers. Follow the upgrade procedure in :doc:`../administration/upgrade` for any server that needs to be upgraded.
+- **Server is down**: If an inter-node communication fails to send a message it will attempt again in 15 seconds. If the second attempt fails, the server is assumed to be down. An error message is written to the logs and the System Console will show a status of red for that server.
```
coreyhulen: The system will continue to ping the down system every 15 seconds and if successful send any new messages.  It will only attempt to send a messages twice before dropping it.

source/deployment/cluster.rst
JeffSchering: I asked Christopher about this while creating the first draft. He said stopping NGINX is required to make sure that all the clients disconnect.

>Its so the clients don't get confused during the upgrade. Stopping NGINX just makes sure all the clients disconnect.

source/deployment/cluster.rst
JeffSchering: Mode.

source/deployment/cluster.rst
JeffSchering: For the third title, is **Updating during scheduled maintenance downtime** ok? The other one is quite long and can mess up the ToC. Or even **Updating during maintenance downtime**?

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: I kept "without service interruption" in a few places. I think there are two mindsets. One is "I want to keep operating continuously" and the other is "I don't want to interrupt the service". I'm attempting to address both. Also, sometimes a negative construct acts as a warning that causes people to pay closer attention to the content, resulting in higher levels of accomplishment.

source/deployment/cluster.rst
JeffSchering: @it33 Should we document only the rolling config update? It seems faster, requires less intervention (ie, no consolidation to a single server), and (it seems to me at least) makes sizing less critical because you're not taking down all-but-one server.

@coreyhulen Are there configuration updates when a rolling update is not feasible?

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: I'll change it to "and multiple read replica". According to http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Limits.html, the limit on Aurora is 5.

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: Just so I'm clear:

1. System A goes down.
1. System B attempts to send a message to System A, but fails
1. System B sends another message at the regularly scheduled time (15 seconds later) but fails again
1. System B marks System A as down and sets its status to red
1. System B continues to ping System A in 15 second intervals
1. When System A comes back, System B sends any messages that it can

Is that correct?

source/deployment/cluster.rst
```diff
@@ -134,75 +144,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
 Sizing databases
-```````````````````````````````````````
-Please see `documentation on sizing database servers <http://docs.mattermost.com/install/requirements.html#hardware-requirements>`_ for guidance to determine appropriate hardware. 
+````````````````
+
+For information about sizing database servers, see the *Hardware Sizing for Enterprise Deployments (Multi-Server)* section of :doc:`../install/requirements`.
 
 In a master/slave environment, make sure to size the slave machine to take 100% of the load in the event that the master machine goes down and you need to fail over.
 
+Deploying a multi-database configuration
```
esethna: Section titles should be sentence case throughout the doc

source/deployment/cluster.rst
```diff
@@ -1,116 +1,126 @@
...
 
-This documentation provides a deployment guide, configuration and compatibility details, as well as troubleshooting advice.
+.. contents::
+  :backlinks: top
+  :local:
+  
+Continuous Operation During Server Updates
+------------------------------------------
 
+To enable continuous operation during updates, you must make sure that the components are properly sized and that you follow the correct sequence for updating each of the system's components.
```
esethna: What is the correct sequence? Maybe link to the section describing this?

source/deployment/cluster.rst
```diff
@@ -1,116 +1,126 @@
...
-- Support for high availability database configurations is available in Mattermost Enterprise Edition E20 version 3.1 and higher. 
-- Support for high availability configurations of the Mattermost application server is available in Mattermost Enterprise Edition E20 version 3.3 and higher. 
+High availability in Mattermost consists of running redundant Mattermost application servers, redundant database servers, and redundant load balancers. The failure of any one of these components does not interrupt operation of the system.
 
-This documentation provides a deployment guide, configuration and compatibility details, as well as troubleshooting advice.
+.. contents::
+  :backlinks: top
+  :local:
+  
+Continuous Operation During Server Updates
```
esethna: Propose calling this section "Continuous Operation Overview" and laying out the requirements for continuous operation similar to how Ian had proposed. Right now it doesn't feel like there is a good flow to this section.

image

source/deployment/cluster.rst
```diff
@@ -1,116 +1,126 @@
...
+In some cases, updates must be applied during a maintenance window
+  Changes to configuration settings that require a server restart, and updates that involve a change to the database schema require a short period of downtime. Downtime for a server restart is around 5 seconds, and for a database schema update, downtime can be up to 30 seconds.
 
------
+Outline of the process
+~~~~~~~~~~~~~~~~~~~~~~
 
+The following outline describes the procedure for updating a high availability system while maintaining continuous operation of the service.
+
+1. **Disconnect redundant servers** - Configure load balancers to consolidate requests to a single application server capable of supporting anticipated traffic during the update process. You might choose to run updates during low traffic periods or during a scheduled maintenance window.
```
esethna: 0/5 could this info be consolidated in the "Update guide" section, rather than having two areas talking about update instructions?

source/deployment/cluster.rst
```diff
@@ -1,116 +1,126 @@
...
+2. **Apply single server update** - Apply the update to a single server by updating ``config.json`` and then loading the new configuration into the application server, which continues to operate without downtime.
+3. **Apply update to redundant servers** - Apply the same updates to the redundant servers.
+4. **Reconnect redundant servers** - Configure load balancers to once again balance traffic across all application servers.
+
+For specific instructions, see the `Update Guide`_
```
esethna: General note: Should we be consistent on using the terms update vs upgrade. It seems like we use them interchangably within this document. It seems like there are **updates** to configs, and **upgrades** to new versions. 0/5 I'm not a dev so I might be lacking technical knowledge here.

source/deployment/cluster.rst
```diff
@@ -134,75 +144,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
+You can apply updates during a period of low load, but if your HA cluster is sized correctly, you can do it at any time. The system downtime is brief, and depends on the number of Mattermost servers in your cluster. Note that you are not restarting the machines, only the Mattermost server applications. A Mattermost server restart generally takes about 5 seconds.
+
+1. Make a backup of your existing ``config.json`` file.
+2. For one of the Mattermost servers, make the configuration changes to ``config.json`` and save the file. Do not reload the file yet.
+3. Copy the ``config.json`` file to the other servers.
+4. Shut down Mattermost on all but one server.
+5. Reload the configuration file on the server that is still running. Go to **System Console > Configuration** and click **Reload Configuration from Disk**
+6. Start the other servers. 
+
+Updating server version while operating continuously
```
esethna: Propose changing to "Upgrading server version..."

source/deployment/cluster.rst
```diff
@@ -134,75 +144,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
+1. Review the upgrade procedure in the *Upgrade Enterprise Edition* section of :doc:`../administration/upgrade`.
+2. Make a backup of your existing ``config.json`` file.
+3. Set your proxy to move all new requests to a single server. If you are using NGINX and it's configured with an upstream backend section in ``/etc/nginx/sites-available/mattermost`` then comment out all but the one server that you intend to update first, and reload NGINX.
+4. Shut down Mattermost on each server except the one that you are updating first.
+5. Upgrade each Mattermost instance that is shut down.
+6. On each server, replace the new ``config.json`` file with your backed-up copy.
+7. Start the Mattermost servers.
+8. Repeat the upgrade procedure for the server that was left running.
+
+Updating during scheduled maintenance downtime
```
esethna: Propose: "Updates requiring service interruption"

source/deployment/cluster.rst
```diff
@@ -134,75 +144,143 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A service interruption is required when the update includes a change to the database schema or when a change to ``config.json`` requires a server restart, such as when making the following changes:
+
+  * Default Server Language
```
esethna: Should we link these to the corresponding config settings doc section?

source/deployment/cluster.rst
```diff
@@ -1,116 +1,126 @@
-High Availability (E20) - Beta
+High Availability Cluster (E20)
 ===============================
 
 *Available in Enterprise Edition E20.*
 
-High Availability support enables a Mattermost server to scale and maintain service during outages and hardware failures through the use of redundant infrastructure. 
+The high availability cluster feature enables a Mattermost system to maintain service during outages and hardware failures through the use of redundant infrastructure.
```
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: I believe the intent of this section is to point out that continuous operation can be achieved if the system is properly sized and the updates are applied in the proper order; and that in some cases continuous operation is not possible. What follows is a brief outline of the process that experienced and knowledgeable admins can scan then implement on their systems. Admins with less experience can use the links to more specific instructions in the Upgrade Guide further down the page.

source/deployment/cluster.rst
```diff
@@ -1,116 +1,126 @@
...
-2. Setup a new Mattermost server following one of our **Install Guides**. This server must use an identical copy of the configuration file, ``config.json``. Verify the server is functioning by hitting the private IP address.
-3. Modify the ``config.json`` files on all servers with the ``ClusterSettings`` as `described in this documentation <https://docs.mattermost.com/administration/config-settings.html#high-availability-beta>`_. Be sure to add the new server URL to ``InterNodeUrls``. 
-4. Verify the configuration files are identical on all servers, then restart each machine in the cluster in sequence with 10 or more seconds between each restart.
-5. `Modify your NGINX setup <https://docs.mattermost.com/deployment/cluster.html#proxy-server-configuration>`_ to add the new server.
-6. Open the **System Console** > **Advanced** > **High Availability (Beta)** to verify all the machines in the cluster are communicating as expected with green status indicators. If not, investigate the log files for any extra information.
+1. Backup your Mattermost database and the file storage location.
```
JeffSchering: The links that were there for the database backup and the file storage location didn't go to information about backing up databases or files so I removed them.

source/deployment/cluster.rst
```diff
@@ -1,116 +1,126 @@
...
+2. Setup a new Mattermost server. This server must use an identical copy of the configuration file, ``config.json``. Verify the server is functioning by hitting the private IP address.
+3. Modify the ``config.json`` files on all servers with the ``ClusterSettings`` as described in the *High Availability (Beta)* section of :doc:`../administration/config-settings`. Be sure to add the new server URL to ``InterNodeUrls``.
```
JeffSchering: That can't be done in Sphinx. It can't link to anchors in other documents, only the document. I can experiment with the :ref: role to see if it works, but that means revising the target docs too.

source/deployment/cluster.rst
it33: We should make a call and put in Style Guide, in my mind if we're doing a dot release it's an update and a major release it is an upgrade and this is about a dot release

JeffSchering: If we use the following definittion:
```
An update is an incremental change to Mattermost server that fixes bugs 
or performance issues. An upgrade adds new or improved functionality 
to the server.
```
Would it be safe to say that an update *never* requires downtime, and an upgrade *might* require downtime?

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: How about we use the following definition for this doc:
```
An update is an incremental change to Mattermost server that fixes bugs 
or performance issues. An upgrade adds new or improved functionality 
to the server.
```

source/deployment/cluster.rst
JeffSchering: Ok.

I assume you meant title case :smiley:

source/deployment/cluster.rst
JeffSchering: I changed it to "Upgrading Server Version While Operating Continuously"

source/deployment/cluster.rst
JeffSchering: Probably ok to not link. The people doing the upgrades have either changed config.json already and are familiar with the settings, or know that config.json was not changed.

I think links to other docs need to add significant value because they are not resilient in the sense that they can get unwittingly broken by changes to the target doc. A broken link is often worse than no link.

source/deployment/cluster.rst
JeffSchering: Huh. I made the change, but for some reason it's not showing up here. You can see it in commit https://github.com/mattermost/docs/pull/825/commits/9504dfa7a9658e6a540f7aab81865078c7e782a7, line 205.

source/deployment/cluster.rst
```diff
@@ -1,116 +1,114 @@
...
 
-1. Follow our `upgrade guide <https://docs.mattermost.com/administration/upgrade.html>`_ to upgrade your Mattermost server to v3.3 or later. 
-2. Setup a new Mattermost server with v3.3 or later following one of our **Install Guides**. This server must use an identical copy of the configuration file, ``config.json``. Verify the servers are functioning by hitting each independent server through it’s private IP address.
-3. Modify the ``config.json`` files on both servers to add the ``ClusterSettings`` as `described in this documentation <https://docs.mattermost.com/administration/config-settings.html#id58>`_. 
+1. Follow our :doc:`../administration/upgrade` to upgrade your Mattermost server to v3.3 or later. 
+2. Setup a new Mattermost server with v3.3 or later following one of our **Install Guides**. This server must use an identical copy of the configuration file, ``config.json``. Verify the servers are functioning by hitting each independent server through its private IP address.
```
lfbrock: "Setup" should be "Set up"

source/deployment/cluster.rst
```diff
@@ -1,116 +1,114 @@
...
+3. Modify the ``config.json`` files on both servers to add the ``ClusterSettings`` as described in the *High Availability (Beta)* section of :doc:`../administration/config-settings`.
 4. Verify the configuration files are identical on both servers then restart each machine in the cluster.
-5. `Modify your NGINX setup <https://docs.mattermost.com/deployment/cluster.html#proxy-server-configuration>`_ so it proxies to both servers.
-6. Open the **System Console** > **Advanced** > **High Availability (Beta)** to verify all the machines in the cluster are communicating as expected with green status indicators. If not, investigate the log files for any extra information.
+5. Modify your NGINX setup  so that it proxies to both servers. For more information about this, see `Proxy Server Configuration`_
```
lfbrock: Missing a period at end of sentence

source/deployment/cluster.rst
```diff
@@ -1,116 +1,114 @@
...
-3. Modify the ``config.json`` files on all servers with the ``ClusterSettings`` as `described in this documentation <https://docs.mattermost.com/administration/config-settings.html#high-availability-beta>`_. Be sure to add the new server URL to ``InterNodeUrls``. 
-4. Verify the configuration files are identical on all servers, then restart each machine in the cluster in sequence with 10 or more seconds between each restart.
-5. `Modify your NGINX setup <https://docs.mattermost.com/deployment/cluster.html#proxy-server-configuration>`_ to add the new server.
-6. Open the **System Console** > **Advanced** > **High Availability (Beta)** to verify all the machines in the cluster are communicating as expected with green status indicators. If not, investigate the log files for any extra information.
+1. Backup your Mattermost database and the file storage location.
+2. Setup a new Mattermost server. This server must use an identical copy of the configuration file, ``config.json``. Verify the server is functioning by hitting the private IP address.
```
lfbrock: "Setup" > "Set up"

source/deployment/cluster.rst
lfbrock: Can we check if we have other info about backing up database / file storage location somewhere? Maybe in the other upgrade guide? I think I remember we added some more details about that somewhere

source/deployment/cluster.rst
lfbrock: I guess we're trying to avoid doing a direct link in case titles change and the links break? 

Eg https://docs.mattermost.com/administration/config-settings.html?highlight=high%20availability#high-availability-beta

At worst, if the titles change I think it just takes you to the top of the document, so it might be better than nothing

source/deployment/cluster.rst
```diff
@@ -134,75 +132,147 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
 
 The new settings can be applied by either stopping and starting the server, or by loading the configuration settings as described in the next section. 
 
-Once loaded, database write requests will be sent to the master database and read requests will be distributed among the other databases in the list.
+Once loaded, database write requests are sent to the master database and read requests are distributed among the other databases in the list.
 
-Loading a multi-database configuration onto an active server
-``````````````````````````````````````````````````````````````````````````````````````````````````
-After a multi-database configuration has been defined in ``config.json`` the following procedure can be used to apply the settings without shutting down the Mattermost server: 
+Loading a Multi-database Configuration Onto an Active Server
```
lfbrock: "Onto" should probably be "onto" for title case

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
```diff
@@ -1,116 +1,115 @@
...
- - `File Storage Location <https://docs.mattermost.com/deployment/cluster.html#file-storage-configuration>`_ - Configure the file storage location to be compatible with a high availability setup.
- - `Database <https://docs.mattermost.com/deployment/cluster.html#database-configuration>`_ - Size and deploy a multi-database configuration for high availability and scaling.
-
-- `Troubleshooting <https://docs.mattermost.com/deployment/cluster.html#id14>`_ - **Advice on troubleshooting your high availability setup.**
-
------
+Continuous single-server update sequence
+  You can apply most configuration changes and dot release security updates without interrupting service, provided that you update the system components in the correct sequence. See the `Upgrade Guide`_ for instructions on how to do this.
 
+In some cases, updates must be applied during a maintenance window
```
esethna: Can we change this to: 

`Exception: Some updates require maintenance downtime

source/deployment/cluster.rst
```diff
@@ -134,75 +133,147 @@ A sample configuration for NGINX is provided below.  It assumes you have two Mat
...
+1. Review the upgrade procedure in the *Upgrade Enterprise Edition* section of :doc:`../administration/upgrade`.
+2. Make a backup of your existing ``config.json`` file.
+3. Set your proxy to move all new requests to a single server. If you are using NGINX and it's configured with an upstream backend section in ``/etc/nginx/sites-available/mattermost`` then comment out all but the one server that you intend to update first, and reload NGINX.
+4. Shut down Mattermost on each server except the one that you are updating first.
+5. Update each Mattermost instance that is shut down.
+6. On each server, replace the new ``config.json`` file with your backed-up copy.
+7. Start the Mattermost servers.
+8. Repeat the update procedure for the server that was left running.
+
+Upgrading Server Version While Operating Continuously
```
esethna: Propose calling this section:

`Server Upgrades Requiring Service Interruption`

esethna: The title now is too similar to the one above only changing "Update/Upgrade"

source/deployment/cluster.rst
```diff
@@ -1,116 +1,115 @@
...
- - `Database <https://docs.mattermost.com/deployment/cluster.html#database-configuration>`_ - Size and deploy a multi-database configuration for high availability and scaling.
-
-- `Troubleshooting <https://docs.mattermost.com/deployment/cluster.html#id14>`_ - **Advice on troubleshooting your high availability setup.**
-
------
+Continuous single-server update sequence
+  You can apply most configuration changes and dot release security updates without interrupting service, provided that you update the system components in the correct sequence. See the `Upgrade Guide`_ for instructions on how to do this.
 
+In some cases, updates must be applied during a maintenance window
+  Changes to configuration settings that require a server restart, and updates that involve a change to the database schema require a short period of downtime. Downtime for a server restart is around 5 seconds, and for a database schema update, downtime can be up to 30 seconds.
```
esethna: `... and updates that involve a change to the database...` > `... and server version upgrades that involve a change to the database...`

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: ok

source/deployment/cluster.rst
JeffSchering: ok

it33:
1) `Continuous single-server update sequence` could be simplified 

I think I wrote this, and it's not really meeting the [Thor test](https://docs.mattermost.com/guides/community.html?highlight=thor#testing-our-documentation), Propose `Use of update procedures for continuous operation` 

image

2) Propose changing the formatting around the EXCEPTION in screenshot below

Visually it looks like 3 requirements, but it's really 2 requirements and an exception

image

3) Add note to Style Guide that exceptions should be disabled as a subordinate point, rather than a sibling point, to the topic they reference. 

Propose using 2) above as a "before" screenshot and have an "after" screenshot on how exceptions should be formatted--clear, but subordinate to headings they reference and not siblings. 

4) Add a checklist item to Style Guide to confirm consistency of capitalization in TOC for new docs

I'm not clear what the capitalization should be here: 

image

it33:
Thanks @JeffSchering https://github.com/mattermost/docs/pull/825/commits/236cc7c7493773e90e11bc719e1397c490c9ffaf changes look good

PR #848, opened 2017-02-07
Slack import when channel name is deleted

JeffSchering:
This is the doc change for [PLT-4378 Slack import when channel name is deleted #4649](https://github.com/mattermost/platform/pull/4649).

The following item is added to the Limitations section:

> If the handle of an imported Slack channel is the same handle as a deleted Mattermost channel, then a random handle is generated for the imported Slack channel.

Also includes doc changes for the following Issues:
+ [PLT-495 Improves slack markup conversion #4914](https://github.com/mattermost/platform/pull/4914)
+ [PLT-4845: Slack import of channel purpose changed messages. #5110](https://github.com/mattermost/platform/pull/5110)

### DIFF
--- a/source/administration/migrating.rst
+++ b/source/administration/migrating.rst
@@ -1,35 +1,35 @@
-Migration Guide 
----------------
+Migration Guide
+===============
 
-Migrating the Mattermost Server  
-===============================
+Migrating the Mattermost Server
+-------------------------------
 
-The following instructions migrate Mattermost from one server to another by backing up and restoring the Mattermost database and ``config.json`` file. For these instructions **SOURCE** refers to the Mattermost server *from which* your system will be migrated and **DESTINATION** refers to the Mattermost server *to which* your system will be migrated. 
+The following instructions migrate Mattermost from one server to another by backing up and restoring the Mattermost database and ``config.json`` file. For these instructions **SOURCE** refers to the Mattermost server *from which* your system will be migrated and **DESTINATION** refers to the Mattermost server *to which* your system will be migrated.
 
-1. Backup your SOURCE Mattermost server 
+1. Backup your SOURCE Mattermost server
     1. See `Backup Guide <https://docs.mattermost.com/administration/backup.html>`_
-2. Upgrade your SOURCE Mattermost server to the latest major build version 
+2. Upgrade your SOURCE Mattermost server to the latest major build version
     1. See `Mattermost Upgrade Guide <upgrade.html>`_
-3. Install the latest major build of Mattermost server as your DESTINATION   
+3. Install the latest major build of Mattermost server as your DESTINATION
     1. See docs.mattermost.com for install guides. Make sure your new instance is properly configured and tested. The database type (MySQL or PostgreSQL) and version of SOURCE and DESTINATION deployments need to match.
     2. Stop the DESTINATION server using `sudo stop mattermost`, then backup the database and `config.json` file.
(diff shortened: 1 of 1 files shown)


source/administration/migrating.rst
```diff
@@ -48,13 +48,13 @@ This feature can be accessed through the `Mattermost Web App <https://docs.matte
...
 Supported Features
-++++++++++++++++++
+~~~~~~~~~~~~~~~~~~
 
 The following key features are supported when importing from Slack:
 
 * User accounts with an email address set
 
-* Public channels and the text messages posted in them
+* Public channels and the text messages posted in them, with formatting
```
lfbrock: Does all formatting work? Slack uses a different markdown than us, so it might not be converted for all things

source/administration/migrating.rst
```diff
@@ -64,19 +64,23 @@ Messages with file attachments are imported as a message containing a link to Sl
 
 Bot and Integration messages are imported by default, but if you would like them to display with the appropriate username when imported, you should ensure that `Enable Integrations to Override Usernames <https://docs.mattermost.com/administration/config-settings.html#enable-integrations-to-override-usernames>`_ is set in **System Console > Integrations > Custom Integrations** *before* doing the import.
 
+When Slack messages that are topic change messages, purpose change messages, and channel name change messages are imported, they appear in Mattermost as posts from the System user.
```
lfbrock: Can we condense this sentence a bit? Maybe something like "When Slack messages about changes to the topic, purpose, or channel name are imported, they appear in Mattermost as posts from the System user."

source/administration/migrating.rst
JeffSchering: My understanding from reading the comments on [PLT-495 Improves slack markup conversion #4914](https://github.com/mattermost/platform/pull/4914) is that all formatting works, but @esethna did the testing and can confirm one way or the other.

source/administration/migrating.rst
JeffSchering: I'd prefer to keep it as is for the reason that although it's longer and not the world's greatest prose, it is very precise about which Slack artifacts appear as posts from the System user.

How about this, which is a bit shorter:

*When topic-change messages, purpose-change messages, and channel name-change messages are imported from Slack, they appear in Mattermost as posts from the System user.*

lfbrock: Yep that sounds good to me.

source/administration/migrating.rst
esethna: According to their help article [here](https://get.slack.help/hc/en-us/articles/202288908-Format-your-messages), we now fully support their markdown for import.

source/administration/migrating.rst
esethna: This still needs to be updated to your proposed sentence, +1 otherwise.

source/administration/migrating.rst
lfbrock: @esethna they use different nonstandard markdown to do the formatting though. I'll see if Linda can help test.

esethna: This PR https://github.com/mattermost/platform/pull/4914 translates their non-standard markdown to our formatting. It would still be good to test though.

lfbrock: Ah okay perfect thanks Eric, I didn't realize we had a fix submitted for that since that time I looked at Slack import.

lfbrock:
+1

PR #872, opened 2017-02-15
Draft of developer toolkit documentation

it33:
Creating to request comments from community


lfbrock:
I added notes from devs so I think we can review this now, one question I have is should this be discoverable somehow from table of contents? I'm not sure what the plan is for this type of doc / where it should go, I'm not sure the developer folder is the appropriate place if we're going to be doing this for other features in the future.

jasonblais:
Reviewed, looks good overall.

Chatted with Joram and Ian, and we might also include some screenshots to help visualize what the toolkit can offer.

PR #865, opened 2017-02-14
Update of Style Guide

JeffSchering:
Updated with info from Issue https://github.com/mattermost/docs/issues/772 and https://github.com/mattermost/docs/issues/842

PR #874, opened 2017-02-15
PLT-5452 Add optional step to install libpng to developer setup

hmhealey:


esethna:
@JeffSchering do we have a guideline on using the indented formatting for troubleshooting sections or similarly formatted docs? If not I think we should add it, this section of the doc is hard to read without indenting:

image

JeffSchering:
@esethna We don't have a guideline specific to troubleshooting sections, but there are general recommendations in the proposed style guide update: https://github.com/mattermost/docs/pull/865/files#diff-d01072906218ee7de592c0b7ff0981e1

PR #897, opened 2017-02-19
Staff Developers

it33:
Drafting info about core & staff developers to aid in recruiting process


it33:
Merging to answer hiring process question

PR #907, opened 2017-02-21
Updated theme examples format and converted to .rst

JeffSchering:
Also converted the field descriptions to a definition list, to close off [Style field documentations should probably be in a table](https://mattermost.atlassian.net/browse/PLT-2830), which called for a table instead of a definition list. A table probably would have been better, but the current CSS on docs.mattermost.com clips wide tables so that horizontal scrolling is required in order to see the whole thing.


source/help/settings/theme-colors.rst
```diff
@@ -0,0 +1,138 @@
+.. _theme-colors:
+
+Theme Colors
+============
+
+The colors of the Mattermost user interface are customizable in **Account Settings > Display > Theme**. You can import your theme colors from Slack, customize the colors yourself, or choose from four standard themes designed by the Mattermost team.
+
+In Enterprise Edition, if you belong to multiple teams, you can choose to apply any changes across teams by selecting "Apply New Theme to All Teams" before saving. If this is not selected, any changes are only applied to the current team.
```
jasonblais: I'm not sure if we should change this where you have the option to apply the theme on a per-team basis.

Because in TE, `you can choose to apply any changes across teams by selecting` actually works, which is the default behaviour.

source/help/settings/theme-colors.rst
```diff
@@ -0,0 +1,138 @@
...
+Go to **Account Settings > Display > Custom Theme > Copy and paste to share theme colors** and copy the theme code to export it.
+
+Importing a Custom Theme
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can import a theme by pasting a "Theme Code" into the Custom Theme menu.
+
+Go to **Account Settings > Display > Custom Theme > Copy and paste to share theme colors** and paste a theme code to import it, then click **Save**.
+
+You can find sample themes below. You can also find `theme codes shared by the community from the Mattermost community forum. <https://forum.mattermost.org/t/share-your-favorite-mattermost-theme-colors/1330>`_
```
jasonblais: These won't work if they're copied over though, I think, since they use the old theme vector format

source/help/settings/theme-colors.rst
JeffSchering: Ok. So is it that by default in EE it's applied to all teams and you can choose to apply to the current team only, or is it that by default it's applied to the current team only and you can choose to apply to all teams?

source/help/settings/theme-colors.rst
JeffSchering: Yes. Can the forum post be updated?

source/help/settings/theme-colors.rst
jasonblais: It's the first `by default it's applied to all teams and you can choose to apply to the current team only`, both in TE and EE.

Then in EE you have the additional option of applying a theme on for the current team only.

source/help/settings/theme-colors.rst
jasonblais: Their from individual users, so not sure.

JeffSchering: The original announcement can be probably be edited though. We could add a note about the new format, and possibly convert the submitted themes and add those to the announcement. At the very least, we could add a reply with the info. I'll create an Issue for it and add it to my to-do list.

PR #912, opened 2017-02-22
Docs for telemetry

jasonblais:
Reviewed in detail by Ian already, so doing one review on proofread and styling should be sufficient.

EDIT:

The goal for this document is to be clear and upfront on what telemetry is collected. At the same time, we wanted to include the details compactly so that it's easy to scan through the categories (e.g. Deployment configuration information, server configuration information, etc.). 

Finally, there are two separate features: "security fix alert" and "error and diagnostics reporting". The telemetry data for these features may overlap.


source/administration/telemetry.rst
```diff
@@ -0,0 +1,80 @@
...
+
```
JeffSchering: Is this a typo?
> bool RestrictDirectMessag;

jasonblais:
This looks good to me, but I would confirm with Ian before merging

PR #934, opened 2017-02-26
Update sending-messages.rst

lindy65:
Fixing some typos/grammar

### DIFF
--- a/source/help/messaging/sending-messages.rst
+++ b/source/help/messaging/sending-messages.rst
@@ -8,7 +8,7 @@ After a message has been sent, you or others can reply to it and link to it. Dep
 Message Types
 -------------
 
-Mattermost has two types of messages: posts, and replies.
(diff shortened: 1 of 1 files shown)


source/help/messaging/sending-messages.rst
```diff
@@ -8,7 +8,7 @@ After a message has been sent, you or others can reply to it and link to it. Dep
 Message Types
 -------------
 
-Mattermost has two types of messages: posts, and replies.
+Mattermost has two types of messages: posts and replies.
```
JeffSchering: I'd argue to retain the comma here, using the serial comma rule. If the sentence was like this for example:

Mattermost has three types of messages: posts, reposts, and replies.

Then we'd definitely need commas. My opinion is that even though the list is only two items long, a comma is still required. But I don't think it's important enough to delay a merge.

PR #942, opened 2017-03-01
Deleted note about attachments not supported for slash commands

JeffSchering:
I also made a few updates for Style Guide.


lfbrock:
+1

Issue #882, opened 2017-02-16
Document how to invoke mattermost CLI tool in GitLab omnibus


jasonblais:
It doesn't look like there's any documentation about how to run the CLI commands in GitLab omnibus.

For instance, running the [`platform team add {team} {users}`](https://docs.mattermost.com/administration/command-line-tools.html#platform-team-add) command in GitLab omnibus is done with the following command.

```
sudo -u mattermost /opt/gitlab/embedded/bin/mattermost --config=/var/opt/gitlab/mattermost/config.json team add ourteam thatuser
```

We'll want to decide whether to add this to [GitLab Mattermost docs](https://docs.gitlab.com/omnibus/gitlab-mattermost/), [Mattermost CLI docs](https://docs.mattermost.com/administration/command-line-tools.html#platform-team-add) or both.

jasonblais:
Resolved by #949

PR #949, opened 2017-03-03
Add legacy CLI commands back to the docs

jasonblais:
Admins are trying to use the new CLI commands on older Mattermost deployments. This is causing a lot of unnecessary pain for them.

Propose we add the legacy CLI commands back until majority of deployments are running Mattermost 3.6 or later

### DIFF
--- a/source/administration/command-line-tools.rst
+++ b/source/administration/command-line-tools.rst
@@ -34,6 +34,11 @@ From the directory where the Mattermost platform is installed, a
 .. contents::
     :backlinks: top
 
+Mattermost 3.6 and later
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+The new CLI tool is supported in Mattermost 3.6 and later. To see available commands in the old CLI tool, see `Mattermost 3.5 and earlier`_.
+
 Typing ``platform help`` and ``platform help {command}`` returns help documentation for the CLI tool or any CLI command in particular.
 
 Notes:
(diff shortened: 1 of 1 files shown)


source/administration/command-line-tools.rst
```diff
@@ -655,3 +660,181 @@ platform version
...
+Mattermost 3.5 and earlier
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Typing ``platform -help`` brings up documentation for the CLI tool.
+
+Notes:
+
+- Parameters in CLI commands are order-specific.
+- If special characters (``!``, ``|``, ``(``, ``)``, ``\``, `````, and ``"``) are used, the entire argument needs to be surrounded by single quotes (e.g. ``-password 'mypassword!'``, or the individual characters need to be escaped out (e.g. ``-password mypassword/!``).
+- Team name and channel name refer to the handles, not the display names. So in the url ``https://pre-release.mattermost.com/core/channels/town-square`` team name would be ``core`` and channel name would be ``town-square``
```
JeffSchering: Is that true that the escape character is a forward slash? Normally it's a backslash. For example: ``-password mypassword\!``

source/administration/command-line-tools.rst
```diff
@@ -34,6 +34,11 @@ From the directory where the Mattermost platform is installed, a
 .. contents::
     :backlinks: top
```
JeffSchering: If you add :local: to this then the page title won't be included in the contents. For example:

```
.. contents::
  :backlinks: top
  :local:
```

source/administration/command-line-tools.rst
```diff
@@ -655,3 +660,181 @@ platform version
     .. code-block:: none
 
       platform version
+
+Mattermost 3.5 and earlier
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Typing ``platform -help`` brings up documentation for the CLI tool.
+
```
JeffSchering: This doesn't work if you simply cd into the bin directory and type ``platform -help``.  You need to use ``./platform -help``. Also, you need to be a user that has permissions to do this, so the command should be ``sudo ./platform -help``. Similarly for all commands, both for 3.6+ and for 3.5 and earlier.

Experienced Linux admins will know this already. But experienced Windows admins don't know the ins and outs of the Linux command line, and when you tell them to do ``platform -help`` they'll wonder for hours why they get the "platform command can't be found error". And when they do finally figure it out, they'll wonder why the thing appears to crash when you invoke any command other than -help without sudo.

jasonblais:
Thanks Jeff for the feedback - I think @esethna prepared the new docs for 3.6, maybe he'd know and be able to answer your questions? 

I'm not really a subject expert on this one..

source/administration/command-line-tools.rst
esethna: Looks like an error in the command as the list of special characters shows a backslash.

source/administration/command-line-tools.rst
esethna: Thanks Jeff. Are you proposing every command leads with `sudo ./platform`? Or could we just add something to the doc intro section to explain this?

JeffSchering: I think adding it only to the examples would be good, plus something to the intro section.

jasonblais:
Updated based on feedback.

Also addressed https://github.com/mattermost/docs/issues/882 which has been brought up a few times by the community

PR #952, opened 2017-03-04
Update channel-settings.rst

lindy65:
Fixed a typo and added a note on how to add markdown links to channel headers

### DIFF
--- a/source/help/settings/channel-settings.rst
+++ b/source/help/settings/channel-settings.rst
@@ -41,7 +41,7 @@ has `restricted the permissions <https://docs.mattermost.com/administration/conf
 Adding links to the Channel Header
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Frequently used document, website or Google Hangouts links can be added to the channel header using markdown.
+Frequently used links to documents, video calls, or other sites can be added to the channel header using markdown.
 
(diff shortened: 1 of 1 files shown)


source/help/settings/channel-settings.rst
```diff
@@ -38,6 +38,14 @@ the channel topic or provide links to frequently accessed documents. Any
 channel member can edit this setting, unless the System Administrator
 has `restricted the permissions <https://docs.mattermost.com/administration/config-settings.html#enable-public-channel-renaming-for>`__.
 
+Adding links to the Channel Header
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Add links to the channel header using markdown.
+
```
JeffSchering: I think it's a stronger sentence if you write:

*Use Markdown to add links to the channel header.*

or even

*You can use Markdown to add links to the channel header.*

But then that raises the question: Is all Markdown formatting available in the header, or just linking?

lindy65: Did a quick test with bold, italics, emoji and code block and it seems all markdown is available in a channel header.

lfbrock:
@lindy65 - can we add an extra sentence explaining why people would want to add links? 

Just something simple, maybe mentioning that frequently used links can be added to the channel header for easy access. Also maybe use something practical for the example, like adding a Google Hangouts link? (With a fake URL)

lfbrock:
+1

PR #955, opened 2017-03-06
Added docs for the bulk loading feature

JeffSchering:
Added the docs, and added them to the ToC in the Deployment section of the Administrator's Guide. 

Note that the doc is modular. That is, there are four files for one doc. To see how the doc looks in it's entirety, see https://jeffschering.github.io/mmdocs/bulk-loading/deployment/bulk-loading.html

### DIFF
--- a/source/deployment/bulk-loading-about.rst
+++ b/source/deployment/bulk-loading-about.rst
@@ -1,0 +1,19 @@
+.. _about-bulk-loading-command:
+
+About the bulk loading command
+==============================
+
+The bulk loading command is interruptible and idempotent
+  If the import is interrupted for any reason, it continues from where it left off the next time you run it. You can run the command repeatedly with the same data file, and the data is imported only once.
+
+You can run the bulk loading command on a live system
+  Although you don't need to shut down Mattermost to run the command, changes made by users of the system between runs can be overwritten if the corresponding fields exist in the data file.
+
+Some data fields are optional
+  Not all fields are mandatory. If an optional field is missing from the object that is being imported, the field's current value in the database is not changed.
+
+The bulk loading command is not a synchronization tool
+  You cannot use the bulk loading command to remove any objects or their fields from the Mattermost database. The command only creates or overwrites fields.
+
+.. important::
+  The bulk loading command runs in the CLI and operates in the security context of the CLI. This means it has full permissions to access and alter everything in the Mattermost database.
--- a/source/deployment/bulk-loading-data.rst
+++ b/source/deployment/bulk-loading-data.rst
@@ -1,0 +1,25 @@
+.. _bulk-loading-data:
+
+Running the bulk loading command
+================================
+
+Before running the bulk loading command, you must first create a JSON file that contains the data that you want to import. After you create the file, run the bulk load command in validation mode. In this mode, the data is checked for correctness, but is not written to the database. After validating, run the command in apply mode, which saves the data to the database.
+
+**To bulk load data**:
+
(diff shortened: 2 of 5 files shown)


source/deployment/bulk-loading.rst
```diff
@@ -0,0 +1,27 @@
+.. _bulk-loading:
+
+=================
+Bulk Loading Data
+=================
+
+Large quantities of data can be imported from a JSON file into Mattermost at the command line using the bulk loading feature. This feature is most suitable for migrating data from an existing messaging system, or for pre-populating a new installation with data.
```
lindy65: "...at the command line using the bulk loading feature." > "using the bulk loading feature from the command line."

source/deployment/bulk-loading.rst
```diff
@@ -0,0 +1,27 @@
...
+You can import the following data types:
+
+- Teams
+- Channels (Public & Private)
+- Users
+- Users' Team memberships
+- Users' Channel memberships
+- Posts (regular, non-reply posts)
+
+Importing Direct Message channels and importing additional types of posts are not yet supported.
```
lindy65: "Importing Direct Message channels and importing additional types of posts are not yet supported." > "Neither importing direct message channels nor importing additional types of posts is yet supported."

source/deployment/bulk-loading-data.rst
```diff
@@ -0,0 +1,25 @@
+.. _bulk-loading-data:
+
+Running the bulk loading command
+================================
+
+Before running the bulk loading command, you must first create a JSON file that contains the data that you want to import. After you create the file, run the bulk load command in validation mode. In this mode, the data is checked for correctness, but is not written to the database. After validating, run the command in apply mode, which saves the data to the database.
```
lindy65: "...the data that you want to import." > "... the data you want to import."

source/deployment/bulk-loading-about.rst
```diff
@@ -0,0 +1,19 @@
+.. _about-bulk-loading-command:
+
+About the bulk loading command
+==============================
+
+The bulk loading command is interruptible and idempotent
```
lindy65: "...idempotent" 

I'm not sure who's going to be reading this particular doc but, as a non-developer, I did not know what this word meant and had to look it up...

Meaning: "denoting an element of a set which is unchanged in value when multiplied or otherwise operated on by itself"

Not sure whether we need to expand on the meaning of the word?

source/deployment/bulk-loading.rst
JeffSchering: I think it's better as-is. When you say "using the bulk loading feature from the command line" you're implying that you can use it from somewhere else; a GUI for example. In this case, the command line is the only way to use the bulk loading feature.

source/deployment/bulk-loading.rst
JeffSchering: I think it's better as-is.

source/deployment/bulk-loading-data.rst
JeffSchering: My understanding is that for those who do not have a firm grasp of English, having "that" makes the sentence easier to parse and understand. It's not grammatically incorrect to have "that", so I think we should keep it here.

source/deployment/bulk-loading-about.rst
JeffSchering: I think idempotent as used here is understood by the target audience.

JeffSchering:
@grundleborg Thanks! I made the updates. Let me know how they look.

PR #960, opened 2017-03-07
Update design process doc

lfbrock:
I don't think the link is working properly here in the training doc, maybe someone can take a look

screenshot 2017-03-07 14 23 59


jasonblais:
@JeffSchering can you also confirm the relative link works correctly? @lfbrock mentioned it might not be working properly

JeffSchering:
@jasonblais @lfbrock  It's not working because the syntax is slightly off. It's either got a missing . or has an extra /

These work:
```
:doc:`./design-process`
:doc:`design-process`
```

PR #964, opened 2017-03-08
Updated for changes to webhook username override behaviour

JeffSchering:
Ref:  [PLT-5279 Integrations: When no override username is set, show the user who set up the integration's username](https://mattermost.atlassian.net/browse/PLT-5279)

For ``webhooks-incoming.rst`` and ``webhooks-outgoing.rst``: I converted them from .md to .rst, so the diffs show that everything as changed, but this is the part we're interested in:

```
1. Login to your Mattermost team account that has the system administrator role.
2. Enable incoming webhooks from **System Console > Integrations > Custom Integrations**.
3. (Optional) Configure the **Enable Overriding of Usernames from Webhooks** option to 
    allow external applications to post messages under any name. If not enabled, 
    the username is set to "webhook".
4. (Optional) Configure the **Enable Overriding of Icon from Webhooks** option to allow 
    external applications to change the icon of the account posting messages. If not 
    enabled, the icon of the creator of the webhook URL is used to post messages.
```

For ``slash-commands.rst``, I made no changes... it's just here because it got caught up in the web of conversions from .md to .rst.


source/administration/config-settings.rst
```diff
@@ -1601,9 +1601,9 @@ Note: OAuth 2.0 applications can be authorized by all users if they have the **C
 
 Enable integrations to override usernames
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-**True**: Webhooks, slash commands and other integrations, such as `Zapier <https://docs.mattermost.com/integrations/zapier.html>`_, will be allowed to change the username they are posting as.
+**True**: Webhooks, slash commands, and other integrations, such as `Zapier <https://docs.mattermost.com/integrations/zapier.html>`_, will be allowed to change the username they are posting as. If no username is present, the username for the post is the same as it would be for a setting of **False**.
 
-**False**: Webhooks, slash commands and OAuth 2.0 apps can only post as the username of the account they were set up with. See http://mattermost.org/webhooks for more details.
+**False**: Custom slash commands and OAuth 2.0 apps can only post as the username of the user who used the slash command or OAuth 2.0 app. For incoming webhooks and outgoing webhooks, the username is "webhook". See http://mattermost.org/webhooks for more details.
```
lfbrock: For OAuth 2 app, is it the username of the user who set up the app? Or is it something similar to slash commands where you can trigger the oauth 2 app and then it uses the username of the person who triggered it? 

Maybe get Linda to help test what the actual behaviour is if it's not clear

JeffSchering: @grundleborg Can you help with this question? Do we need to call out "OAuth 2.0 apps" in the description for **False**, or are they part of "other integrations" as in the description for **True**? We should be consistent... either "other integrations" or "OAuth 2.0 apps" in both cases. Or are they different things?

grundleborg: I think any change to the wording of "False" would be wrong. The reason "true" is worded the way it is appears to be an attempt to namedrop Zapier integrations in there, probably to draw attention to the fact we support them. So, if you wanted, you could explicitly list OAuth2 apps in the description of "true".

grundleborg: Also, I'm not sure of the answer to Lindsay's question, so if you haven't got an answer to that yet, you would need to ask someone to actually set up some of the integrations and see what happens.

lindalumitchell: @JeffSchering I did some testing, and posts made by a Zapier OAuth2 integration post as the user who set up the integration. 

Also confirmed that custom slash commands post with the username of the user who posted the slash command in the channel (not necessarily the user who created the slash command). Confirmed that incoming webhook posts as the user who set up the webhook.

Do you want me to test further? Or is that enough info to verify?

JeffSchering: Thanks @lindalumitchell !

> I did some testing, and posts made by a Zapier OAuth2 integration post as the user who set up the integration.

Ok, so it looks like the doc might be wrong, but to confirm, are these statements correct for a setting of **False**?

1. Custom slash commands can only post as the username of the user who used the slash command
1. OAuth 2.0 apps can only post as the username of the user who set up the integration

lindalumitchell: Those statements are correct according to my testing, yes (though I only tested the one OAuth integration). Let me know if you want me to confirm on another one or two OAuth examples; I should have time later tonight if need be. But yes, from the testing I did, those are correct statements.

lfbrock:
+1

PR #969, opened 2017-03-10
Note that Segment developer key was removed

jasonblais:


jasonblais:
@esethna please help re-review?

PR #975, opened 2017-03-11
EE install for Ubuntu 16.04

it33:


it33:
Merging to answer customer question

PR #997, opened 2017-03-16
Added LimitNOFILE=49152 to unit file.

JeffSchering:
I copied the value 49152 from the RHEL 7.1 docs.

Also, deleted trailing whitespace.

### DIFF
--- a/source/install/install-ubuntu-1604-mattermost.rst
+++ b/source/install/install-ubuntu-1604-mattermost.rst
@@ -41,11 +41,11 @@ Assume that the IP address of this server is 10.10.10.2.
 
 6. Set up the database driver in the file ``/opt/mattermost/config/config.json``. Open the file in a text editor and make the following changes:
 
-  -  If you are using PostgreSQL:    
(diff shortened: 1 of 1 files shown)


source/install/install-ubuntu-1604-mattermost.rst
```diff
@@ -81,13 +81,14 @@ Assume that the IP address of this server is 10.10.10.2.
     WorkingDirectory=/opt/mattermost
     User=mattermost
     Group=mattermost
+    LimitNOFILE=49152
```
enahum: Why this number? Can you confirm with CH?

JeffSchering: I copied the value from the RHEL 7.1 install docs. @coreyhulen Can you confirm that this is ok, or should I change it?

coreyhulen: I think it's fine.  I think in other places we use 50000, but 49152 is fine.

PR #1000, opened 2017-03-17
Note how to find troubleshooting help and editing

it33:


source/install/install-ubuntu-1404.rst
```diff
@@ -12,7 +12,7 @@ Install a production-ready Mattermost system on 1 to 3 machines.
   :backlinks: top
   :local:
   
-.. note:: If you have any issues with these installation instructions, `please see consult our troubleshooting forum. <https://www.mattermost.org/troubleshoot/>`_ 
+.. attention:: If you have any issues with these installation instructions, `please see consult our troubleshooting forum. <https://www.mattermost.org/troubleshoot/>`_ To submit and improvement or correction, use the "Edit" button at the top of this page.
```
it33: Okay on note instead of attention and moving the note

I think we should keep the call out to "Edit" since we want to draw attention to the fact that we welcome the community to improve the docs.

PR #1005, opened 2017-03-19
WIP: Adding quote on commuting and bandwidth

it33:


amyblais:
Adding key changes from this PR to https://github.com/mattermost/docs/pull/1112, so they can be reviewed at the same time.

PR #1012, opened 2017-03-24
Added instructions to limit access to port 8065

JeffSchering:
After the NGINX proxy is set up and serving Mattermost on port 80, you should restrict access to port 8065 to the machine that hosts NGINX, otherwise Mattermost is available on port 8065 to every machine on the network.

This is done by modifying "ListenAddress" in config.json.

See http://forum.mattermost.org/t/after-nginx-install-still-port-8065-open/3078 for discussion.

### DIFF
--- a/source/install/config-proxy-nginx.rst
+++ b/source/install/config-proxy-nginx.rst
@@ -71,13 +71,9 @@ NGINX is configured using a file in the ``/etc/nginx/sites-available`` directory
 
 6. Restart NGINX.
 
-  On Ubuntu 14.04 and RHEL 6.6:
+  On Ubuntu 14.04 and RHEL 6.6: ``sudo service nginx restart``
 
-  ``sudo service nginx restart``
-
-  On Ubuntu 16.04 and RHEL 7.1:
-
-  ``sudo systemctl restart nginx``
+  On Ubuntu 16.04 and RHEL 7.1: ``sudo systemctl restart nginx``
 
 7. Verify that you can see Mattermost through the proxy.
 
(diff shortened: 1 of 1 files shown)


source/install/config-proxy-nginx.rst
```diff
@@ -71,20 +71,28 @@ NGINX is configured using a file in the ``/etc/nginx/sites-available`` directory
...
 
 7. Verify that you can see Mattermost through the proxy.
 
   ``curl http://localhost``
-  
+
   If everything is working, you will see the HTML for the Mattermost signup page.
 
-**What to do next**
+8. Make sure that Mattermost accepts connections only from the proxy.
```
coreyhulen: Can we make this step optional.  It is more secure like this, but there are times when you may want to connect to the app server directly to administer the system console of each server when in HA mode for example.

coreyhulen: Most would usually do this via iptables or an AWS security group vs modifying the MM port.  For example the IP port may change.

JeffSchering: My understanding is that in HA, you can't modify the settings in the System Console because config.json is identical on each server and the System Console is read-only. The only way to update the config is to SSH into the machine and modify config.json directly.

The only time you'd need the System Console is when updating config changes without any downtime, in which case you have to use the System Console at each machine to reload the config. I added a paragraph that should cover this.

I changed {IP-address}  to {IP-or-domain} to make it more clear that ListenAddress can take a domain as well as an address.

coreyhulen: You'll still want to see the individual HA consoles to see if any are having trouble talking even if you cannot set props in the console.  You'll also need individual access to each console to call re-load config from disk.

Changing the bind address is a distant second to change an iptable or security group.  We should at the very least make a recommendation that it would be better to change this using iptables or security groups.  The bind feature really exists when you have multiple nic cards and want Mattermost to bind to only 1.

coreyhulen: It's not the recommended way to secure a port.

JeffSchering: Is this ok...

>8. Restrict access to port 8065.
    By default, the Mattermost server accepts connections on port 8065 from every machine on the network. Use your firewall to deny connections on port 8065 to all machines except the machine that hosts NGINX and the machine that you use to administer Mattermost server.

Admins who don't know how to use a firewall on Linux will need to learn how, those who do know about firewalls will know what to do, and those using AWS should be able to figure it out, but maybe for them we can add something like "If you're installing on Amazon Web Services, you can use security groups to block access."

coreyhulen:
Sounds good

PR #1034, opened 2017-03-30
Fixed broken links and turned off index page generation

JeffSchering:
In addition to fixing broken links, modified config.py so that the auto-generated index page, which was empty anyway, is not generated. The generated page had an "Edit on Github" link but since it's a generated page, there's nothing on Gitub to edit resulting in chronic 404ism.

And my text editor removed superfluous whitespace.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -128,11 +128,11 @@ todo_include_todos = False
 html_context = {
     # Enable the "Edit in GitHub link within the header of each page.
     'display_github': True,
-    # Set the following variables to generate the resulting github URL for each page. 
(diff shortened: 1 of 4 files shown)


source/conf.py
```diff
@@ -195,7 +195,7 @@ def setup(app):
 #html_domain_indices = True
 
 # If false, no index is generated.
-#html_use_index = True
+html_use_index = False
```
esethna: Does this change need to be tested locally? If you haven't done so already? I'm not clear what this is doing.

PR #1040, opened 2017-03-31
Add to desktop app troubleshooting guide

jasonblais:


source/install/desktop.rst
```diff
@@ -85,3 +85,6 @@ Desktop App window is black and doesn't load the page
     - Finally, try disabling GPU hardware acceleration by using the `--disable-gpu <http://peter.sh/experiments/chromium-command-line-switches/#disable-gpu>`_ Chromium command line switch.
 
     If none of the above steps resolve the issue, please open a new ticket in the `Mattermost Troubleshooting Forum <https://forum.mattermost.org/t/how-to-use-the-troubleshooting-forum/150>`_. 
+
+Desktop app appears off-screen after disconnecting from a secondary screen
+    Quit the app and delete the following file: ``USERPROFILE%\AppData\Roaming\Mattermost\bounds-info.json``
```
JeffSchering: Is this intended for end users? If so, then they need more explicit instructions on how to delete the file.

How does the end user distinguish between the app being off-screen and having crashed?

What do you mean by off-screen? Is it a multiple monitor setup? Or like Linux where you can have multiple workspaces, each with its own set of open windows?

jasonblais: Repro steps:

1-Open mattermost.
2-Move the window to a secondary screen.
3-Access the computer using the remote desktop feature as the same user.
4-Disconnect from the remote desktop session.
5-After logging in the computer directly, the mattermost window is still running but it spawns to an off-screen location on the secondary screen.

1-Open mattermost.
2-Move the window to a secondary monitor.
3-Move Mattermost to the secondary monitor.
4-Disconnect from the second monitor.
5-After logging in the computer directly, the mattermost window is still running but it spawns to an off-screen location on the secondary monitor.

Would you have a suggestion on how to better describe the issue? After re-reading I agree the `off-screen` isn't clear.

And yes, this is for the end user (since each user installs the app).

JeffSchering: Maybe something like this:

**Desktop App is not visible, but you can see the Mattermost icon in the Task Bar**
  This issue can occur on Windows in a multiple-monitor setup. When you disconnect the monitor that Mattermost is displayed on, Mattermost continues to display at screen coordinates that no longer exist.

  To resolve this issue, you can reset the Desktop App screen location by deleting the screen location file. When the file is not present, the Desktop App displays on the primary monitor by default. 
  
  **To reset the Desktop App screen location**:
    1. If the Desktop App is running, right-click the Mattermost icon in the task bar and click **Close Window**.
    2. Open Windows File Explorer, and navigate to the ``%USERPROFILE%\AppData\Roaming\Mattermost`` folder.
    3. Delete the file ``bounds-info.json``

jasonblais: nice 👍

PR #1045, opened 2017-04-02
Add start of core developer handbook

jwilander:
The purpose of this is to move information relevant to core developers from any old private wikis we have into the docs repo and to allow the development team to track decisions made surrounding the technical vision of Mattermost.

This should also be a useful tool for on-boarding new core developers.

### DIFF
--- a/source/developer/core-developer-handbook.md
+++ b/source/developer/core-developer-handbook.md
@@ -1,0 +1,36 @@
+Core Developer Handbook
+-----------------------------
+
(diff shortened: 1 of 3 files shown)


jwilander:
Good questions @it33

1. By Mattermost project I mean everything under the mattermost org on GitHub
2. If they have merge privileges on some repositories and responsible for merging/reviewing PRs then yes I think we should consider them core and I can add them to the list mentioning the repos they are maintainers of

PR #1051, opened 2017-04-04
Added a Mattermost Rollout Template

JeffSchering:
This template provides sample text and links for internal distribution. It can be used by organizations to help their users get started with Mattermost, and assumes that the users are not familiar with other messaging systems.


lindy65:
Thanks Jeff, editor review done. +1

PR #1052, opened 2017-04-05
Update contribution-guide.md

lindy65:
Added a point for Asaad to review and approve any PR with CSS changes.

### DIFF
--- a/source/developer/contribution-guide.md
+++ b/source/developer/contribution-guide.md
@@ -63,11 +63,11 @@ After submitting a PR, before it is merged:
         - Works as expected
         - Meets [UX guidelines](https://docs.mattermost.com/developer/fx-guidelines.html)
     - The Product Manager may come back with some bugs or UI improvements to fix before the pull request moves on to dev review.
-3. Dev review must pass
+3. CSS review must pass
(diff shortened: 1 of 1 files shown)


source/developer/contribution-guide.md
```diff
@@ -66,6 +66,8 @@ After submitting a PR, before it is merged:
 3. Dev review must pass
     - Two core committers will review the pull request and either give feedback or approve the PR.
     - Any comments will need to be addressed before the pull request is ready to merge.
+4. CSS review must pass
```
jasonblais: propose we move it before dev review

also, would it be in PM or Dev review stage following the contributing guidelines? https://github.com/mattermost/platform/blob/master/CONTRIBUTING.md

lindy65: Thanks Jason, I've moved the CSS review to before the Dev review.

Re the contributing guidelines, I would say CSS review should happen after PM review and before Dev review.  Would we need a label in GitHub for that?

jasonblais: Hmm good question.

I think that can happen during the "PM review" stage, so it can be added in this section: https://github.com/mattermost/platform/blob/master/CONTRIBUTING.md#stage-1-pm-review

PR #1058, opened 2017-04-06
Added note about attachments in search results

JeffSchering:
File attachments and their filenames are not searched. If file attachments show up in the search results, it's because they are attached to messages that match the search query.

### DIFF
--- a/source/help/getting-started/searching.rst
+++ b/source/help/getting-started/searching.rst
@@ -1,22 +1,15 @@
 Searching
 =========
 
---------------
+Use the search box to find messages and replies that match the search terms that you enter. All channels that you are a member of are searched.
 
-The search box in Mattermost returns search results from any channel of
-which you’re a member.
-
--  Searching multiple words returns results containing both of the words
-   listed.
+-  Multiple-word search terms return results that contain all of the terms.
 -  When results appear, click the **Jump** link to view that post in the
    channel archive.
--  You can expand the search results window by clicking the “Expand”
-   button in the top right corner.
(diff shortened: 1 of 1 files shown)


source/help/getting-started/searching.rst
```diff
@@ -1,22 +1,15 @@
...
--  Use search modifiers, like ``from:dave``, to return results only from
-   certain people or in certain channels (see below).
+- File attachments and their filenames are not searched. If file attachments show up in the search results, it's because they are attached to messages that match the search query.
+-  You can use search modifiers such as ``from:dave`` to return results only from certain people or in certain channels. For more information about this, see the `Search Modifiers`_ section.
 
-Like many search engines, highly common words like ``the``, ``which``,
-``are`` (known as “stop words”), as well as two-letter and one-letter
+Like many search engines, common words such as ``the``, ``which``, and ``are`` (known as "stop words"), as well as two-letter and one-letter
 search terms, are not shown in search because they typically return too
 may results. See database documentation on `MySQL`_ and
```
lindy65: "may" > "many"

source/help/getting-started/searching.rst
```diff
@@ -1,24 +1,17 @@
 Searching
 =========
 
---------------
+Use the search box to find messages and replies that match the search terms that you enter. All channels that you are a member of are searched.
```
jasonblais: We might want to include a note that only the team that you are currently viewing is  searched

e.g. if you're viewing team A, only channels (that you're a member of) are searched within team A, but not channels in other teams.

PR #1041, opened 2017-03-31
Update release-process.md

esethna:
Opening a WIP PR for release process updates

PR #1080, opened 2017-04-14
Create deprecated-features.rst

jasonblais:
FYI - https://github.com/mattermost/docs/pull/1041 contains relevant changes to the release process

### DIFF
--- a/source/process/deprecated-features.rst
+++ b/source/process/deprecated-features.rst
@@ -1,0 +1,14 @@
+Announcing Deprecated Features
+===============================
+
+This document outlines the process for announcing deprecated features to the community. The guiding principle is `no surprises <https://docs.mattermost.com/developer/manifesto.html#no-surprises>`_ where users should never run into anything unexpected with Mattermost.
+
+Once the decision to deprecate a feature or functionality is made, the acting release manager carries out the following actions:
+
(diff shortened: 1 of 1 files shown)


jasonblais:
Thanks!

> The deprecated features page, https://about.mattermost.com/deprecated-features/, needs updating to move the scheduled April content into the removed features section.

That's correct, was going to do it today, should be updated now

> In Step 1, is linking to the changelog and the upgrade guide a one-time process? Or do you want those docs updated with links to specific items in the deprecated features page? If the latter, then you need to revise that page to include html targets for linking to.

What I meant is that this link is also included in the changelog (see [this for an example](https://docs.mattermost.com/administration/changelog.html#removed-and-deprecated-features)) and the upgrade guide (see [step 5 of the upgrade guide](https://docs.mattermost.com/administration/upgrade.html#upgrade-team-edition-to-3-1-x-and-later)), so no further updates are needed from the PM.

> Are items 2, 3, and 4 going to be added to the Release Process page?

Yes - #1041 contains relevant changes to the release process

> Is item 4 redundant? It's the same as "When the decision to deprecate a feature or function ..." and is essentially Step 1.

It is, but it's a check for the PM to ensure the page is up-to-date before the release and updated with anything that might have been missed before

JeffSchering:
I updated PR https://github.com/mattermost/docs/pull/1093

Issue #805, opened 2017-01-20
UX guidelines for on-screen text


jasonblais:
Since there's a PR to update our UX guidelines (#759), I'll create a new issue to track new guidelines for on-screen text, rather than creating a WIP PR for it which would result in merge conflicts:

1 - The title of a setting should explicitly describe what a setting affects. 
Example: A setting to default the expanded or collapsed state of link previews
       - incorrect: "Link previews"
       - correct: "Automatically expand link previews"

2 - Use the active voice
     - incorrect: "This action will delete..."
     - correct: "This action deletes..."

3 - Be concise
     - incorrect: "Are you sure that you want..."
     - correct: "Are you sure you want..."

4 - Ensure plural interpretations work when applicable
     - incorrect: "...breaks the integration using it"
     - correct: "...breaks any integrations using it"

5 - Test text with machine translation to reduce non-obvious, English-specific jardon: 
     - incorrect: "There are a few configuration settings you might want to fiddle with when setting up your instance of Mattermost"
     - correct: “There are a few configuration settings you can adjust when setting up your instance of Mattermost.”

Note: The incorrect string above machine-translated into German and then back into English looks like this: "There are some configuration settings you could know if your instance Matter Most violin".

See https://pre-release.mattermost.com/core/pl/nf9agdb48ife9yrhzw7z4cj6do for more details

it33:
Great start!

Propose we include the "Thor test", or refer to it as the embodiment of the "Mattermost voice for documentation and help text." The Thor test asks "How would Agent Coulson explain this to Thor?" which means explaining something to an alien being in simple terms without jargon or complexity. 

This is probably the "north star" to think about (include being concise, easy to translate) and the other guidelines seem like finer details/safety net checks.

jasonblais:
1. consideration of surrounding UI, including titles above the input field (e.g. "Notification Preferences for [CHANNEL]) and
2. consistent with naming of input fields with similar actions, e.g. "Send email notifications", "Send desktop notifications".

jasonblais:
See #1084 for proposed changes

PR #1084, opened 2017-04-15
Update on-screen UX guidelines

jasonblais:
Ticket: #805

### DIFF
--- a/source/developer/fx-guidelines.rst
+++ b/source/developer/fx-guidelines.rst
@@ -362,7 +362,7 @@ input field should trigger default dialog button.
 Input: Hints
 ------------
 
-Mattermost voice for documentation and help text follows this simple guideline: explain something to an alien in simple terms without jargon or complexity.
+Mattermost voice for documentation and help text follows this simple guideline: explain something to a user with basic computer skills in simple terms without jargon or complexity.
 
 **Fieldname:**
 
@@ -384,7 +384,7 @@ Mattermost voice for documentation and help text follows this simple guideline: 
 
 **Help text:**
 
-    Text below an input field should clearly and concisely describe the purpose of the input in active tense.
+    Text below an input field should clearly and concisely describe the purpose of the input in active voice.
 
     In general, avoid describing the technical requirements of the field in Help Text, and use Placeholder input text, field validation and error messages to convey requirements. The exception to this guideline is if requirements are non-obvious, such as passwords needing different numbers of characters, symbols, etc.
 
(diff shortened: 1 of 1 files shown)


source/developer/fx-guidelines.rst
```diff
@@ -391,18 +396,25 @@ Input: Hints
     Incorrect:
         When a new message is received, flash the taskbar icon.
 
+    Ensure plural interpretations work when applicable and test help text with machine translation to reduce non-obvious, English-specific jardon:
     
+    Correct:
+        There are a few configuration settings you can adjust when setting up your instance of Mattermost.
```
JeffSchering: As a general rule, you should avoid starting a sentence with "There are..." because it moves the subject of the sentence way out to the middle or the end and sometimes makes it non-obvious what the subject of the sentence actually is. In this case, the subject is either "you" or "your instance of Mattermost". You can rewrite it in two ways:

1. You can adjust a few configuration settings when setting up your instance of Mattermost.
1. Your Mattermost instance has a few configuration settings that can be adjusted when setting it up.

Both put the subject at the beginning where it's easier to understand. The first example is better because it uses active voice and is in the imperative mood (present tense, second person)

source/developer/fx-guidelines.rst
```diff
@@ -362,24 +362,29 @@ input field should trigger default dialog button.
...
+    Correct:
+        Automatically expand link previews
+        
+    Incorrect:
+        Link previews
 
 **Help text:**
 
-    Text below an input field should clearly and concisely describe the purpose of the input.
+    Text below an input field should clearly and concisely describe the purpose of the input in active tense.
```
jasonblais: sorry yes, "active voice"

source/developer/fx-guidelines.rst
```diff
@@ -362,24 +362,29 @@ input field should trigger default dialog button.
 Input: Hints
 ------------
 
+Mattermost voice for documentation and help text follows this simple guideline: explain something to an alien in simple terms without jargon or complexity.
```
jasonblais: maybe replace "alien" with "user with basic computer skills"?

JeffSchering: Ok, that works.

JeffSchering:
@esethna Can you revert this change? The phrase "active voice" can have either no article or it can have a direct article. It cannot have an indirect article because it changes the meaning of the sentence.

Correct:
  "... of the input in active voice."
  "... of the input in the active voice."

Not Correct:
  "... of the input in an active voice."

See http://writing.wisc.edu/Handbook/CCS_activevoice.html and http://www.quickanddirtytips.com/education/grammar/active-voice-versus-passive-voice for examples.

Issue #1018, opened 2017-03-26
Make configuration settings more discoverable


it33:
Most of what people want is in Configuration Settings and they're not show at TOC level and they should be: https://docs.mattermost.com/guides/administrator.html#administration

People need this: 

![image](https://cloud.githubusercontent.com/assets/177788/24335881/36f0ecde-123a-11e7-91b5-f7f7105da92e.png)

But people get this: 

![image](https://cloud.githubusercontent.com/assets/177788/24335879/28fccaa8-123a-11e7-8951-5aac968a39e1.png)

JeffSchering:
Fixed in [Increased ToC maxdepth to 3 in admin section #1097](https://github.com/mattermost/docs/issues/1097)

PR #1097, opened 2017-04-17
Increased ToC maxdepth to 3 in admin section

JeffSchering:
Fixes [Make configuration settings more discoverable #1018](https://github.com/mattermost/docs/issues/1018)

### DIFF
--- a/source/guides/administrator.rst
+++ b/source/guides/administrator.rst
@@ -48,7 +48,7 @@ Administration
 --------------
 
 .. toctree::
-   :maxdepth: 2
+   :maxdepth: 3
    :glob:
 
    /administration/command*


esethna:
@JeffSchering can we please test this on a local copy before merging. I'm worried this will make the whole toc for the admin section unreadable.

Is there a way to only expand the TOC for this doc but not others in this section?

JeffSchering:
@esethna I always test on a local copy before creating a PR. The ToC for the Admin section is not unreadable. Most of the extra lines are in the Configurations Settings section, which is what's asked for in https://github.com/mattermost/docs/issues/1018.

There is a way to expand only the Configuration Settings section, but the side effect of that is extra vertical whitespace.

esethna:
Okay if you're happy with the appearance after testing then this should be okay to accept. Thanks!

PR #1099, opened 2017-04-18
Update toolkit.rst

jasonblais:
An estimated timeline for the toolkit from a discussion with Joram

### DIFF
--- a/source/developer/toolkit.rst
+++ b/source/developer/toolkit.rst
@@ -22,27 +22,30 @@ Features of the Developer Toolkit
 
 Below is a list of planned features of the developer toolkit with estimated time of delivery.
 
-1. Q2/2017:
+1. Completed:
+
+ - Webhooks and slash commands to allow easy, low-effort extension and integration.
+
+2. Q2/2017:
 
  - Mattermost HTTP REST APIv4 allowing for much more powerful server interaction.
  - Mattermost Webapp moved over to Redux infrustructure.
  
-2. Q3/2017:
+3. Q3/2017:
 
  - Mattermost Webapp UI components moved over to be Redux compatible.
  - The ability to build webapp client plugins to override existing UI components (replace posts with your custom components, use your own video services etc.), modify/extend client drivers to interact with custom server API endpoints, and add whole new UI views in predetermined places.
 
-3. Q4/2017:
+4. Q4/2017:
  - The ability to build Go 1.8 server plugins to hook directly into server events (think new post events, user update events, etc.), have some form of database access (possibly access to certain tables, and the ability to create new tables) and to add custom endpoints to extend the Mattermost REST API.
 
-4. Q1/2018:
+5. Q1/2018:
  - The ability to build plugins similar to the webapp but for React Native apps for iOS and Android.
  - A system or architecture to combine the above plugins into one easy-to-share and easy-to-install package.
 
-5. Q2/2018:
+6. Q2/2018:
  - A market or directory to find official and/or certified by Mattermost plugins and a process to get your plugin certified.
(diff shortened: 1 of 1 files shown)


jwilander:
Looks good to me, let me bring it up during the developer meeting for their opinions first then I'll give my final review

source/developer/toolkit.rst
```diff
@@ -20,15 +20,31 @@ Consequently this will save time for system administrators as well as developers
...
+ - The ability to build Go 1.8 server plugins to hook directly into server events (think new post events, user update events, etc.), have some form of database access (possibly access to certain tables, and the ability to create new tables) and to add custom endpoints to extend the Mattermost REST API.
+
+4. Q1/2018:
+ - The ability to build plugins similar to the webapp but for React Native apps for iOS and Android.
+ - A system or architecture to combine the above plugins into one easy-to-share and easy-to-install package.
+
+5. Q2/2018:
+ - A market or directory to find official and/or certified by Mattermost plugins and a process to get your plugin certified.
+ - A guide or system to enable the embedding of Mattermost into other apps as a chat service.
+ - Webhooks and slash commands to allow easy, low-effort extension and integration.
```
jwilander: Oh, we already have webhooks and slash commands so this should be considered complete. My mistake, I missed that when we reviewed earlier

jasonblais: Right, makes sense :)

Moved to 'completed' section, since it can be considered to be part of the toolkit.

jwilander:
Chatted with dev team, estimated timeline looks good

lfbrock:
@jwilander @jasonblais - do we really expect this to take a year, or is that a conservative estimate?

jwilander:
It's a bit conservative but there's a lot of work to do and we'll need to dedicate a good chunk of development mana to it if we want it done a lot sooner

lfbrock:
Hmm yeah it might be better to dedicate more resources to it rather than drag it out. We should figure out priority compared to other large items and consider hiring plan goals as well, I'll chat with Jason.

jasonblais:
It's a conservative estimate with the resources we have now. I expect the timeline to adjust as we start working on it and dedicate more resources to it.

Maybe for the long-term items it would be better to include them in an "Unscheduled" category rather than have scheduled times, i.e.:

```
1. Completed:

 - Webhooks and slash commands to allow easy, low-effort extension and integration.

2. Q2/2017:

 - Mattermost HTTP REST APIv4 allowing for much more powerful server interaction.
 - Mattermost Webapp moved over to Redux infrustructure.
 
3. Q3/2017:

 - Mattermost Webapp UI components moved over to be Redux compatible.
 - The ability to build webapp client plugins to override existing UI components (replace posts with your custom components, use your own video services etc.), modify/extend client drivers to interact with custom server API endpoints, and add whole new UI views in predetermined places.

4. Unscheduled:
 - The ability to build Go 1.8 server plugins to hook directly into server events (think new post events, user update events, etc.), have some form of database access (possibly access to certain tables, and the ability to create new tables) and to add custom endpoints to extend the Mattermost REST API.
 - The ability to build plugins similar to the webapp but for React Native apps for iOS and Android.
 - A system or architecture to combine the above plugins into one easy-to-share and easy-to-install package.
 - A market or directory to find official and/or certified by Mattermost plugins and a process to get your plugin certified.
 - A guide or system to enable the embedding of Mattermost into other apps as a chat service.

All of the documentation required to support the use and building of the toolkit will be done as we work through the above systems.
```

jasonblais:
Reviewed changes with Joram and Lindsay here: https://pre-release.mattermost.com/core/pl/o3k9jfp7wp8u3mxhoii9jm8chr

Should be fine to merge,

PR #1101, opened 2017-04-19
V3.9 documentation - Upgrade guide documentation

gluelue:
Update upgrade guide with the following:

1.  Run platform -version to check the current version of your Mattermost server to

1.  Run ./platform version in the /opt/mattermost/bin directory to determine your current version. 

Note that -version is deprecated and you do not need to use the dash in front of version.


gluelue:
Further, the following needs to be added to the upgrade guide so that there are correct ownership settings and write capabilities on /opt/mattermost after the upgrade file has been un-tarred.

Set the user and group mattermost as the owner of the Mattermost files:
sudo chown -R mattermost:mattermost /opt/mattermost

Give write permissions to the mattermost group:
sudo chmod -R g+w /opt/mattermost

JeffSchering:
Thanks @gluelue. It looks like you chose the v3.9-documentation branch to merge into master instead of the branch that contains the changes that you made.

The changes that you propose are ok, but they assume that people have installed Mattermost according to the current installation instructions in /opt/mattermost. However, because those instructions are relatively new, and for various other reasons, we can't assume a common installation directory when writing the upgrade doc.

Also, I'm rewriting the upgrade doc anyway right now, and it takes your proposed changes into consideration. You can see a related conversation about this in the Documentation channel on the Mattermost pre-release server at https://pre-release.mattermost.com/core/pl/1kpgxf8nctr6jqc6enangefsrc.

I'm going to close this PR primarily because the v3.9-documentation branch is not correct for this change. Feel free to join the Documentation channel and we can discuss further. Would you be interested in reviewing the new upgrade doc when I create the PR for it?

gluelue:
Sure would, glad to help!

Cheers,

Tom

PR #1112, opened 2017-04-22
Changing "remote" to "distributed"

it33:
"remote" implies a center from which people are remote, whereas distributed implies no center.


source/process/working-at-mattermost.rst
```diff
@@ -146,10 +150,11 @@ General Benefits
 Results matter more far more than hours in an office. Your time at Mattermost is flexible. Go see the dentist, visit a friend, walk your dog--no need to ask permission.
 
 If there's something urgent going on at work, just let teammates know your status.
+For most roles, we require the ability to join meetings between 8am and 1pm California time. The exception is people who are dedicated to support in other timezones, or team members that don't need to be in real-time discussions. 
 
-**2. Work from anywhere with a strong internet connection**
+**2. Work from anywhere (with good internet)** 
 
-The comfort of home, freedom to travel, no commute--learn more about our `remote work policies <https://docs.mattermost.com/process/working-at-mattermost.html#working-remotely>`_.
+The comfort of home, freedom to travel, no commute--learn more about our `work anywhere policies <https://docs.mattermost.com/process/working-at-mattermost.html#working-remotely>`_.
```
lfbrock: I this this link needs to be updated since the title of the section earlier in the doc is updated. Would also be good to switch it to a relative link.

lfbrock: @amyblais would you be able to update the link?

lfbrock:
Just a note: I believe there was some internal conversation previously on whether "remote" is a more common term ([see thread in Mattermost](https://pre-release.mattermost.com/core/pl/gecpeizhm7r5zquxa1f775sjmw)). Reviewing the doc, it seems like most references are actually switched to "work anywhere", so I think it's fine to make the change here and for recruiting purposes & job postings we can use "remote" if it works better.

source/process/working-at-mattermost.rst
amyblais: @lfbrock I updated the link.

Issue #989, opened 2017-03-15
Add link to troubleshooting process in all our install docs


it33:
Prominently and at the top 

https://www.mattermost.org/troubleshoot/

lindy65:
I've added a note to the troubleshooting guide to the docs I thought needed it.  Could whoever reviews the PRs let me know whether they think any other doc needs a reference to the troubleshooting guide?  Thanks!

PR #1116, opened 2017-04-22
Update smtp-email-setup.rst

lindy65:
Added a note referencing troubleshooting guide as per this issue:  https://github.com/mattermost/docs/issues/989

### DIFF
--- a/source/install/smtp-email-setup.rst
+++ b/source/install/smtp-email-setup.rst
@@ -5,6 +5,8 @@ SMTP Email Setup
 
 To run in production, Mattermost requires SMTP email to be enabled for email notifications and password reset for systems using email-based authentication.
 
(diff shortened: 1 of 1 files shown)


source/install/smtp-email-setup.rst
```diff
@@ -5,6 +5,8 @@ SMTP Email Setup
 
 To run in production, Mattermost requires SMTP email to be enabled for email notifications and password reset for systems using email-based authentication.
 
+If you have any problems installing, see the `troubleshooting guide <https://www.mattermost.org/troubleshoot/>`__. To submit an improvement or correction, click Edit at the top of this page.
```
JeffSchering: This should probably be "setting up" rather than "installing". Also, see https://docs.mattermost.com/install/install-ubuntu-1604.html for an example of this note.

PR #1117, opened 2017-04-22
Update desktop.rst

lindy65:
Added a note to reference the troubleshooting guide as per this issue: https://github.com/mattermost/docs/issues/989

### DIFF
--- a/source/install/desktop.rst
+++ b/source/install/desktop.rst
@@ -16,6 +16,8 @@ Below is a list of additional resources:
  - `Changelog <https://docs.mattermost.com/help/apps/desktop-changelog.html>`_
  - Contributor’s guide (coming soon)
  - `Source code <https://github.com/mattermost/desktop>`_
+ 
+If you have any problems installing Mattermost, see the `troubleshooting guide <https://www.mattermost.org/troubleshoot/>`_. To submit an improvement or correction, click Edit at the top of this page.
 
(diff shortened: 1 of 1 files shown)


source/install/desktop.rst
```diff
@@ -16,6 +16,8 @@ Below is a list of additional resources:
  - `Changelog <https://docs.mattermost.com/help/apps/desktop-changelog.html>`_
  - Contributor’s guide (coming soon)
  - `Source code <https://github.com/mattermost/desktop>`_
+ 
+If you have any problems installing Mattermost, see the `troubleshooting guide <https://www.mattermost.org/troubleshoot/>`_. To submit an improvement or correction, click Edit at the top of this page.
```
JeffSchering: This should probably be "installing the desktop app" instead of installing Mattermost. Also, this page already has a troubleshooting section. You should try to integrate this note with it.

source/install/desktop.rst
```diff
@@ -16,6 +16,10 @@ Below is a list of additional resources:
  - `Changelog <https://docs.mattermost.com/help/apps/desktop-changelog.html>`_
  - Contributor’s guide (coming soon)
  - `Source code <https://github.com/mattermost/desktop>`_
+ 
+.. note::
```
JeffSchering: Can you integrate this info with the troubleshooting section at the bottom of this doc? Maybe something like "If these troubleshooting tips are of no help, see..."

PR #1120, opened 2017-04-24
Document how to use message attachments with posts created with API

jwilander:


jwilander:
I'm 0/5 but I think the goal is to make it easier on the user and having it here is helpful. Feel free to make a PR to the API reference if you think it fits better there

JeffSchering:
Closing because this is API documentation that belongs on https://api.mattermost.com

rqc:
Hi, I don't see a cross-reference to a PR to the api reference repo. Could you provide this?

JeffSchering:
Hi @rqc It's on my to-do list but keeps getting pushed down... would you have time to create a PR for it in the API repo?

PR #1129, opened 2017-04-25
Added production docker instructions for Fedora

ivernus:
As per amendments in #1118 for local docker instructions on Fedora, addition of production docker instructions.

### DIFF
--- a/source/install/prod-docker.rst
+++ b/source/install/prod-docker.rst
@@ -70,15 +70,7 @@ Production Docker Setup on Fedora
        docker-compose build
        docker-compose up -d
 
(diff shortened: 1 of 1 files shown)


JeffSchering:
Thanks for this PR @ivernus! I chatted with the Product Managers and because Fedora is not supported in production and is not planned to be supported in production, we shouldn't have Fedora in the official production install docs.

I'm not sure where this info could go instead... maybe you could post it in the forum to help the other Fedora users out there. https://forum.mattermost.org/

ivernus:
thanks for the details @JeffSchering ill see if I can drop it in there anywhere.

PR #1131, opened 2017-04-25
Changed 'make inactive' to 'deactivate' and rewrote

JeffSchering:
This is for [[PLT-2181] Change wording for deactivating a user from "Make Inactive" to "Deactivate" #6106](https://github.com/mattermost/platform/pull/6106)


JeffSchering:
Not merging until 3.9 is released on May 16.

PR #1139, opened 2017-04-27
Adding a 404 page for docs.mattermost.com

JeffSchering:
Adding this page in preparation for server fix that will display a proper 404 instead of Access Denied.


asaadmahmood:
+1

PR #1096, opened 2017-04-17
Added a location entry for v4 API

JeffSchering:
Added a location for v4 API to the NGINX config, with contents identical to the /api/v3/users/websocket location. See https://pre-release.mattermost.com/core/pl/mrux4tmk37dnjgee9uccaoou8y

```
       location /api/v4/websocket {
           proxy_set_header Upgrade $http_upgrade;
           proxy_set_header Connection "upgrade";
           client_max_body_size 50M;
           proxy_set_header Host $http_host;
           proxy_set_header X-Real-IP $remote_addr;
           proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
           proxy_set_header X-Forwarded-Proto $scheme;
           proxy_set_header X-Frame-Options SAMEORIGIN;
           proxy_buffers 256 16k;
           proxy_buffer_size 16k;
           proxy_read_timeout 600s;
           proxy_pass http://backend;
       }
```

PR #1144, opened 2017-04-28
Update NGINX configuration for websocket connections

jasonblais:
This is an update to the original PR which added separate configuration for v3 and v4 endpoints: https://github.com/mattermost/docs/pull/1096

Note that this configuration change is already needed for Mattermost server v3.8 for anyone who wants to run the new React Native mobile apps. Hence, submitting against `master` instead of the 3.9 docs branch.

See discussion for context: https://pre-release.mattermost.com/core/pl/m4y8efh9d3gxiyc7rpybnk155r


jasonblais:
@JeffSchering propose another dev such as Harrison to take a look, since Joram shared the one-line change with me -- would be good to have a second pair of eyes on this from the dev side, to confirm the change.

jwilander:
Corey and Elias already looked at it but doesn't hurt to have another pair of eyes on it

PR #1145, opened 2017-04-28
Change SAML verification and encryption from "optional" to "recommended"

jasonblais:
Optional can make it sound like it's not that important, although we recommend verification and encryption to be enabled.

Related PR to update System Console help text, and the default settings, is here: https://github.com/mattermost/platform/pull/6330/

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -1089,10 +1089,12 @@ The public authentication certificate issued by your Identity Provider.
 
 Verify Signature
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-``true``: When true, Mattermost verifies that the signature sent from the SAML Response matches the Service Provider Login URL.
+**True**: Mattermost verifies that the signature sent from the SAML Response matches the Service Provider Login URL.
+
+**False**: Verification is disabled. Not recommended for production environments. For testing only.
 
 +----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-| This feature's ``config.json`` setting is ``"Verify": false`` with string input.                                                                                     |
+| This feature's ``config.json`` setting is ``"Verify": true`` with string input.                                                                                      |
 +----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
 Service Provider Login URL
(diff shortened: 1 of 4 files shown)


source/administration/config-settings.rst
```diff
@@ -1089,10 +1089,12 @@ The public authentication certificate issued by your Identity Provider.
...
 Verify Signature
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-``true``: When true, Mattermost verifies that the signature sent from the SAML Response matches the Service Provider Login URL.
+**True**: Mattermost verifies that the signature sent from the SAML Response matches the Service Provider Login URL.
+
+**False**: Verification is disabled. Not recommended for production environments. For testing only.
 
 +----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-| This feature's ``config.json`` setting is ``"Verify": false`` with string input.                                                                                     |
+| This feature's ``config.json`` setting is ``"Verify": true`` with string input.                                                                                      |
```
JeffSchering: ~with string input~ with options `true` and `false`.

source/administration/config-settings.rst
```diff
@@ -1089,10 +1089,12 @@ The public authentication certificate issued by your Identity Provider.
 
 Verify Signature
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-``true``: When true, Mattermost verifies that the signature sent from the SAML Response matches the Service Provider Login URL.
+**True**: Mattermost verifies that the signature sent from the SAML Response matches the Service Provider Login URL.
+
+**False**: Verification is disabled. Not recommended for production environments. For testing only.
```
JeffSchering: Maybe completely removing "Verification is disabled" might be more emphatic. 0/5

source/administration/config-settings.rst
```diff
@@ -1107,10 +1109,10 @@ Enable Encryption
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 **True**: Mattermost will decrypt SAML Assertions encrypted with your Service Provider Public Certificate.
 
-**False**: Encyption is disabled.
+**False**: Encyption is disabled. Not recommended for production environments. For testing only.
```
JeffSchering: typo... should be "Encryption". But "Encryption is disabled." doesn't seem like the correct statement anyway. When `true` decryption occurs, so when `false`, then decryption is not occurring. Maybe "Decryption is disabled." is correct?

Removing that sentence might be more emphatic.

source/administration/config-settings.rst
```diff
@@ -1107,10 +1109,10 @@ Enable Encryption
...
 
 +----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-| This feature's ``config.json`` setting is ``"Encrypt": false`` with options ``true`` and ``false`` for above settings respectively.                                  |
+| This feature's ``config.json`` setting is ``"Encrypt": true`` with options ``true`` and ``false`` for above settings respectively.                                   |
```
JeffSchering: ~for above settings respectively.~
Removing those words would make it inconsistent with the rest of the doc, but eventually I think they should all be removed as they are superfluous and add very little or nothing to understanding the doc.

source/administration/config-settings.rst
jasonblais: Do we have a principle/guideline we can follow for this one?

JeffSchering: I don't think so, but if the description starts out with "Not recommended..." then it's more clear that trouble awaits those who are not true.

PR #1146, opened 2017-04-28
Added note about user count

JeffSchering:
This is the note:

Make sure that the new license is for a number of users that is greater than or equal to the current total number of users on your system. To find the total number of users, go to the REPORTING section of the System Console and click **Site Statistics**. The total number of users is in the **Total Users** field. The license is rejected if this value is greater than allowed by the key.

### DIFF
--- a/source/install/ee-install.rst
+++ b/source/install/ee-install.rst
@@ -54,6 +54,6 @@ Make sure that the new license is for a number of users that is greater than or 
 
 1. In the OTHER section of the System Console, click **Edition and License**. The *Edition and License* page opens.
 2. Click **Remove Enterprise License and Downgrade Server**. This clears the license from the server and refreshes the System Console.
-3. Return to the *Edition and License* page and upload the new license key file.
+3. Upload the new license key file.
 
 After the key is uploaded and installed, the **Edition** and **License** sections on the page show the capabilities of your new license key.
--- a/source/install/ee-install.rst
+++ b/source/install/ee-install.rst
@@ -1,15 +1,15 @@
 ..  _ee-install:
 
 ===========================================
-Enterprise Install and Upgrade 
+Enterprise Install and Upgrade
 ===========================================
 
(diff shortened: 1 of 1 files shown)


source/install/ee-install.rst
```diff
@@ -48,10 +48,12 @@ Upgrading Enterprise Edition to a newer version
...
-To change an existing license key on an Enterprise Edition server to a new license key:
+Make sure that the new license is for a number of users that is greater than or equal to the current total number of users on your system. To find the total number of users, go to the REPORTING section of the System Console and click **Site Statistics**. The total number of users is in the **Total Users** field. The license is rejected if this value is greater than allowed by the key.
 
-1. In the **System Console**, go to **OTHER** > **Edition and License** > **License Key** and click "Remove Enterprise License and Downgrade Server". This clears the license from the server and refreshes the System Console. 
-2. In the **System Console**, go to **OTHER** > **Edition and License** > **License Key** and upload the new license key file.
-   
-The **Edition** and **License** sections on the page should update to show the capabilities of your new license key. 
+**To install a new license key**:
 
+1. In the OTHER section of the System Console, click **Edition and License**. The *Edition and License* page opens.
```
lindy65: "OTHER" > "**OTHER**" ?

JeffSchering: I was thinking that too, at first. But in the System Console, OTHER is not clickable; it's just a label. If it was something that you had to click in order to see its items, then **OTHER** would be correct. Does that make sense?

lindy65: Yep, makes sense, thanks :)

source/install/ee-install.rst
```diff
@@ -48,10 +48,12 @@ Upgrading Enterprise Edition to a newer version
...
+2. Click **Remove Enterprise License and Downgrade Server**. This clears the license from the server and refreshes the System Console.
+3. Return to the *Edition and License* page and upload the new license key file.
```
jasonblais: I think you're already on the "Edition and License" page after the page refreshes

lindalumitchell: Yes, @JeffSchering I just tested removing a license, and confirmed that after refresh, I am still viewing /admin_console/license.

PR #1148, opened 2017-05-01
Update config-tls-mattermost.rst with note on skipping section

dinofizz:
The added note makes it clear that you need not specify TLS settings in both the Mattermost application as well as in the NGINX configuration. If you follow the guide (without these changes) from top to bottom, for say Ubuntu 16.04, then it might be a bit confusing as to why you would be specifying all the TLS settings and certs in the app - if later the guide explains how to do this for NGINX SSL termination.

(I did not create a ticket for this work as reading the contribution guidelines it sounds like this satisfies the "if the change to existing behaviour is small" clause for submitting without a Help Wanted ticket)

### DIFF
--- a/source/install/config-tls-mattermost.rst
+++ b/source/install/config-tls-mattermost.rst
@@ -1,30 +1,33 @@
-Configuring TLS on the Mattermost Server
-========================================
+Configuring TLS on Mattermost Server
+====================================
 
-. Note::
-  You can skip this section if you will be using NGINX to specify the certificates in use and configure the TLS/SSL settings.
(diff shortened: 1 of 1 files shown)


JeffSchering:
Hi @dinofizz 

Thanks for this! Since the topic was being changed anyway, I went an rewrote it to give some more context to readers and to make it more compliant with the style guide. Let me know if my changes successfully incorporated your intent.

dinofizz:
@JeffSchering Thanks! That does read a bit nicer, and indeed provides the intended information I was aiming to relay.

PR #1155, opened 2017-05-03
Add documentation on how to build a webapp component

jwilander:


jwilander:
TODO: Need to add a documentation step

jwilander:
@hmhealey can you review the selector section when you have a chance?

@enahum same but for the action section?

Feel free to provide feedback on any other parts as well

source/developer/redux.rst
```diff
@@ -0,0 +1,261 @@
...
+
+  export function getUser(state, id) {
+      return state.entities.users.profiles[id];
+  }
+
+The above example is just simply pulling a user out of the profiles entity and requires no computation or formatting.
+
+If your selector needs to select based on some more advanced requirements or needs the result in a specific format then you'll need to make use of the ``createSelector`` function from `reselect <reselect <https://github.com/reactjs/reselect>`__. If you're not sure what this is good for, `check out the previously mentioned developer talk <https://www.youtube.com/watch?v=6N2X7gEwmaQ>`__. The short form reason is using reselect allows for memoization and only runs the computation of selectors when the state affecting that selector has actually changed.
+
+The basic usage for ``createSelector`` is pass it all the selector functions needed as inputs to your computation. The last argument will then be a function that takes in the results of each selector you specified and performs your computation on that data before returning it.
```
hmhealey: "is pass" -> "is to pass"
"The last argument will then" -> "The last argument should then"
"each selector you specified and performs your computation on that data before returning it." -> "each previous selector, performs some computations, and then returns the result."

source/developer/redux.rst
```diff
@@ -0,0 +1,261 @@
...
+              }
+          }
+
+          return usersByUsername;
+      }
+  );
+
+Here we're using the ``getUsers`` selector to feed users into our function that builds a map of users with username as the key.
+
+So far that's pretty straight forward but what if you want to select some data based on an argument but also need to do some computation? That is a little more tricky if you haven't wrapped your head around the purpose of reselect and how createSelector works, so if you haven't watched the developer talk linked above I would strongly suggest it.
```
hmhealey: The first sentence is a bit awkward with the double "but"s. Maybe remove the second one and the bit following it like:
"So far that's pretty straightforward, but what if you want to select some data based on an argument?"

"straight forward but what if" -> "straightforward, but what if"
"linked above I would" -> "linked above, I would"

source/developer/redux.rst
```diff
@@ -0,0 +1,261 @@
...
+              return fileIdsForPost.map((id) => allFiles[id]);
+          }
+      );
+  }
+
+  // Usage by a third party application
+  const getFilesForPost = makeGetFilesForPost();
+  const files = getFilesForPost(state, 'somepostid');
+
+This can look a bit confusing but there is little happening here we haven't seen before. All that we're doing is using three selectors with ``createSelector``, the third selector just happens to be returning an argument so that our final function has access to it. Remember that every selector always takes state in as the first argument.
```
hmhealey: "a bit confusing but there is" -> "a bit confusing, but there is"
"returning an argument" -> "returning its second argument"

source/developer/redux.rst
```diff
@@ -0,0 +1,261 @@
...
+
+If you're thinking, I don't get it why can't we just create the selector normally? Then think about how selectors work and remember that if the state changes then the computation happens again. In this case if our ``postId`` changes that counts as a state change since one of our selectors is returning it. This means every time we provide a different ``postId`` to our selector we lose all the benefits of memoization, which is the whole reason for using reselect. So instead we create a new selector for every post id that we want to select for. That may seem a little crazy at first, but think about how componentized React is and it's not that crazy. All you really need to do is use the factory function to create your selector in the constructor or container of your component and use it solely for that component.
```
hmhealey: "If you're thinking, I don't get it why can't we just create the selector normally? Then think about" -> `If you're thinking, "I don't get it. Why can't we just create the selector normally?" then think about`

"In this case if our `postId` changes that counts as a state change since one of our selectors is returning it. This means every time" -> "When the `postId` changes, that counts as a state change, so every time" (The sentences kind of serve the same purpose, so I think we could combine them)
"to our selector we lose all the benefits of memoization, which is" -> "to our selector, we lose all the benefits of memoization which is"

"So instead we create" -> "Instead, we create"
"a new selector for every post id that we want to select for." -> "copies of our selector everywhere we know the post id shouldn't change frequently." (Just to be clear that it's not the worst thing if post id changes)

"but think about" -> "but, if you think about"
"is and it's not that crazy" -> "is, it's not that bad" (comma and avoiding repeated word)
"to create your selector" -> "to create an instance of your selector"
"in the constructor or container of your component" -> "for each component"

source/developer/webapp-component.rst
```diff
@@ -0,0 +1,180 @@
+Building a Webapp React Component
+==========================
+
+This page describes how to build a new React component in the webapp. All new components must meet the following requirements:
+
+1. Be pure, meaning that all information required to render are passed in by props
```
hmhealey: "are passed" -> "is passed"

source/developer/webapp-component.rst
```diff
@@ -0,0 +1,180 @@
...
+2. Have no direct store interaction, use an index.js to wrap the component if needed
```
hmhealey: "use an index.js" -> "use a container"

source/developer/webapp-component.rst
```diff
@@ -0,0 +1,180 @@
...
+3. Have component tests
+4. Be generic and re-usable when possible
+5. Have documented prop types
```
hmhealey: Maybe change this to "Have documented props" since we want the purpose of the props to be clear

source/developer/webapp-component.rst
```diff
@@ -0,0 +1,180 @@
...
+
+- https://facebook.github.io/react/tutorial/tutorial.html
+- http://redux.js.org/
+
+The components for the webapp currently live in the ``webapp/components/`` directory in the `platform repository <https://github.com/mattermost/platform>`__.
+
+Designing your Component
+------------------
+
+The most important part of designing your component is decidiing on what the props will be. Props are very much the API for your component, think of them as a contract between your component and the users of it.
```
hmhealey: "decidiing" -> "deciding"

source/developer/webapp-component.rst
```diff
@@ -0,0 +1,180 @@
...
+1. All components must extend ``React.PureComponent``
+2. All data for rendering must come from props
+3. Local rendering information that does not affect the store may be kept in state
+
+Our ItemList example might look something like this:
+
+.. code-block:: javascript
+  :linenos:
+
+  export default class ItemList extends React.PureComponent {
```
hmhealey: `React.PureComponent` -> `PureComponent` if we're following the style of the RN app. I'm 0/5 which way we go though

source/developer/webapp-component.rst
```diff
@@ -0,0 +1,180 @@
...
+2. Needs to be able to perform some sort of action that affects the state of the store
+
+Continuing the ItemList example above, maybe our parent component doesn't care about our list of items and doesn't have access to them. Let's also imagine that we want to let the user remove items from the list by clicking on them. This means our component now needs a container for both criteria above and our props will change slightly:
+
+.. code-block:: javascript
+  :linenos:
+
+  static propTypes = {
+      /**
+       * Sets the title of the list
```
hmhealey: What about using either single line comments or just `/*block comments like this*/` instead of having these on multiple lines? I don't like the extra vertical space used just for the  start and end of the comments

jwilander: 0/5 but I've already started using it like that and that is what react-bootstrap and every one else was doing

source/developer/webapp-component.rst
jwilander: I've been using `React.PureComponent` everywhere

jwilander:
@hmhealey changes made. Maybe just commit your spelling/grammar fixes directly next time though, it's a bit annoying to copy/paste changes in :P

source/developer/webapp-component.rst
hmhealey: Okay, I'll go back to that on the web app then. It'll save us an import, especially when prop types becomes its own library

source/developer/webapp-component.rst
hmhealey: Alright, well it doesn't matter too much anyway. It's not like Javadocs where the exact type of comment matters

PR #1156, opened 2017-05-03
Added a password field to the user object

JeffSchering:
This is for [Bulk import password #6268](https://github.com/mattermost/platform/pull/6268)

### DIFF
--- a/source/deployment/bulk-loading-data-format.rst
+++ b/source/deployment/bulk-loading-data-format.rst
@@ -284,8 +284,8 @@ For clarity, the object is shown using regular JSON formatting, but in the data 
     user: {
       username: "username",
       email: "email@example.com",
-      auth_service: "password",
-      auth_data: "ldap_id",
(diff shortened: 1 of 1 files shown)


source/deployment/bulk-loading-data-format.rst
```diff
@@ -366,6 +367,13 @@ Fields of the User object
       <td align="center" valign="middle">No</td>
     </tr>
     <tr class="row-odd">
+      <td valign="middle">password</td>
+      <td valign="middle">string</td>
+      <td>A password for the user. If not present, the bulk loader generates a password.</td>
```
grundleborg: There are two other things we should note here:
1. The password field (or autogenerated password if the field is omitted) is only applicable if `auth_service` is empty (i.e. Password Authentication is the selected auth service).
2. `auth_data` and `password` are mutually exclusive. You can only use one in a given `User`.

PR #1166, opened 2017-05-08
Added enable searching of posts and fixed typos

JeffSchering:
This is for [Adding the ability to disable full text search queries for master #6102](https://github.com/mattermost/platform/pull/6102)

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -2320,9 +2320,18 @@ This setting defines the frequency of cluster request time logging for :doc:`../
 | This feature's ``config.json`` setting is ``"ClusterLogTimeoutMilliseconds": 2000`` with whole number input.                                                         |
 +----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
+Enable Searching of Posts
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This setting enables the ability for users to search messages. If not enabled, when users attempt to use the search box they get an error message.
+
++---------------------------------------------------------------------------------------------------------------------------------------------+
+| This feature's ``config.json`` setting is ``"EnablePostSearch": true`` with options ``true`` and ``false`` for above settings respectively. |
++---------------------------------------------------------------------------------------------------------------------------------------------+
+
 Enable User Typing Messages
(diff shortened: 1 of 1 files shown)


source/administration/config-settings.rst
```diff
@@ -2320,9 +2320,18 @@ This setting defines the frequency of cluster request time logging for :doc:`../
 | This feature's ``config.json`` setting is ``"ClusterLogTimeoutMilliseconds": 2000`` with whole number input.                                                         |
 +----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
+Enable Searching of Posts
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This setting enables the ability for users to search messages. If not enabled, when users attempt to use the search box they get an error message.
```
lindy65: Thought about this paragraph for quite a while and, to me, it just doesn't read smoothly...

How about:
`If this setting is enabled, users can search messages.  If it's disabled, users get an error message when they attempt to use the search box.`

JeffSchering: Done. Thanks!

source/administration/config-settings.rst
```diff
@@ -2320,9 +2320,18 @@ This setting defines the frequency of cluster request time logging for :doc:`../
 | This feature's ``config.json`` setting is ``"ClusterLogTimeoutMilliseconds": 2000`` with whole number input.                                                         |
 +----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
+Enable Searching of Posts
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If this setting is enabled, users can search messages. If it's disabled, users get an error message when they attempt to use the search box.
```
jasonblais: 1) Maybe add a note it's for improving performance?

E.g. if lots of users do searching in a deployment with millions of messages, it has the potential of causing the server to slow down

2) Also propose adding a note that if you're looking to improve performance, you can first try to use the search read replicas, then disable search

JeffSchering: 1. Ok. But it seems a bit odd... "If the feature is too popular, turn it off." :)
2. I'm not sure about mentioning search replicas. They are only available in E20 with a cluster config.

PR #1171, opened 2017-05-10
Clarified the value to use for team name in CLI

JeffSchering:
Comes from this forum post: https://forum.mattermost.org/t/solved-how-can-i-delete-teams/220/12

The **platform team** commands require a team name, but they don't work with the value that is displayed in the UI (DisplayName column in the Teams table). Instead, they use the value that is stored in the Name column, which is almost always different.

### DIFF
--- a/source/administration/command-line-tools.rst
+++ b/source/administration/command-line-tools.rst
@@ -374,10 +374,9 @@ platform team
 .. note::
     **{team-name} value**
 
-    For the *add*, *delete*, and *remove* commands, the *{team-name}* value comes from the *Name*  column in the *Teams* database, and is not the name that is displayed in the UI.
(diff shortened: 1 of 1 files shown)


source/administration/command-line-tools.rst
```diff
@@ -369,16 +369,26 @@ platform team
...
     -  `platform team remove`_ - Remove users from a team
 
+.. _team-value-note:
+
+.. note::
+    **{team-name} value**
+
+    For the *add*, *delete*, and *remove* commands, the *{team-name}* value comes from the *Name*  column in the *Teams* database, and is not the name that is displayed in the UI.
+
+    You can usually derive the *{team-name}* from the display name by converting the display name to all lowercase letters and replacing spaces with dashes. For example, *My Team* becomes *my-team*. This works for teams that were created using the UI, but won't work if a custom value for the team name was specified when using the `platform team create`_ command. In that case, you must examine the database to find out the team name.
```
JeffSchering: That's much better... thanks!

PR #1180, opened 2017-05-13
Update smtp-email-setup.rst

docdawning:
TLS Connection Security for gmail doesn't work for me, but STARTTLS does. Maybe this is particular to the fact I'm using Google Apps and therefore my email addresses aren't @gmail.com, but my own domain. I suspect that STARTTLS is actually what everyone's going to need to use gmail's SMTP at this point. I tried TLS many times and it failed every time until I switched to STARTTLS.


docdawning:
@mattermod Okay, I've submitted your contributor form thing

PR #1184, opened 2017-05-16
Update RHEL7 Documentation

kjkeane:
The command `sudo service postgresql initdb` doesn't work within CentOS 7.1, you receive the following error

`The service command supports only basic LSB actions (start, stop, restart, try-restart, reload, force-reload, status). For other actions, please try to use systemctl.`

 I have updated this to `/usr/pgsql-9.4/bin/postgresql94-setup initdb`

### DIFF
--- a/source/install/install-rhel-71-postgresql.rst
+++ b/source/install/install-rhel-71-postgresql.rst
@@ -19,7 +19,7 @@ Installing PostgreSQL Database
 
 5. Initialize the database.
 
-  ``sudo service postgresql initdb``
+  ``sudo /usr/pgsql-9.4/bin/postgresql94-setup initdb``
 
 6. Set PostgreSQL to start on boot.
 
--- a/source/install/install-rhel-71-postgresql.rst
+++ b/source/install/install-rhel-71-postgresql.rst
@@ -7,7 +7,14 @@ Installing PostgreSQL Database
 
 2. Download the PostgreSQL 9.4 Yum repository.
 
-  ``curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-redhat94-9.4-3.noarch.rpm``
+  For RHEL 7
+    ``curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-redhat94-9.4-3.noarch.rpm``
+  For CentOS 7
+    ``curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-centos94-9.4-3.noarch.rpm``
+  For Scientific Linux
+    ``curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-sl94-9.4-3.noarch.rpm``
+  For Oracle
+    ``curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-oraclelinux-9.4-3.noarch.rpm``
 
 3. Install the Yum repository from the file that you downloaded.
 
--- a/source/install/install-rhel-71-postgresql.rst
+++ b/source/install/install-rhel-71-postgresql.rst
@@ -31,7 +31,7 @@ Installing PostgreSQL Database
 
 8. Switch to the *postgres* Linux user account that was created during the installation.
 
-  ``sudo --login --user postgres``
+  ``sudo su postgres``
 
 9. Start the PostgreSQL interactive terminal.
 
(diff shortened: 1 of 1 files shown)


kjkeane:
I have completed the license agreement as procedure, but there are few more changes I have made.  This came up because I currently attempting to migrate Mattermost from Ubuntu 16.04 to CentOS 7.

### Changes
* Update `.conf` locations
* Update service name when using `systemctl`
* Update `sudo` command to from `sudo --login --user postgres` to `sudo su postgres`

JeffSchering:
Hi @kjkeane thanks!

Are these changes strictly for CentOS, or do they apply to RHEL too?

kjkeane:
So far only CentOS since I do not have RHEL image to configure, but from my experience they are pretty similar.  I am going to see if I can do this in a RHEL evaluation image.

kjkeane:
Hey @JeffSchering I was able to confirm this morning that they are the same.  I used an evaluation ISO (RHEL 7.3) I got from Red Hat, but I had to use `curl` instead of `wget` to download the rpm for Postgres, because `wget` is not installed by default in CentOS or RHEL.

JeffSchering:
Thanks again @kjkeane! Would you mind also updating the download step to use curl instead of wget?

kjkeane:
@JeffSchering No problem, I have updated `wget` to `curl -O` in the documentation.

source/install/install-rhel-71-postgresql.rst
```diff
@@ -19,19 +19,19 @@ Installing PostgreSQL Database
...
 
 7. Start the PostgreSQL server.
 
-  ``sudo systemctl start postgresql``
+  ``sudo systemctl start postgresql-9.4``
 
 8. Switch to the *postgres* Linux user account that was created during the installation.
 
-  ``sudo --login --user postgres``
+  ``sudo su postgres``
```
JeffSchering: Ok. `--login` and `--user` work with the latest version of su, but that might not be on RHEL 7.1

The equivalent short options are `-i` and `-u`. The `-i` option is different than the `-s` option though... I think it changes you to the postgres user directory, among other things.

If you can confirm that `su` works and that `iu` is not needed, then we can keep `su`.

source/install/install-rhel-71-postgresql.rst
```diff
@@ -72,19 +72,19 @@ Installing PostgreSQL Database
   
     ``listen_addresses = '*'``
 
-16. If the Mattermost server is on a separate machine, modify the file ``pg_hbe.conf`` to allow the Mattermost server to communicate with the database.
+16. If the Mattermost server is on a separate machine, modify the file ``pg_hba.conf`` to allow the Mattermost server to communicate with the database.
 
   If the Mattermost server and the database are on the same machine, then you can skip this step.
 
-  a. Open ``/etc/postgresql/9.4/main/pg_hba.conf`` in a text editor.
+  a. Open ``/var/lib/pgsql/9/4/data/pg_hba.conf`` in a text editor.
```
JeffSchering: Should this be `/var/lib/pgsql/9.4/data/pg_hba.conf`? -- 9.4 instead of 9/4

source/install/install-rhel-71-postgresql.rst
```diff
@@ -7,7 +7,7 @@ Installing PostgreSQL Database
 
 2. Download the PostgreSQL 9.4 Yum repository.
 
-  ``wget https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-redhat94-9.4-3.noarch.rpm``
+  ``curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-redhat94-9.4-3.noarch.rpm``
```
JeffSchering: Sorry, one last thing (I hope).

It looks like PostgreSQL provides a separate rpm for each distro that these instructions apply to https://yum.postgresql.org/repopackages.php

```
https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-redhat94-9.4-3.noarch.rpm
https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-centos94-9.4-3.noarch.rpm
https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-sl94-9.4-3.noarch.rpm
https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-oraclelinux94-9.4-3.noarch.rpm
```
The format for .rst should look like this:

```
For RHEL 7
    curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-redhat94-9.4-3.noarch.rpm
For CentOS 7
    curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/centos-7-x86_64/pgdg-redhat94-9.4-3.noarch.rpm

etc
```
Can you add that in?

kjkeane: I can definitely do that.  Just give me a few minutes.

jasonblais:
Big thanks @kjkeane and @JeffSchering!

PR #1199, opened 2017-05-20
Self-awareness as a value

it33:
Circling back on additional value that I think we have already, and need to filter for in hiring.


it33:
Approved in leads meeting today

PR #1204, opened 2017-05-22
Update training.rst

lindy65:
Added note to Avatar section re character being 'human-sized'...


source/process/training.rst
```diff
@@ -556,6 +556,7 @@ To have a Mattermost avatar created, you'll be invited to create a Mattermost av
 
 1. Please use the email subject "[YOUR_FULL_NAME] as [CHARACTER_NAME]", for example "Corey Hulen as Han Solo". 
 2. Attach a clear image at least 600 pixels high and 600 pixels wide showing your character's full body in a standing pose. 
+   - Note: Character should be human-sized (no giant characters).
```
JeffSchering: The reStructured text formatting treats the note as a new sentence of the previous paragraph instead of as its own paragraph. To fix this, you can either put another line between the note and the previous paragraph (remove the - too), or you can do this:

"... full body in a standing pose. Your character should be human-sized; no giant characters."

PR #1216, opened 2017-05-26
Updated PostgreSQL installation instructions

JeffSchering:
Added a step to restart the database after changing the listen_addresses setting, and revised instructions on how to connect to the database from localhost and from another machine.


JeffSchering:
@jwilander 

It's a side-effect of each OS having a different default version of PostgreSQL in their repos, and whether the OS uses yum or apt, and what the default service daemon is on each OS (upstart or systemctl). The alternative is to have one file with a convoluted mess of if statements.

PR #1217, opened 2017-05-26
Added note about ability to turn off message editing

JeffSchering:


source/overview/security.rst
```diff
@@ -36,17 +36,17 @@ Transmission Security
...
+   - Ability to exclude or include the `contents of messages in push notifications <https://docs.mattermost.com/administration/config-settings.html#push-notification-contents>`_ to avoid disclosure on locked mobile screens, and via relay servers from Apple and Google when sending notifications to iOS or Android mobile apps (relevant to compliance standards such as HIPAA)
 
 Integrity & Audit Controls
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   - By default, Mattermost stores a complete history of messages, including edits and deletes, along with all files uploaded. User interface actions for "deleting" messages and channels remove the data only from the user interface; the data is retained within your database.
+   - By default, Mattermost stores a complete history of messages, including edits and deletes, along with all files uploaded. User interface actions for "deleting" messages and channels remove the data only from the user interface; the data is retained within your database. Furthermore, you can turn off the ability for users to edit their messages after they're posted.
```
lfbrock: Maybe we can go with something like: 

"If your compliance guidelines require it, you can optionally turn off the ability to edit and delete messages after they are posted." 

0/5 - I think the idea is to explain why people would want to turn off edit/delete, and we should probably include the delete in there as well as edit.

JeffSchering: Done.

source/overview/security.rst
```diff
@@ -43,7 +43,7 @@ Transmission Security
 Integrity & Audit Controls
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   - By default, Mattermost stores a complete history of messages, including edits and deletes, along with all files uploaded. User interface actions for "deleting" messages and channels remove the data only from the user interface; the data is retained within your database. Furthermore, you can turn off the ability for users to edit their messages after they're posted.
+   - By default, Mattermost stores a complete history of messages, including edits and deletes, along with all files uploaded. User interface actions for "deleting" messages and channels remove the data only from the user interface; the data is retained within your database. If your compliance guidelines require it, you can turn off the ability of users to edit and delete their messages after they are posted.
```
lfbrock: Should this be "of users" or "for users?

PR #1218, opened 2017-05-26
Include zip install instructions

kjcsb1:


kjcsb1:
Done

JeffSchering:
Thanks @kjcsb1!

Can you tell us a bit about why this change is needed? Is it for building Windows packages? This doc: https://docs.mattermost.com/developer/dev-setup-compiling.html has that info already.

If it's easy to miss in the compiling doc, then maybe we can move it to the install docs, but then it should be in the OSX and Arch docs too.

kjcsb1:
When I ran make on Ubuntu it failed because zip wasn't installed. The current instructions suggest that it is only necessary if make packages is run. Happy to go with how you think it best to document it.

JeffSchering:
@kjcsb1 Thanks! You make a good point.

Also, when cross-compiling, it's more than just *zip* that's missing. There's also a couple of commands that need to be run, as described in the **make package** section of https://docs.mattermost.com/developer/dev-setup-compiling.html.

It's probably a good idea to move those additional setup instructions from that file to a Step 9 (or more) in the dev setup docs. Is that something you have time to do? If not, I'm happy to do it myself.

hmhealey:
@kjcsb1 Do you mean that zip was needed when you ran a command other than `make package`?

Including zip as part of the regular dev setup instructions is fine since it's fairly common anyway, but I don't think we need to include the requirements for cross-compilation unless the user actually needs them for `make package`

JeffSchering:
@hmhealey I think that because `package` is in the default target in the Makefile, it's probably a good idea to have the cross-compile setup in the dev setup instructions. When people get to the end of the dev setup, they might think that they're good to go, and they run `make` under the assumption that the default target is the one they want.

hmhealey:
Perhaps we should change the default target then. I'd like to simplify the dev setup if possible, not add stuff that isn't really necessary. I'll queue it for the dev meeting to see how people feel about just making it `make run`

hmhealey:
We're going to change the default makefile target so that it won't need these extra dependencies. Propose closing this PR then

JeffSchering:
Thanks @kjcsb1! Although your PR isn't going to get merged into the docs, it resulted in a change in the design of the makefile... a much larger impact. Thanks for contributing!

PR #1226, opened 2017-05-30
Removed port number from siteURL spec

JeffSchering:


JeffSchering:
@jasonblais I added a note about when a port is required, based on https://pre-release.mattermost.com/core/pl/bzppox6rqf8fiytjh68w4dzbpo

>...  modify it to make it clear that the port is required if not using standard ports such as 80 and 443.

PR #1237, opened 2017-06-06
Missing spaces added

ponsfrilus:
Missing spaces added to fix the backtick syntax which was broken.


ponsfrilus:
@JeffSchering I've signed the CLA and I'm not sure if you expect me to do something else.

PR #1246, opened 2017-06-08
PLT-6673: Google Analytics Reintegration

shieldsjared:
At the request of lindsay, corrected the google analytics build errors by abandoning the usage of the plugin and injecting the GA script directly.  Since the script is very small and hasn't changed in _years_, it seemed simpler and safe to directly inject over fixing the plugin.  

GA can still be configured through the `conf.py` file to disable/enable and specify the GA key.

https://mattermost.atlassian.net/browse/PLT-6673


JeffSchering:
Thanks @shieldsjared!

PR #1262, opened 2017-06-14
Add a note that Firefox ESR 45 is no longer supported

jasonblais:
Firefox ESR 45 had it's end of life on June 13: https://en.wikipedia.org/wiki/Firefox_version_history


jasonblais:
@esethna We do here: https://docs.mattermost.com/install/requirements.html#pc-web-experience

Do we need to mention it for each release?

PR #1270, opened 2017-06-15
Localization process updates

jasonblais:
Add a note that 
 - strings with `mobile` prefix are from the RN app
 - PR is submitted to both platform and mattermost-mobile repos


jasonblais:
@enahum let me know what you think? 

I'll let the localization community know as well once this PR is approved

PR #1276, opened 2017-06-16
Added Debian install instructions to ToC

JeffSchering:


jasonblais:
Do we want to add unofficial docs to the official install TOC? I'm not sure if we added Apache2 proxy config, but if we did, I think we shouldn't have.

Maybe there can be a separate section for unofficial community guides like Windows install, Debian install, Apache2 proxy config

JeffSchering:
I think it's a good idea to have the Debian docs in the ToC where they can be found. As it is now, they're hidden unless you have a direct link or use the search. I didn't know there was a Windows install until I searched for it. In the ToC its in the "Unofficial Enhancements" section of the Developer's Guide.

jasonblais:
Right now scanning the TOC it looks like they're all official

image

Sure we have a note in the guide, but not everyone will read it and simply scan straight to the instructions.

Maybe we have `(Unofficial)` in the front of the title? Or we have a separate TOC for unofficial install and configuration guides (where we list Apache2, Debian, Windows)?

0/5 Just think it needs to be immediately obvious that it's "unofficial" and not after-the-fact.

JeffSchering:
I think with more visibility we'll get greater participation from the community. For eg, the Apache docs have already attracted at least one PR, but the Windows install docs go virtually unnoticed. (Between Jan 1 and June 20, of the ~102,000 page views on the docs.mattermost.com/install/ path, only 900 of them were the Windows install, compared with ~13,500 for Ubuntu).

JeffSchering:
@jasonblais I added "Unofficial" to the title and ToC entry,

PR #1278, opened 2017-06-19
Revert "Deprecating G Suite and O365 SSO from E20"

it33:
Reverts mattermost/docs#1263


it33:
Need to think about this more...

PR #1282, opened 2017-06-20
Rewrote upgrade guide and split into four files

JeffSchering:
This is an attempt to make following the upgrade instructions easier, especially for those who are upgrading from very old versions such as 3.0.

I split it into four files:

**`upgrade.rst`** - original file. Has a table where you look up your current version to find out which special notices apply to you and which set of instructions to follow.
**`upgrading-to-latest.rst`** - instructions for upgrading from version 3.0 and newer
**`upgrading-to-3.0.rst`** - for upgrading from 2.0.0, 2.1.0, and 2.2.0 to version 3.0.3. (you must upgrade to 3.0.3 before upgrading to any higher version because of a special database upgrade script that must be run, which allows one userid across all teams)
**`upgrading-to-2.0.rst`** - when upgrading from very early versions, you must upgrade in sequence until you hit version 2.0.

I reviewed the changelog for each release to make sure any breaking changes were in the Important Notices section.

I tested the instructions by installing version 2.0.0, then upgrading to 3.0.3, and then upgrading to 3.10.0.


source/administration/upgrade.rst
```diff
@@ -1,120 +1,75 @@
...
-Location of your local storage directory
-  The local storage directory contains all the files that users have attached to their messages. If you don't know its location, open the System Console and go to **Files > Storage** and read the value in **Local Storage Directory**. Relative paths are relative to the ``mattermost`` directory. For example, if the local storage directory is ``./data/`` then the absolute path is ``{install-path}/mattermost/data``.
-Owner and group of the install directory - *{owner}* and *{group}*
-  Use the ``ls -l {install-path}/mattermost/bin/platform`` command to get the owner and group.
+In most cases you can upgrade Mattermost Server in a few minutes, but the upgrade can take longer depending on several factors, including the size and complexity of your installation, and the version that you're upgrading from.
+
+.. _upgrade-table:
+
+In the following table, look up your current version, review the applicable Important Notices, then click the link to the upgrade instructions.
```
esethna: How do people find what their current version is?

source/administration/upgrade.rst
```diff
@@ -1,120 +1,75 @@
...
+
+.. csv-table::
+  :header: "If your current version is:", "Read these Important Notices:", "Use these instructions:"
+  :widths: auto
+
+  "3.9.0", "Not Applicable", "`Upgrading to the latest version <../administration/upgrading-to-latest.html>`_"
```
esethna: Has this table concept been reviewed from an "administrator" perspective? Are we basing this concept off of any examples in other software docs?

source/administration/upgrade.rst
```diff
@@ -1,120 +1,75 @@
...
+   :maxdepth: 1
+
+   /administration/upgrading-to-latest.rst
+   /administration/upgrading-to-3.0.rst
+   /administration/upgrading-to-2.0.rst
 
 Important Notices
 -----------------
 
-1. Security related changes were made in 3.9.0 that cause any previously created team invite links, password reset links, and email verification links to no longer work. You must update any place where you have published these links.
```
esethna: "After upgrading to version 3.4" > I think this means version 3.4 and later, from any version before 3.4 (need to figure out wording). I think we need to be explicit in describing when people will hit the notes

source/administration/upgrade.rst
```diff
@@ -1,120 +1,75 @@
...
+1. After upgrading to version 3.4: If public links are enabled, existing public links will no longer be valid. This is because in earlier versions, existing public links were not invalidated when the Public Link Salt was regenerated. You must update any place where you have published these links.
 
-2. Security-related changes were made in 3.7.5 and 3.8.0 that require you to verify settings in the System Console before upgrading from version 3.7.4 and earlier.
+2. Enterprise Edition only: After upgrading to version 3.6.0, if you previously had values set for *RestrictPublicChannelManagement* and *RestrictPrivateChannelManagement*, the new settings for *RestrictPublicChannelCreation*, *RestrictPrivateChannelCreation*, *RestrictPublicChannelDeletion*, and *RestrictPrivateChannelDeletion* will take those settings as their default values.
```
esethna: Same note as above on version numbers

source/administration/upgrade.rst
```diff
@@ -1,120 +1,75 @@
...
 
-  1. In the GENERAL section of the System Console, click **Configuration** and make sure that the **Site URL** is specified. It must not be empty. For more information about SiteURL, see `Configuration Settings <config-settings.html#site-url>`_
-  2. In the GENERAL section of the System Console, click **Logging** and make sure that the **File Log Directory** field is either empty or has a directory path only. It must not have a filename as part of the path.
+3. In version 3.6, update the maximum number of files that can be open:
```
esethna: "update the maximum number of files that can be open" seems confusing. Do we need to describe it in this way?

Also, does this step require action or is it just verifying? The two verbs are conflicting "update" and "verify"

source/administration/upgrade.rst
```diff
@@ -1,120 +1,75 @@
...
-
-  Enterprise Edition
-    ``wget https://releases.mattermost.com/X.X.X/mattermost-X.X.X-linux-amd64.tar.gz``
-  Team Edition
-    ``wget https://releases.mattermost.com/X.X.X/mattermost-team-X.X.X-linux-amd64.tar.gz``
-
-4. Extract the Mattermost Server files.
-
-  ``tar -xzf mattermost*.gz``
+6. Security-related changes were made in 3.7.5 and 3.8.0 that require you to verify settings in the System Console before upgrading from version 3.7.4 and earlier.
```
esethna: Security-related changes > Security patches

source/administration/upgrade.rst
```diff
@@ -1,120 +1,75 @@
...
-
-After the server is upgraded, users might need to refresh their browsers to experience any new features.
-
-Upgrading Team Edition to Enterprise Edition
---------------------------------------------
+  1. In the GENERAL section of the System Console, click **Configuration** and make sure that the **Site URL** is specified. It must not be empty. For more information about SiteURL, see `Configuration Settings <config-settings.html#site-url>`_
+  2. In the GENERAL section of the System Console, click **Logging** and make sure that the **File Log Directory** field is either empty or has a directory path only. It must not have a filename as part of the path.
 
-To upgrade from the Team Edition to the Enterprise Edition, follow the normal upgrade instructions above, but make sure that you download the Enterprise Edition in Step 3.
+7. Security related changes were made in version 3.9.0 that cause any previously created team invite links, password reset links, and email verification links to no longer work. You must update any place where you have published these links.
```
esethna: Security related changes > Security patches

source/administration/upgrading-to-2.0.rst
```diff
@@ -0,0 +1,82 @@
+Upgrading to version 2.0
+========================
+
+To upgrade an early edition of Mattermost to the latest edition, you must first upgrade in a specific sequence until you get to version 2.0.0. After that, you must upgrade to version 3.0.3, and from there you can upgrade directly to the latest Mattermost version.
```
esethna: Propose putting something here about which server versions this guide is for. And if they have a different server version than that mentioned, which guides w=should they be using instead. It makes me nervous to have a lot of different guides without explicitly stating what each are for

- [ ]

source/administration/upgrading-to-3.0.rst
```diff
@@ -0,0 +1,113 @@
+Upgrading to version 3.0
+========================
+
+Use these instructions if you are upgrading from 2.0.0, 2.1.0, or 2.2.0. You must upgrade to 3.0 before upgrading to the latest version of Mattermost.
```
esethna: We seem to be inconsistent in there docs when revering to verison numbers. Sometimes it;s vX.X, other times it's X.X and other times it's version X.X. We need to pick one and be consistent

lfbrock:
Before reviewing further, can we pause on this and make sure it's headed in the right direction? 

I'm worried about the complexity of the first page, I think people upgrading from very old versions is a rare case so I don't know if it makes sense to have them all so prominently placed on the first page people are sent to when they're trying to upgrade. 

The important notices is also a bit hard to follow, not sure if we can make it easier somehow.

lfbrock:
Also can we get @jasonblais reviewing instead of ES, since this is more of a System Admin document? 

(Currently the PM owners are Eric on end user features, Jason on System Admin and Integrations). 

After review it should be signed off by Corey, and I would like to review as well.

JeffSchering:
I don't think the first page is all that complex, especially when compared to the upgrade of [Wordpress](https://codex.wordpress.org/Upgrading_WordPress_-_Extended_Instructions) or [Apache](https://httpd.apache.org/docs/2.4/upgrading.html) or [IBM Connections](http://www-01.ibm.com/support/docview.wss?uid=swg21972646). 

And the instructions can't be any less complex than the process they describe unless you want to leave information out. For example, we could get rid of the Important Notices section altogether, but then the amount of successful upgrades would decrease, and the amount of support emails and forum posts about failed upgrades would increase.

It's definitely not as simple as [Gitlab](https://about.gitlab.com/update/) (`sudo apt-get install gitlab-ce`), but on the spectrum of complexity, upgrading Mattermost is closer to the GitLab scenario than the Wordpress, Apache, or IBM Connections scenario.

To upgrade, look up your current version on the table, see if any special instructions apply, and click the link to the appropriate upgrade page.

I don't feel that the older versions are prominent; they are at the bottom of the list and are not highlighted in any way. I think that whether or not we support an upgrade from an older version is a PM decision, not a doc decision, and I can easily remove those instructions.

The instructions here are not additional. They've only been re-organized; for the most part, the text was copy/pasted from the existing upgrade guide. The table merely makes them more visible than they were before.

jasonblais:
@JeffSchering 

1) Propose we come up with best practices or guidelines for the upgrade guide, that we can then use for others as well (e.g. install guides). I think that could help determine if the current direction is the right way to go.

2) I'm not sure the goal mentioned in the PR description is what we should be aiming for

> This is an attempt to make following the upgrade instructions easier, especially for those who are upgrading from very old versions such as 3.0.

~80% of deployments are running 3.6 or later, 97% are running 3.0 or later. 

So for the majority of admins, all they really want is to go straight to upgrading to the latest Mattermost version. And along the way be informed of potential breaking changes like the NGINX config change.

JeffSchering:
@jasonblais 

1. Is the Style Guide not sufficient for that? What do you have in mind?

2. I understand that only ~20% of deployments are running versions less than 3.6, but that's still a significant number to just ignore them. That can be lots of questions on the forum, and lots of questions on pre-release. Proper documentation can avoid all that support time and effort.

I could remove the rows from the table for versions less than 3.6, but then how do those people find which upgrade instructions to use? We'd have to add more notes similar to "if you have less than 3.6, but more than 3.0 then figure out which upgrade notes apply to you..." etc which is more difficult to use than a simple lookup table: Just find your version and click the link.... no mental math and no guessing is required.

On the other hand, if it's decided to not support upgrading to the latest version from some older version, (1.0?, 2.0?, 3.0?) then that info can be removed from the docs, or at least put somewhere that can be linked to when a question arises on the forum or on pre-release.

And I agree that for the majority of admins, they just want to go straight to upgrading, but what they *want* to do and what they *should* do for a successful upgrade are not always the same thing. They need to be aware of breaking changes, and they need to know whether or not any additional actions are required either before or after the upgrade, and that varies depending on which version you're upgrading from.

jasonblais:
Sorry, I'm not saying we should ignore admins who are upgrading from 3.6 or earlier. What I'm saying is that I don't think upgrading from 3.0 or an older version should be the focus of this doc change.

One option to simplify the upgrade guide is to have
 - a single upgrade instruction on the upgrade page, e.g. the one you created in https://github.com/mattermost/docs/blob/889f044fa5ba863166d4dcfa2205d4db4f84c54e/source/administration/upgrading-to-latest.rst
 - for older versions, just have a link saying something like `If you're upgrading from v2.X, see {link}. From v1.X see {link}.` This would cover the case for 3% of the admins who are running such an old version of Mattermost.
 - for important notices, have them concisely in `Before you begin` section with links to more information if needed. Right now they take up a lot of space.
    - I'm 0/5 on the best solution, but one option would be listing the most recent/critical changes that affect most of the admins, and for older breaking changes, linking them to the changelog. E.g. the 3.9 breaking changes are listed here: https://docs.mattermost.com/administration/changelog.html#breaking-changes

it33:
@JeffSchering please take guidance from Lindsay. 

You and I have discussed and agreed that [current upgrade guide is not an A.](https://docs.mattermost.com/administration/upgrade.html) 

I have ask Lindsay to be accountable for getting the Upgrade Guide to an "A", and she needs your help as part of the team to do it. 

- To start, let's try having no more than one sentence in the main upgrade guide to direct people to a link where they can figure out what to do with older versions. 

- Regarding breaking changes, those are important and need to be CONCISELY communicated where they are relevant. 

@JeffSchering can you help Lindsay? 

It's been two weeks and I'm not seeing any commits based on her feedback.

crspeller:
Assigning to @lfbrock per off-boarding guide.

JeffSchering:
@lfbrock I made further updates. You can see how the guide looks when published here: https://jeffschering.github.io/mmdocs/upgrade/administration/upgrade.html

lfbrock:
Remaining questions: 
1. Upgrading with HA - I don’t see any sections on this, are there special instructions we need to add for upgrading an HA cluster?  
2. Should we move Version Archive to its own separate page? 
3. Important Upgrade Notes page:
    - I tried moving the upgrade notes to a table to see if it’s easier to read (screenshot below), if we do a table we’ll need to do some css changes to make the table text wrap as per: 
http://rackerlabs.github.io/docs-rackspace/tools/rtd-tables.html 
    - Looking at other examples: 
      - [MongoDB upgrade guide](https://docs.mongodb.com/manual/release-notes/3.2-upgrade/): Links to [compatibility notes](https://docs.mongodb.com/v3.2/release-notes/3.2-compatibility/)
      - [RedHat](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Migration_Planning_Guide/chap-Red_Hat_Enterprise_Linux-Migration_Planning_Guide-Upgrading.html) seems to have a tool that checks for issues/gives information about the upgrade
      - [Ubuntu](https://tutorials.ubuntu.com/tutorial/tutorial-upgrading-ubuntu-desktop#0) links to release notes for recent versions and upgrade notes for upgrading from older versions

lfbrock:
Made some edits so removing myself as a reviewer, we should get Jason + Corey to review now.

jasonblais:
1 -  There's an upgrade guide for HA here: https://docs.mattermost.com/deployment/cluster.html#upgrade-guide

That's a good point -- we should either find a way to add these to the docs or link to the HA doc.

2 - Yes, I think we should move this to a separate page

3 - Okay, sounds like linking to another page is common so we can do that. Once concern with tables is that if someone is upgrading from 3.8, is it clear that all notices from 3.8, 3.9, etc. apply to them, not just those listed for 3.8? Any thoughts?

jasonblais:
For 2 - separated to another page.

Let me know if that looks good and I can add it to TOC below upgrade guide & update the version archive with latest changes after 4.1

Also, removed "supported" and "unsupported" version numbers in the version archive, I don't think that's necessary.

jasonblais:
1 - Added a note for HA guide

2 - Separated to another page. Also, removed "supported" and "unsupported" categories in the version archive, I don't think that's necessary.

3 - I think the table is fine for now, I'll get AM's help look at the CSS if CH agrees with the change.

source/administration/upgrade.rst
```diff
@@ -1,241 +1,110 @@
...
+2. In a terminal window on the server that hosts Mattermost Server, change to your home directory.
+
+  ``cd ~``
+
+3. Delete any files and directories that might still exist from a previous download.
+
+  .. code-block:: text
+
+    rm mattermost*.gz
+    rm -r mattermost
```
coreyhulen: Looks like you're telling the user to delete the `data` dir before backing it up proplerly ?

coreyhulen: probably better to recommend a `mv` command to something-like `mattermost-back-YYYY-MM-DD` then when you all finished and verified the system is running correctly them can archive or remove the dir.

asaadmahmood:
@lfbrock Submitted a fix for the table wrapping stuff, that should be fixed now.

source/administration/upgrade.rst
jasonblais: @coreyhulen Yeah, I submitted an update to the upgrade guide last month but Jeff mentioned the correct steps are correct. Wondering if you have thoughts on that? Happy to make changes if you see fit.

PR: https://github.com/mattermost/docs/pull/1367

Jeff's note:

```
Step 3 deletes any previously downloaded files in the home directory. It doesn't delete any files in the running Mattermost instance. In step 6, the config.json is copied from the running Mattermost instance to the home directory.

The Mattermost server is stopped in Step 7 instead of Step 1 to help ensure that it is stopped for the shortest time possible during the upgrade.

If you follow the existing procedure on your own instance, copy/pasting from the guide to your terminal window, you'll see that it works perfectly as-is. I think this PR is trying to fix something that isn't broken.
```

source/theme.css
```diff
@@ -303,7 +303,7 @@ header .header__searchbar {
 		display: none;
 	}
 =======
-	
+
 >>>>>>> c8fd417... Added a search box to the nav header for the mobile view.
```
coreyhulen: This looks like a bad merge?

jasonblais: @coreyhulen Good catch, will submit a separate PR for it, looks like it's from an old PR

PR #1284, opened 2017-06-22
Update yum localinstall command

acgustafson:
Makes the yum localinstall command given work for all the distributions above.

### DIFF
--- a/source/install/install-rhel-71-postgresql.rst
+++ b/source/install/install-rhel-71-postgresql.rst
@@ -18,7 +18,7 @@ Installing PostgreSQL Database
 
 3. Install the Yum repository from the file that you downloaded.
 
-  ``sudo yum localinstall pgdg-*94-9.4-3.noarch.rpm``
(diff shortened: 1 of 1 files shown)


acgustafson:
On second thought this change would not work with Oracle Linux. That said I don't know why there is a link for Oracle on the RHEL page.

JeffSchering:
@acgustafson Would `pgdg-*-9.4-3.noarch.rpm` work for the Oracle case too?

The reason we have CentOS, Oracle Linux, and Scientific Linux here is because the installation is otherwise identical. See https://docs.mattermost.com/install/install-centos-oracle-scientific.html and the first paragraph of https://docs.mattermost.com/install/install-rhel-71.html

acgustafson:
@JeffSchering I was talking with @kjkeane and he mentioned he didn't think Oracle uses yum. According to the Oracle Wikipedia page it looks like it does use yum so `sudo yum localinstall pgdg-*94-9.4-3.noarch.rpm` should work. Sorry for the confusion I just haven't used Oracle Linux before.

PR #1300, opened 2017-06-29
Updating init.d script which does not work without SystemD

mioux:


mioux:
For informations, I had to change the --chuid $MATTERMOST_USER:$MATTERMOST_GROUP

It does not work with the user:group syntax.

There is also a useless starting space in the init script.

JeffSchering:
@mioux thanks!

>It does not work with the user:group syntax.

Can you verify that? According to the doc at http://man7.org/linux/man-pages/man8/start-stop-daemon.8.html, it should work. Actually, according to the description there for --chuid, the group part is superfluous. This is what it says: 

>When using this option you
              must realize that the primary and supplemental groups are set
              as well, even if the --group option is not specified. The
              --group option is only for groups that the user isn't normally
              a member of

mioux:
When I have debugged, I have checked by removing option by option to find which one was making the script crash.
When using user:group, the script does not run but returns 0 instead of non-zero as execution result.
I don't know if it's my dpkg version which makes this bug
```
root@mattermost:~# LANG=C sudo apt list --installed dpkg
Listing... Done
dpkg/oldstable,now 1.17.27 amd64 [installed]
N: There is 1 additional version. Please use the '-a' switch to see it
```

JeffSchering:
Ok, thanks! After you sign the CLA we can proceed with the reviews.

mioux:
OK it is signed

JeffSchering:
thanks @mioux!

Issue #1275, opened 2017-06-16
[Help Wanted] Documentation for installing on Debian Jessie


JeffSchering:
Help is needed to incorporate the feedback from https://github.com/mattermost/platform/issues/1185 into the Debian Jessie docs at https://docs.mattermost.com/install/prod-debian.html.

If you're willing to take this on, reply here and join the Documentation channel on the Mattermost public server at https://pre-release.mattermost.com/core/channels/documentation, where you can ask questions and coordinate efforts with anyone else who happens to be working on this too.

Thanks!

acgustafson:
I would love to take this on. Let me know if there is anything specific you are looking for. in the document.

JeffSchering:
Thanks @acgustafson! Basically the end goal is to have the Debian install look like the other installs (https://docs.mattermost.com/install/install-rhel-71.html and https://docs.mattermost.com/install/install-ubuntu-1604.html for eg.)

Ideally, someone should be able to, for the most part, copy/paste the commands from the instructions into their terminal and successfully install Mattermost.

JeffSchering:
Thanks for [Modified Install Guide for Debian Jessie #1301](https://github.com/mattermost/docs/pull/1301) @acgustafson!

PR #1301, opened 2017-06-30
Modified Install Guide for Debian Jessie

acgustafson:
This is in response to #1275 and [issue #1185](https://github.com/mattermost/platform/issues/1185).

Most of the pages are copied from the install-ubuntu-1604 guide. 

Some things I would like to bring attention to:

1. I included instructions to install and setup sudo and curl since Debian does not come with either out of the box. While this isn't directly related to installing Mattermost every other Mattermost install uses it and it is best practice to use sudo rather than running everything from the root account.

2. I provide a link to download MySQL from a different repo than the official Debian repos because the Debian repos only have MySQL 5.5.

3. I replaced the content on the old Debian page and replaced it with a link to the new guide similar to the prod-ubuntu-16.04.rst page.


JeffSchering:
Thanks @acgustafson! This is totally amazing! We'll start with the reviews right away.

kjkeane:
Just going to mention something I noticed

 * mysql install the ``mysql> exit`` has an extra space in front
 * also it should be step 10 instead of step 9

acgustafson:
Just updated with the two changes from @kjkeane

cpanato:
@acgustafson Thanks so much for this update :+1: 

@JeffSchering @jasonblais I tested the steps and all worked fine :)

source/install/install-debian-88-server.rst
```diff
@@ -0,0 +1,26 @@
...
+
+1. To install Debian Jessie, see the `Debian Installation Guide (PDF). <https://www.debian.org/releases/stable/amd64/install.pdf>`_
+
+2. After the system is installed, make sure that it's up to date with the most recent security patches. Open a terminal window and issue the following commands as root:
+
+  ``apt-get update``
+
+  ``apt-get upgrade``
+
+3. Debian does not come with sudo or curl which is used later in this install so let's install that with the following commands:
```
JeffSchering: This should be something like:
```
Install *sudo* and *curl* for use later in the installation.
```

source/install/install-debian-88-server.rst
```diff
@@ -0,0 +1,26 @@
...
+
+  ``apt-get install sudo curl``
+
+4. Let's add our non-root user to the sudo group so we don't have to be logged in as the root user.
```
JeffSchering: This should be something like:

```
Add your username to the sudo group. In the following command, replace {username} with your own username.

``usermod -a -G sudo {username}``

```

acgustafson:
@JeffSchering made those two changes!

jasonblais:
Ah sorry, there's just a couple of merge conflicts @acgustafson, can you help with those?

Once again, huge thanks for this contribution!!

JeffSchering:
@jasonblais Traditionally, there's been a separate set for each. There's some minor differences in a couple of places, but I don't see any problem combining the two sets into one, aside from the time and effort involved.

jasonblais:
@JeffSchering What are the differences? They seem identical to me apart from the title.

image

image

JeffSchering:
@jasonblais These are identical. I was talking about the existing EE install docs. I think mostly the differences are cosmetic, but some are not. For eg, the RHEL 7 EE doc has a section on using Unix domain sockets for communications between the database and Mattermost Server.

jasonblais:
Thanks @JeffSchering, let's go with what we have now.

I'll see if I can resolve the conflicts and we can merge this

jasonblais:
@JeffSchering Resolved conflicts + added to https://docs.mattermost.com/install/ee-install.html

Should be ready to merge.

Big thanks @acgustafson!!!

acgustafson:
@jasonblais No problem! Thanks for reviewing this!

PR #1303, opened 2017-07-01
Adding mindset on prioritization

it33:


it33:
Thanks @JeffSchering!!

it33:
@lfbrock merging as we'd discussed in leads, and I need to reference this, open to changing. "Needs PM Review" label remains.

PR #1306, opened 2017-07-03
Updates to developer toolkit

jasonblais:
PR against 4.0 doc branch, so the updates don't go out until the release

### DIFF
--- a/source/developer/toolkit.rst
+++ b/source/developer/toolkit.rst
@@ -34,7 +34,6 @@ Below is a list of planned features of the developer toolkit with estimated time
  - The ability to build webapp client plugins to override existing UI components (replace posts with your custom components, use your own video services etc.), modify/extend client drivers to interact with custom server API endpoints, and add whole new UI views in predetermined places.
 
 3. Unscheduled:
- - Mattermost Webapp UI components moved over to be Redux compatible.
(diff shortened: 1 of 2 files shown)


source/developer/toolkit.rst
```diff
@@ -25,18 +25,16 @@ Below is a list of planned features of the developer toolkit with estimated time
...
-3. Q3/2017:
+2. Q3/2017:
 
- - Mattermost Webapp UI components moved over to be Redux compatible.
+ - API developer token to provide a simple method to authenticate to the Mattermost REST API.
  - The ability to build webapp client plugins to override existing UI components (replace posts with your custom components, use your own video services etc.), modify/extend client drivers to interact with custom server API endpoints, and add whole new UI views in predetermined places.
 
-4. Unscheduled:
+3. Unscheduled:
+ - Mattermost Webapp UI components moved over to be Redux compatible.
```
jwilander: I think we can remove this. It doesn't really affect the developer toolkit anymore now that the client is using Redux on the backend. We'll still migrate individual components before letting client plugins override them but that's not a big deal. We could actually add `Move webapp to use Redux` into the completed section

jasonblais: Thanks, removed that item. There's already an item about moving the webapp to Redux in the completed section

PR #1314, opened 2017-07-06
Update ee-prod-rhel-7.rst

barrydegraaff:
start mattermost after mysqld in systemd

### DIFF
--- a/source/install/ee-prod-rhel-7.rst
+++ b/source/install/ee-prod-rhel-7.rst
@@ -190,7 +190,7 @@ Set up Mattermost Server
 
           [Unit]
           Description=Mattermost server
-          After=network.target postgresql-9.4.service
+          After=network.target postgresql-9.4.service mysqld.service
 
           [Service]
           User=mattermost


barrydegraaff:
@mattermod CLA signed

source/install/ee-prod-rhel-7.rst
```diff
@@ -190,7 +190,7 @@ Set up Mattermost Server
 
           [Unit]
           Description=Mattermost server
-          After=network.target postgresql-9.4.service
+          After=network.target postgresql-9.4.service mysqld.service
```
JeffSchering: The unit waits until after postgresql starts, but if the postgresql-9.4.service unit doesn't exist, will Mattermost still get started? Is systemd smart enough to overlook a missing unit?

If not, it might be better to indicate that if using mysql, then replace the postgresql-9.4.service with mysqld.service

barrydegraaff:
YES I can confirm that systemd ignores unit files that do not exist.

So when I follow the Mattermost install readme and opt to use mysql, upon copy-paste the original mattermost.service file, in most cases after a reboot Mattermost won't work as mysql is started after Mattermost. 

Appending mysqld.service in the After section solves this, even if no postgresql is installed. I did not test the other way around.

@JeffSchering the only othe solution would be adding separate mattermost.service unit files. One for mysqld and one for postgresql. 

What do you prefer? The current solution is easy on the eyes.

barrydegraaff:
https://github.com/mattermost/docs/pull/1316

JeffSchering:
Personally, I find it messy to list a unit that doesn't exist, even though it doesn't break anything. It's cleaner and more "pure" to have just the one that's needed.

Maybe retain just the one file, listing both postgresql and mysql, but add a line in the step that says something like "delete the one you don't need". If people miss or ignore that line, it's no big deal; everything still works.

barrydegraaff:
Yea, I closed this PR and made a new one, that does not list non-existing units.

PR #1315, opened 2017-07-06
Updated message table image

amyblais:


jasonblais:
Thanks @amyblais! I noticed the file extension doesn't match, can you help
 - rename the file to `messagesTable1`
 - take a screenshot with Snipping Tool (that'll add the .PNG extension, instead of .png)

PR #1333, opened 2017-07-13
Update for v4.0 release

JeffSchering:
Preliminary update. Need to add checksums when they become available.


jasonblais:
^while the changes were made in 3.8, it won't actually be a breaking change until 4.0. 

We wanted to give admins an early heads up, so that hopefully many have already made the NGINX change.

jasonblais:
@JeffSchering Once 4.0.0 is cut and we add the checksums, we should also update the version archive for 3.10.X, 3.9.X that we'll be cutting likely later on Friday, and for 3.8.X and 3.7.X that we cut a week ago

jasonblais:
We can merge this and other 4.0 docs to master, once everything is reviewed.

I can take care of updating the version archive stuff, which might not be until later tonight.

PR #1340, opened 2017-07-15
Remove Beta from Slack import

it33:
Propose removing "Beta" from Slack import since, 1) we already have warning that Slack's schema can change without notice, and 2) we already tell people to try on a backup before attempting full import.


it33:
@grundleborg wondering your thoughts on this?

grundleborg:
We know it doesn't work as smoothly as users might expect (e.g. it not being idempotent frequently catches people out), and it frequently has bugs introduced by (1). If you think that meets the definition of "production quality", then go ahead. Otherwise, better keep the beta label.

JeffSchering:
@it33 any decision on what to do with this yet? Do you want input from PMs?

it33:
Will queue for Platform Meeting

Issue #1161, opened 2017-05-05
Replace embedded images in docs with RST tables where possible


jasonblais:
There are examples of embedded images in our docs, which could simply be RST tables. The embedded images makes it hard to
 - update information when something has changed
 - copy the text

Some examples include:
 - embedded image for Markdown Basics (if a user sees the help table, there's a good chance they try to copy-paste the Markdown formatting into a channel): https://docs.mattermost.com/help/getting-started/messaging-basics.html#messaging-basics
 - embedded image for list of commands (makes it hard to update if anything is changed): https://docs.mattermost.com/help/messaging/executing-commands.html

Note: Some of these help docs are in our platform repo too as they're translated across different languages: https://github.com/mattermost/platform

JeffSchering:
Closing. Resolved by https://github.com/mattermost/docs/pull/1356

PR #1356, opened 2017-07-19
Replace images of tables with table markup

JeffSchering:
This is for [Replace embedded images in docs with RST tables where possible #1161](https://github.com/mattermost/docs/issues/1161)

Only two tables were replaced. The others had formatting that is not possible to achieve with reStructuredText markup.

This is for the tables only. I didn't review the text for compliance with Style Guide or other issues. My understanding is that @amyblais will do that at some point.


crspeller:
Assigning to @lfbrock per off-boarding guide.

PR #1358, opened 2017-07-20
Create dev-setup-centos-7.rst

prixone:
Installing Developer Components on CentOS 7

### DIFF
--- a/source/developer/dev-setup-centos-7.rst
+++ b/source/developer/dev-setup-centos-7.rst
@@ -19,17 +19,17 @@ Set up your development environment for building, running, and testing Mattermos
 
   d. Restart the Docker daemon.
 
-    ``sudo service docker restart``
+    ``sudo systemctl restart docker``
 
   e. Change your current group ID to the *docker* group.
 
     ``newgrp docker``
 
-2. Install the development tools package and libpng12 required by pngquant(a library used by Mattermost).
+2. Install the development tools package, wget and libpng12 required by pngquant(a library used by Mattermost).
 
     a. ``sudo yum group install "Development Tools"``
 
-    b. ``sudo yum install -y libpng12``
+    b. ``sudo yum install -y wget libpng12``
 
 3. Download and install Go 1.8 for Linux:
 
--- a/source/developer/dev-setup-centos-7.rst
+++ b/source/developer/dev-setup-centos-7.rst
@@ -29,7 +29,7 @@ Set up your development environment for building, running, and testing Mattermos
 
     a. ``sudo yum group install "Development Tools"``
 
-    b. ``yum install -y libpng12``
+    b. ``sudo yum install -y libpng12``
 
 3. Download and install Go 1.8 for Linux:
 
--- a/source/developer/dev-setup-centos-7.rst
+++ b/source/developer/dev-setup-centos-7.rst
@@ -1,0 +1,98 @@
+.. _dev-setup-centos-7:
+
+Installing Developer Components on CentOS 7
+===========================================
+
+Set up your development environment for building, running, and testing Mattermost.
+
+1. If you are developing with the Docker container, install and configure Docker CE:
+
+  a. Follow the Docker installation instructions at https://docs.docker.com/engine/installation/linux/docker-ce/centos/.
+
+  b. Edit your ``/etc/hosts`` file to include the following line:
+
+    ``127.0.0.1     dockerhost``
+
+  c. Add your username to the *docker* group so that you can run Docker without using sudo. Replace *{username}* with your username.
+
+    ``sudo gpasswd -a {username} docker``
+
+  d. Restart the Docker daemon.
+
+    ``sudo service docker restart``
+
+  e. Change your current group ID to the *docker* group.
+
+    ``newgrp docker``
+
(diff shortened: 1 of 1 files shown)


source/developer/dev-setup-centos-7.rst
```diff
@@ -0,0 +1,98 @@
...
+
+    ``127.0.0.1     dockerhost``
+
+  c. Add your username to the *docker* group so that you can run Docker without using sudo. Replace *{username}* with your username.
+
+    ``sudo gpasswd -a {username} docker``
+
+  d. Restart the Docker daemon.
+
+    ``sudo service docker restart``
```
JeffSchering: I believe on CentOS 7 this should be ` sudo systemctl start docker`

source/developer/dev-setup-centos-7.rst
```diff
@@ -0,0 +1,98 @@
...
+
+    a. ``sudo yum group install "Development Tools"``
+
+    b. ``sudo yum install -y libpng12``
+
+3. Download and install Go 1.8 for Linux:
+
+    a. Download the Go binary.
+
+       ``wget https://storage.googleapis.com/golang/go1.8.linux-amd64.tar.gz``
```
JeffSchering: wget is not available on CentOS 7 out of the box. Need to either install wget, or use curl.
```
curl -O https://storage.googleapis.com/golang/go1.8.linux-amd64.tar.gz
```

The yarn install instructions use wget, so probably best to install wget. On the other hand, the audience for this doc is developers, so if wget isn't on their system, they know what to do. It's up to you if you want to add a separate line for installing wget.

source/developer/dev-setup-centos-7.rst
```diff
@@ -0,0 +1,98 @@
...
+
+  d. Restart the Docker daemon.
+
+    ``sudo service docker restart``
+
+  e. Change your current group ID to the *docker* group.
+
+    ``newgrp docker``
+
+2. Install the development tools package and libpng12 required by pngquant(a library used by Mattermost).
```
JeffSchering: Is CentOS 7 the only OS that needs libpng installed separately? Should installing libpng12 be added to the other instruction sets? (Ubuntu, Mac, etc)

source/developer/dev-setup-centos-7.rst
```diff
@@ -0,0 +1,98 @@
...
+
+      export GOPATH=$HOME/go
+      export PATH=$PATH:$GOPATH/bin
+      export PATH=$PATH:/usr/local/go/bin
+
+  c. Reload your bash configuration.
+
+      source ~/.bashrc
+
+  d. Since you cannot run ``ulimit -n 8096`` with a regular user, we have to edit the file ``sudo nano /etc/security/limits.conf`` to include the following 2 lines at the end of it:
```
JeffSchering: Are you sure this is needed? Why the hard limit? If only changing the soft limit (as in the other setups, Ubuntu 16.04 for example), you don't need to be root. You only need to be root if you want to increase the hard limit.

source/developer/dev-setup-centos-7.rst
prixone: @JeffSchering both ways work, the system will automatically convert from system to systermctl

source/developer/dev-setup-centos-7.rst
prixone: I don't think other OS's need that, at least my installation on Ubuntu did not require me to install it separately but the CentOS one did.

source/developer/dev-setup-centos-7.rst
prixone: > On the other hand, the audience for this doc is developers, so if wget isn't on their system, they know what to do.

That was my initial though but I guess I should add wget along with libpng as I did, because even though the target is developers its not only experienced developers that use the dev version, some new comers and even less experienced people also use it to get start with something, and ended up being a good help to Mattermost so having it more complete is better.

Thanks for pointing it out, doing the change now.

source/developer/dev-setup-centos-7.rst
prixone: I don't know if that happens on Ubuntu, I can try later and let u know, but on CentOS 7 you cannot set that feature as a user on the default installation without doing it as a root, or change the settings like that. And the ulimit AFAIK is important to mattermost even though I am not sure it is needed for a dev setup unless they are also using it for load tests among other things that might trigger the nofile limit.

prixone: @JeffSchering also a think I would like to point out on the Ubuntu is that YOU should NOT need to chown the go folder or binaries, the ZIP/Tarball it comes with should already set the necessary chmod to allow users to run it while retaining everything else as root. You don't normally go on OS's setting binaries to the user group permission but instead you set the chmod to allow them to use it. I assume ubuntu did not set the right permissions and u had to do that? Instead it might be better to just set it to 755

JeffSchering: Can you create a PR to change the chown to a chmod for Ubuntu?

prixone: Yep https://github.com/mattermost/docs/pull/1363

prixone:
I have already done all the changes to this file since I replied to it yesterday is there something still missing here I overlooked?

PR #1363, opened 2017-07-21
Update dev-setup-ubuntu-1604.rst

prixone:
Instead of changing the owner of the golang binaries, simple set the right permission.

### DIFF
--- a/source/developer/dev-setup-ubuntu-1604.rst
+++ b/source/developer/dev-setup-ubuntu-1604.rst
@@ -39,14 +39,6 @@ Set up your development environment for building, running, and testing Mattermos
 
        ``sudo tar -C /usr/local -xzf go1.8.linux-amd64.tar.gz``
 
-    c. Ensure the binary file of golang have the right permission, by running:
-
-      ``sudo ls -la /usr/local/go/bin/go``
-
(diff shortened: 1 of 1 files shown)


source/developer/dev-setup-ubuntu-1604.rst
```diff
@@ -39,9 +39,13 @@ Set up your development environment for building, running, and testing Mattermos
 
        ``sudo tar -C /usr/local -xzf go1.8.linux-amd64.tar.gz``
 
-    c. Modify permissions on ``/usr/local/go``. Replace *{user}* and *{group}* with the user and group that you are logged in with.
+    c. Ensure the binary file of golang have the right permission, by running:
```
JeffSchering: Looks like the permissions are already 755 after extracting go from the tarball.

On my local Ubuntu install, I changed the owner:group of those files to root:root and the build works fine.

So maybe Step c is completely unnecessary and can be eliminated rather than changing it and adding Step d.

prixone: If you tried that and that was the result, then yes, it could be completely removed. I have not tested it myself but normally when you extract a tarball it keeps the permissions, I can't speak for all *nix, but at the very least that's how it is for CentOS 5~7 and apparently for Ubuntu. So I will remove c and d.

PR #1364, opened 2017-07-21
Add limitation about Slack import not being idempotent

JeffSchering:


source/administration/migrating.rst
```diff
@@ -73,6 +73,8 @@ Limitations
 
 The following limitations are present when importing from Slack:
 
+* The import is not idempotent, which means that duplicate entries are created if you import the same data more than once.
```
grundleborg: I would say "duplicate posts", as they are the only thing which is duplicated.

JeffSchering: Done.

PR #1367, opened 2017-07-22
Reorder steps in upgrade.rst

jasonblais:
Change upgrade guide based on feedback from @saturninoabril 

> step #5 should be moved atleast before step #2 since #2 deletes the folder of `mattermost` and will not be able to copy the `config.json` afterwards.
> (especially for 7b if they are storing data at `mattermost/data` as we've documented in installation)

### DIFF
--- a/source/administration/upgrade.rst
+++ b/source/administration/upgrade.rst
@@ -45,41 +45,41 @@ Important Notices
 
 **To upgrade Mattermost Server**:
 
-1. In a terminal window on the server that hosts Mattermost Server, change to your home directory.
+1. Stop Mattermost Server.
+
+  On Ubuntu 14.04 and RHEL 6.6: ``sudo service mattermost stop``
+
+  On Ubuntu 16.04 and RHEL 7.1: ``sudo systemctl stop mattermost``
+
+2. Make a copy of your configuration file. The existing file is overwritten during the upgrade, so it's important that you don't forget this step
+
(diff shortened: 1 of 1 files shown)


JeffSchering:
Step 2 deletes any previously downloaded files in the home directory. *It doesn't delete any files in the running Mattermost instance.* In step 5, the `config.json` is copied  from the running Mattermost instance to the home directory.

The Mattermost server is stopped in Step 6 instead of Step 1 to help ensure that it is stopped for the shortest time possible during the upgrade.

If you follow the existing procedure on your own instance, copy/pasting from the guide to your terminal window, you'll see that it works perfectly as-is. I think this PR is trying to fix something that isn't broken.

jasonblais:
@saturninoabril Any thoughts on Jeff's note above?

saturninoabril:
@JeffSchering I stand corrected on the wrong assumption I've made. Honestly, I was previously confused about the directories `mattermost` and `{install-path}`, and initially thought that those were the same. Thanks for the clarification.
@jasonblais Sorry for the wrong input. The existing upgrade steps are accurate. Kindly close the PR. thanks!

jasonblais:
Thanks @saturninoabril! No worries

I'm wondering if there's an opportunity to clarify `{install-path}` and that it's not the same as `mattermost`? @JeffSchering 

I'll close this PR in the meantime.

PR #1372, opened 2017-07-26
Proposing process to post deprecation plans

it33:
Proposal to prevent future "surprises" for the community over changes to features such as thumbnail controls: https://about.mattermost.com/mattermost-4-0/#comment-3436068337

### DIFF
--- a/source/process/release-process.md
+++ b/source/process/release-process.md
@@ -82,7 +82,7 @@ Day when leads and PM area owners decide which major features are included in th
     - Confirm, in Judgement Day meeting, date of marketing announcement for the release and update release channel header if needed
 3. Leads:
     - Finalize roadmap for next release, and identify planned marketing bullet points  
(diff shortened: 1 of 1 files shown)


lfbrock:
we already have a process for this so you need to update it in both places and make sure it's consistent: https://github.com/mattermost/docs/blob/master/source/process/deprecated-features.rst

I think we should add a guideline about how often in advance we should try to announce the deprecation as well.

hmhealey:
In cases like those image sizes, we could also include a brief point in the marketing about why the settings are being removed. I think that guy thinks we're removing the ability to post images like Slack does which wasn't how it worked in the first place.

jasonblais:
We also had forum posts on deprecations (except the thumbnail one).

Maybe we should link to those in the release announce for community feedback whenever we mention upcoming and current deprecations.

it33:
Closing and moving discussion to Platform Meeting

PR #1389, opened 2017-08-01
Changing channel to channel_name

it33:
to make commands more clear


jasonblais:
I don't think the CLI command changed. I added a note to clarify how to specify `{channel}` for CLI commands.

PR #1407, opened 2017-08-10
Updated executing-commands.rst

amyblais:
Made changes to the slash commands table


jasonblais:
Thanks! Similar to #1408, I forgot to mention that this will need to be for the 4.1-documentation branch. My mistake.

No need to re-submit, we can just merge it next week on the 15th.

PR #1408, opened 2017-08-10
Updated slash-commands.rst

amyblais:
Made the same changes to the slash commands table as in my previous PR.


jasonblais:
Thanks! I forgot to mention that this will need to be for the 4.1-documentation branch. My mistake.

No need to re-submit, we can just merge it next week on the 15th.

PR #1412, opened 2017-08-12
Adding MySql User creation error solution

pruthvip:
Explains the error, when the mysql user is not created. And shows the steps on how to create a user and give all the permissions


pruthvip:
Have submitted the form.

PR #1414, opened 2017-08-14
Slash command doc updates

jasonblais:


source/developer/slash-commands.rst
```diff
@@ -3,111 +3,72 @@
...
+------------------------
 
-Each Mattermost installation comes with some built-in slash commands that are ready to use. These commands are listed below:
+Each Mattermost installation comes with some built-in slash commands that are ready to use. These commands are available in the `latest Mattermost release <https://about.mattermost.com/download/>`_:
 
 .. csv-table::
     :header: "Command", "Description", "Example"
 
-    "/away", "Set your status to away", "/away"
+    "/away", "Set your status away", "/away"
```
hmhealey: 0/5 on this, but if we're grouping these, `/code`, `/echo`, `/me`, and `/shrug` could probably go together since they all print messages. We may also want to consider breaking the table up into sections so we can still keep them alphabetical, but it's not big enough that that would be a problem yet

source/developer/slash-commands.rst
```diff
@@ -3,111 +3,72 @@
...
-Enabling Custom Commands
-------------------------
+Custom Slash Command
+------------------------------
+
+Suppose you want to write an external application that is able to check the weather for certain cities. By creating a custom slash command, and setting up the application to handle the HTTP POST or GET from the command, you can let your users check the weather in their city using your command, say ``/weather toronto week``.
+
+You can follow these general guidelines to set up a custom Mattermost slash command for your application.
+
+1 - First, go to **Main Menu > Integrations > Slash Commands**. If you don't have the **Integrations** option in your Main Menu, slash commands may not be enabled on your Mattermost server. Enable them from **System Console > Integrations > Custom Integrations** or ask your System Administrator to do so.
```
hmhealey: We might want a comment about them being disabled for non-admins?

source/developer/slash-commands.rst
```diff
@@ -128,54 +89,184 @@ If you'd like to build your own integration that uses slash commands, you can fo
...
+  .. code-block::
+
+    payload={"response_type": "ephemeral", "goto_location": "https://about.mattermost.com", "text": "Hello, this is some text\nThis is more text. :tada:"}
+
+The parameter supports any custom protocol including ``http://``, ``https://``, ``ftp://``, ``ssh://`` and ``mailto://``.
+
+Markdown formatting
+~~~~~~~~~~~~~~~~~~~~
+
+A rich range of formatting unavailable in Slack is made possible through :doc:`markdown support <../help/messaging/formatting-text>` in Mattermost, including headings, formatted fonts, tables, inline images and other options supported by Mattermost Markdown. All of these options are also supported by slash commands.
```
hmhealey: I wouldn't bother mentioning Slack here, but 0/5 on that.

source/developer/slash-commands.rst
```diff
@@ -128,54 +89,184 @@ If you'd like to build your own integration that uses slash commands, you can fo
...
+
+    payload={"icon_url": "http://example.com/somecoolimage.jpg", "text": "Hello, this is some text\nThis is more text. :tada:"}
+
+  .. note::
+    `Enable integrations to override profile picture icons <https://docs.mattermost.com/administration/config-settings.html#enable-integrations-to-override-profile-picture-icons>`_ must be set to `true` in `config.json` to override usernames. Enable them from **System Console > Integrations > Custom Integrations** or ask your System Administrator. If not enabled, the icon of the creator of the webhook URL is used to post messages.
+
+Tips and Best Practices
+------------------------
+
+1. Slash commands are designed to post messages. For other actions such as channel creation, use the `Mattermost APIs <../developer/api.html>`_.
```
hmhealey: Maybe change the first sentence to "Slash commands are designed to easily allow you to post messages. For other actions such as channel creation, you must also use the Mattermost APIs"? Right now, it sounds a bit like you can't have them do other things

source/developer/slash-commands.rst
jasonblais: Yeah, it might make sense to start breaking the commands into sections, similar to keyboard shortcuts.

Will consider that next time I update this list.

jasonblais:
Thanks for the review! Have made updates based on feedback.

PR #1436, opened 2017-08-17
Add personal access token role to managing-members.rst

jasonblais:
I wasn't quite sure if this fits here, so let me know. We might want to separate the "roles and permissions" into a separate doc later.


esethna:
Ya I think it's fine there for now, looks good

PR #1440, opened 2017-08-18
Create pm-faq.rst

lfbrock:
Work in progress - not quite sure where this doc should go so started a new one, open to suggestions though. 

We should also decide what number of users we want to put for the "scalable" item

### DIFF
--- a/source/process/pm-faq.rst
+++ b/source/process/pm-faq.rst
@@ -1,0 +1,14 @@
+Product Manager FAQ
+===================
+
+Can we take this feature out of beta?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To officially release a feature, it should meet the following criteria: 
+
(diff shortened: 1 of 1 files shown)


amyblais:
There is this existing WIP draft for PM documentation, I think the best option would be to put the information from this PR there: https://github.com/mattermost/docs/blob/master/source/process/product-manager.rst

lfbrock:
@amyblais - sorry for the delayed response; I think it might be good to keep them in separate places, I think the product-manager doc is more for hiring purposes, so we may want to put other PM handbook stuff elsewhere. 

I think in the future we'd like to move our core team handbook to its own website at some point, so maybe it's fine to just leave this as it's own doc for now and then re-organize when we migrate the docs.

amyblais:
@lfbrock Sorry for the long delay on this one, I added this to a table of contents now.

PR #1446, opened 2017-08-21
Add common licensing questions to FAQ page

jasonblais:
I'm not sure if this is the best approach, but a customer searched our docs and couldn't find answers to these questions which were on our pricing page.

So added the questions to our FAQ and linked to the pricing page for the answers: https://about.mattermost.com/pricing/#frequently-asked-questions-about-licensing

### DIFF
--- a/source/overview/faq.rst
+++ b/source/overview/faq.rst
@@ -1,6 +1,9 @@
 Frequently Asked Questions (FAQ)
 =================================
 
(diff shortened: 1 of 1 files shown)


lfbrock:
yeah not sure what other options there could be, I think this approach is fine for now. Better to have the questions there than not have them at all, +1

PR #1452, opened 2017-08-23
Fix default values of SiteURL and SignInWithUsername in docs

balasankarc:
Closes https://github.com/mattermost/platform/issues/7277


jasonblais:
Thanks @balasankarc!

I reverted the change to Site URL. I discussed this with a developer and https://github.com/mattermost/platform/blob/master/config/default.json actually doesn't contain the default values in release builds. A few of the settings are changed in release builds [here](https://github.com/mattermost/platform/blob/master/Makefile#L396). The first part surrounded by pipe characters in the sed command is the line in the config file and the second part is what it's replaced with.

The reason why some fields like Site URL is set to ``http://localhost:8065`` is for easier development.

Otherwise looks good though, appreciate it :)

PR #1456, opened 2017-08-28
Update telemetry.rst

jasonblais:


source/administration/telemetry.rst
```diff
@@ -69,7 +69,7 @@ Non-personally Identifiable Diagnostic Information
...
   - *Core Feature Discovery Diagnostics:* Created, edited or deleted a message; posted a message containing a hashtag, link, mention or file attachment; searched for a term; searched for flagged posts or recent mentions
-  - *Advanced Feature Discovery Diagnostics:* Reacted to a message; favorited or un-favorited a channel; flagged or un-flagged a message; pinned or un-pinned a message; replied to a message; expanded the right-hand sidebar; started or finished a WebRTC video call
+  - *Advanced Feature Discovery Diagnostics:* Reacted to a message; favorited or un-favorited a channel; flagged or un-flagged a message; pinned or un-pinned a message; replied to a message; expanded the right-hand sidebar; started or finished a WebRTC video call; added or removed a user role; created or deleted a personal access token
```
esethna: Not totally clear on what "added or removed a user role" means? Is that referring to promotion or demotion from channel admin, team admin, system admin?

jasonblais: @esethna Good point..would `added or removed post:all or post:channels permission` be better? (it's specifically for those two)

esethna: Yep, updated the PR, thanks

PR #1473, opened 2017-09-06
Appconfig docs

lfbrock:
note: 
- check relative links, seemed like they weren't working quite right on local copy of docs 
- probably needs to be added to table of contents somewhere?


lindy65:
@esethna, will check the links after merge.

PR #1478, opened 2017-09-08
Update 4.2 Config Settings

esethna:


esethna:
Thanks Jason. I've just added an "(Experimental)" tag to the settings similar to how we did "(Beta)" tags. I think that's fine for now, I worry it will get confusing and hard to follow if we split the doc into too many sections.

PR #1481, opened 2017-09-12
Create interactive-message-buttons.rst

jasonblais:
@ccbrown I mentioned you in a few places in the doc where it would be good to get your review/help with.

Also, if there's any additional info you think would be useful for a developer creating an integration with buttons, let me know.


source/developer/interactive-message-buttons.rst
```diff
@@ -0,0 +1,116 @@
+.. _interactive-message-buttons:
+
+Interactive Message Buttons (Beta)
+===================================
+
+Mattermost supports interactive message buttons for `incoming <https://docs.mattermost.com/developer/webhooks-incoming.html>`_ and `outgoing webhooks <https://docs.mattermost.com/developer/webhooks-outgoing.html>`_, and for `custom slash commands <https://docs.mattermost.com/developer/slash-commands.html>`_ via actions. They help make your integrations richer by completing common tasks inside Mattermost covnersations, increasing user engagement and productivity.
```
ccbrown: "covnersations"

source/developer/interactive-message-buttons.rst
```diff
@@ -0,0 +1,116 @@
...
+
+// XXX Insert an image here
+
+Below we give a brief description of each parameter to help you customize the webhook post in Mattermost. For more information on message attachments, `see our documentation <https://docs.mattermost.com/developer/message-attachments.html>`_.
+
+Name
+  Give your action a descriptive name.
+
+URL
+  The actions are backed by an integration that handles HTTP POST requests when users clicks the message button. The URL parameter determines where this action is sent to. The request contains an ``application/x-www-form-urlencoded`` JSON string. // XXX @ccbrown is this correct?
```
ccbrown: Not `application/x-www-form-urlencoded`. Just `application/json`.

source/developer/interactive-message-buttons.rst
```diff
@@ -0,0 +1,116 @@
...
+
+Context // XXX @ccbrown Need a bit of help to more specifically explain this parameter.
```
ccbrown: There are two things that virtually every integration will need to use this for:

1. Identifying which action was triggered: For example, a GitHub integration might store something like this in the context:

    ```json
    {
        "repo": "mattermost/mattermost-server",
        "pr": 1234,
        "action": "merge"
    }
    ```

    Then when the button is clicked and it receives that context back, the integration knows that the intention was to merge the pull request identified by the context.

2. Authenticating the server: If your integration isn't behind a firewall, anyone can forge requests that look like they came from the Mattermost server. One very important property of the context is that the entire context is kept confidential, so it's safe to put secrets in it if needed. Users won't be able to see them. So if your integration isn't behind a firewall, you might add a token to your context:

    ```json
    {
        "repo": "mattermost/mattermost-server",
        "pr": 1234,
        "action": "merge",
        "token": "somerandomlygeneratedsecret"
    }
    ```

    Then, when you receive the request, you can verify that the token matches one that you previously generated and know that the request is legitimately coming from the Mattermost server.

    Depending on the application, integrations may be able to perform authentication statelessly with cryptographic signatures:

    ```json
    {
        "repo": "mattermost/mattermost-server",
        "pr": 1234,
        "action": "merge",
        "signature": "mycryptographicsignature"
    }
    ```

It's also possible for integrations to do both of these things with a single token and simple store something like this as context:

```json
{
    "action_id": "someunguessableactionid"
}
```

Then when the integration gets the request, it can look up what to do based on the action id. But this requires integrations to maintain a bit more state on their end.

source/developer/interactive-message-buttons.rst
```diff
@@ -0,0 +1,116 @@
...
+-----------------------
+
+If you've built an integration for Mattermost, please consider `sharing your work <https://www.mattermost.org/share-your-mattermost-projects/>`_ in our `app directory <https://about.mattermost.com/default-app-directory/>`_.
+
+The `app directory <https://about.mattermost.com/default-app-directory/>`_ lists open source integrations developed by the Mattermost community and are available for download, customization and deployment to your private cloud or on-prem infrastructure.
+
+Slack Compatibility
+--------------------
+
+// XXX @ccbrown, please help with how one can turn a Slack-type message button into a Mattermost one? It was just a few tweaks if I remember correctly?
```
ccbrown: Hmm. They might be a little too distant from each other for a concise guide... The format for the message attachment JSON is somewhat close, but the integration request and response are totally different.

source/developer/interactive-message-buttons.rst
```diff
@@ -0,0 +1,116 @@
...
+    "user_id": "rd49ehbqyjytddasoownkuqrxe",
+    "context": {
+      "action": "ephemeral"
+      }
+    }
+
+Tips and Best Practices
+------------------------
+
+// XXX @ccbrown Do we want to clean up the polling integration demo and provide it as an example? https://github.com/ccbrown/interactive-post-demo
```
ccbrown: Sounds like a good idea to me. If we do that, I'll move it to the mattermost GitHub organization though.

source/developer/interactive-message-buttons.rst
```diff
@@ -0,0 +1,170 @@
...
+
+Context
+  The requests sent to the specified URL contain the user id and any context that was provided in the action definition. A simple example is given below:
+  
+  .. code-block:: text
+
+    {
+    "user_id": "rd49ehbqyjytddasoownkuqrxe",
+    "context": {
+      "action": "ephemeral"
```
ccbrown: I would suggest changing the example to something else like `"action": "do_something"` just to avoid confusion and make it obvious that the value is user-defined and has no special meaning.

source/developer/interactive-message-buttons.rst
ccbrown: Maybe something like...

> Interactive messages are not intended to be Slack compatible.
> 
> Slack requires a Slack App and action URL to be pre-configured beforehand. Mattermost allows any webhook or slash command to create an interactive message without any additional configuration.
> 
> Like Slack, actions are specified in an "actions" list within the Slack attachment. However, the schema for these objects is completely different.
> 
> With Slack, the request made to your integration when a user performs an action contains information such as channel and team ids. With Mattermost, it contains only a user id and whatever information you specified in your context. You can put the team id or channel id in your context when you post the message if you'll need these things again later.
> 
> Like Slack, your integrations can react with ephemeral messages or message updates. But again, the schema for this is slightly different.

:-/

source/developer/interactive-message-buttons.rst
jasonblais: Good point, will do

jasonblais:
@ccbrown Updated, let me know how that looks?

source/developer/interactive-message-buttons.rst
```diff
@@ -0,0 +1,174 @@
...
+
+- update sales opportunities in your CRM system from Mattermost
+- conduct a customer survey or a poll
+- file a report on market trends
+
+To try the message buttons out, you can use this `demo integration <https://github.com/mattermost/mattermost-interactive-post-demo>`_ to add polling to Mattermost channels via a `/poll` slash command.
+
+Interactive message buttons are not yet supported on Mattermost mobile apps.
+
+// XXX Add GIF image (AB preparing one for marketing)
```
esethna: Still needs GIF

Issue #1513, opened 2017-09-19
Provide a troubleshooting section specific to MySQL installation errors


peterdamian:
In a bare-bones installation (see #1504) the only possible point of failure apart from the Mattermost server, is the database. MySQL database error messages are not always helpful. Give examples of error messages and show how they relate to following types of errors:

- incorrect settings in the database driver file
- incorrect owner of mattermost database
- incorrect database user

This could be a new "Installation Troubleshooting" section.

lindy65:
Hi @peterdamian 

Thanks for your feedback!  Appreciated...

Is this something you'd be interested in doing a pull request for?

peterdamian:
Hi @lindy65,

I'll start working on this.

lindy65:
Thanks @peterdamian :)

justinegeffen:
Hi @peterdamian. I wanted to follow up on this and find out whether you'd be open to submitting a pull request for your changes? They look very useful and it would be great to have the team look them over. Thank you!

justinegeffen:
This work has been completed and can be found here: https://docs.mattermost.com/install/trouble_mysql.html. Closing. :)

Issue #1507, opened 2017-09-18
Change all instances of "Mac OS X" to "macOS"


peterdamian:
On https://docs.mattermost.com/developer/dev-setup-osx.html the name of the Mac operating system is given as "Mac OS X". Apple now calls their OS "macOS". Change all instances of "Mac OS X" to "macOS".

The following files contain this phrase (result of ``grep -Ri "Mac OS X" * ``):

````
source/developer/dev-setup-compiling.rst:  - :doc:`Mac OS X <dev-setup-osx>`
source/developer/dev-setup-osx.rst:Installing Developer Components on Mac OS X
source/developer/dev-setup-troubleshooting.rst:Mac OS X: I get the following error when running ``make run``: "Cannot connect to the Docker daemon"
source/developer/dev-setup-troubleshooting.rst:Mac OS X: I get the following error or something similar when running ``make run``: "Module build failed: Error: dyld: Library not loaded: /usr/local/opt/libpng/lib/libpng16.16.dylib"
source/developer/dev-setup-troubleshooting.rst:Mac OS X: I get the following error or something similar when running ``make run``: [EROR] Failed to ping DB retrying in 10 seconds err=Error 1045: Access denied for user ‘mmuser’@’localhost’ (using password: YES) 
source/developer/dev-setup.rst:    Setting up Mac OS X <dev-setup-osx.rst>
source/developer/mobile-developer-setup.rst:Mac OS X
source/developer/mobile-developer-setup.rst:  b. Unzip the SDK to somewhere on your hard drive. For example, ``/Users/<username>/Library/Android/sdk`` on Mac OS X.
source/install/desktop.rst:Mac OS X 10.9+
source/install/docker-local-machine.rst:Mac OS X
source/install/prod-docker.rst:Production Docker Setup on Mac OS X
source/install/prod-docker.rst:You can run a production deployment on Mac OS X by `installing Docker Compose using the online guide <http://docs.docker.com/installation/mac/>`_ then following the above instructions.
````

PR #1525, opened 2017-09-25
WIP: Changes for issues #1513 and #1507

peterdamian:
Add section on MySQL installation troubleshooting and change 'Mac OS X' to 'macOS".


source/install/trouble_mysql.rst
```diff
@@ -0,0 +1,250 @@
...
+
+    Uptime: 877134  Threads: 1  Questions: 9902  Slow queries: 0  Opens: 522  
+    Flush tables: 1  Open tables: 371  Queries per second avg: 0.011
+
+If MySQL is not running, review the instructions for installation on
+your distribution.
+
+.. warning:: Some of the commands used in this section alter the database. **Use these commands only if your Mattermost installation has failed. Do not directly manipulate the MySQL database for a working Mattermost installation**.
+
+The mattermost Database
```
hmhealey: I know this is the name of the database that we use internally, but I think it looks kind of weird to have it lower case in headings since it's easy to confuse with the product name itself. I think it would be fine to have it capitalized here and refer to it in lower case elsewhere.

It might actually make sense to separate the heading from the database name in case some users have decided to give their database a different name

source/install/trouble_mysql.rst
```diff
@@ -0,0 +1,250 @@
...
+editor, and review the value of ``"DataSource"``. It should be: ::
+
+    "mmuser:<mmuser-password>@tcp(<host-name-or-IP>:3306)/mattermost?charset-utf8mb4,utf8&readTimeout-30s&writeTimeout-30s"
+
+Replace `mmuser-password` with the password created during installation
+and `host-name-or-IP` with the appropriate value.
+You should also confirm that the ``DriverName`` element (found immediately
+above the ``DataSource`` element) is set to ``mysql``.
+
+If you correct an error, restart the Mattermost server by navigating to
```
hmhealey: Seems unnecessary to tell them to restart, but I don't feel strongly about that

source/install/trouble_mysql.rst
```diff
@@ -0,0 +1,250 @@
...
+.. note:: MySQL users are fully defined by their username and the host that they access MySQL from. These elements are separated by the ``@`` sign. The ``%`` character is a wild card indicating that the user can access MySQL from any IP address or host. If the user you created accesses MySQL from a specific host such as ``10.10.10.2``, please substitute that IP address for ``%``.
+
+If the user and host combination that you created does not exist, you
+will see an error such as: ::
+
+    [2017/09/20 17:06:18 EDT] [INFO] Pinging SQL master database
+    [2017/09/20 17:06:18 EDT] [EROR] Failed to ping DB retrying in 10 seconds 
+    err-Error 1045: Access denied for user 'mmuser'@'localhost' (using password: YES)
+
+**Checking that mmuser Exists**
```
hmhealey: We just refer to the mmuser account as "The Database User" above, so we should probably be consistent with that

source/install/trouble_mysql.rst
```diff
@@ -0,0 +1,250 @@
...
+
+If ``'mmuser'@'%'`` does not exist, create this user by logging into
+MySQL as root and issuing the command: 
+:samp:`create user 'mmuser'@'%' identified by '{mmuser-password}';`.
+
+After creating a user, ensure that this user has rights to the
+``mattermost`` database by following the instructions given in
+:ref:`mysql_grants`.
+
+**User Exists**
```
hmhealey: This section is repeating the "Database Exists" section from above. Maybe it would make sense to include them both in a "Database and User Exist" section below?

source/install/trouble_mysql.rst
```diff
@@ -0,0 +1,268 @@
+.. _mysql-installation-troubleshooting:
+
+MySQL Installation Troubleshooting
+----------------------------------
+
+Before you can run the Mattermost server, you must first install and
+configure a database. You can then start Mattermost by navigating to
+``/opt/mattermost/bin`` and entering the command ``sudo -u mattermost ./platform``.
```
hmhealey: Specifying file paths and how to start the server seems unnecessary here since we're focused on the database here. The paths may also be incorrect depending on how Mattermost is installed (eg through the GitLab Omnibus or Docker)

source/install/trouble_mysql.rst
```diff
@@ -0,0 +1,268 @@
...
+If the Mattermost server cannot connect to the database, Mattermost
+will fail to start. This section deals with MySQL database issues that
+you may encounter when you start up Mattermost for the first time.
+
+How you install MySQL varies depending upon which Linux distribution you
+use, but once MySQL is installed the configuration instructions are the
+same for all supported distributions. You must create a ``mattermost`` database
+and a ``mattermost`` database user. Failure to create these database
+objects or improperly referencing them from the Mattermost configuration
+file, ``/opt/mattermost/config/config.json``, prevents Mattermost from
```
hmhealey: Same comment as above. We can probably just refer to this as config.json without the full path since we have the user edit it in the installation guides

source/install/trouble_mysql.rst
```diff
@@ -0,0 +1,268 @@
...
+starting. The troubleshooting tips given here deal with these specific
+issues.
+
+Before proceeding, confirm that your MySQL server is running. You can do
+this by issuing the following command: ::
+
+    mysqladmin -u root -p status
```
hmhealey: Is it possible to do this with the regular mysql command? I think this command needs to be installed separately from the regular MySQL client.

This also assumes that the MySQL server is running on the same machine which it likely won't be. We should include whatever options are required to check if a database on another machine is running.

source/install/trouble_mysql.rst
```diff
@@ -0,0 +1,268 @@
...
+After creation of the database, attempt to restart the Mattermost server
+by navigating to the ``/opt/mattermost/bin`` directory and entering the
+command ``sudo -u mattermost ./platform``.
+
+**The mattermost Database Exists**
+
+If the ``mattermost`` database does exist, confirm that you have set up
+the database driver correctly in the
+``/opt/mattermost/config/config.json`` file. Open this file in a text
+editor, and review the value of ``"DataSource"``. It should be: ::
```
hmhealey: We usually refer to these settings by their full name so this would be `DatabaseSettings.DataSource` and below we'd have `DatabaseSettings.DriverName`

source/install/trouble_mysql.rst
```diff
@@ -0,0 +1,268 @@
...
+The Database User
+~~~~~~~~~~~~~~~~~
+
+During installation, you create a MySQL database user from the *mysql*
+prompt by issuing the following command:::
+
+    create user 'mmuser'@'%' identified by '{mmuser-password}';`
+    
+The ``mmuser-password`` value is a placeholder for the password you chose.
+You may also have specified an IP address rather than the wild card
```
hmhealey: You more fully explain the wildcard below, so this sentence is unneeded

source/install/trouble_mysql.rst
```diff
@@ -0,0 +1,268 @@
...
+**Checking that mmuser Exists**
+
+To check that this user exists log in to MySQL as the root user: ::
+
+    mysql -u root -p
+
+When prompted, enter the root password that you chose 
+when installing MySQL. From the ``mysql`` prompt enter the following command: ::
+
+    select User, Host from mysql.user;
```
hmhealey: Maybe change this to `select User, Host from mysql.user where User = 'mmuser';` to more clearly show what the user should be seeing? They'll usually see the other accounts even if mmuser doesn't exist, so they might think that's close enough to what's shown in the example

hmhealey:
Assigning myself so I remember to followup

amyblais:
@peterdamian @hmhealey Is this PR still WIP or ready for review, do you have any questions / need help with anything here?

amyblais:
Closing this as there are a lot of merge conflicts and it would be easier to start this over.

PR #1532, opened 2017-10-02
Docs for data retention (E20)

jasonblais:
@grundleborg It would be particularly helpful to get your help to review this FAQ:

```
Why didn't an old file get deleted after running the deletion job?
```


source/administration/data-retention.rst
```diff
@@ -0,0 +1,68 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Setting custom policies for each team and channel are in the roadmap but not yet supported.
+
+If you are interested in this feature, consider upvoting the `existing feature proposal <https://mattermost.uservoice.com/forums/306457-general/suggestions/31731844-ee-data-retention-policy-for-individual-teams-and>`_ and share your feedback in the comments.
+
+How is data retention handled in the mobile apps?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When a message or file is deleted, they are als
```
lindy65: Not sure where the rest of the sentence is?

source/administration/config-settings.rst
```diff
@@ -2443,6 +2443,60 @@ The workflow for failover without downing the server is to change the database l
...
+
+and
+
++-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| This feature's ``config.json`` setting is ``"FileRetentionDays": 365`` with whole number input.                                                                               |
++-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+
+Data Deletion Time
+^^^^^^^^^^^^^^^^^^^
+Set the start time of the daily scheduled data retention job. Choose a time when fewer people are using your system. Must be a 24-hour time stamp in the form HH:MM.
```
grundleborg: maybe note that this is in the "local time" of the server?

source/administration/data-retention.rst
```diff
@@ -0,0 +1,68 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The file attachment is removed from the Mattermost user interface, and deleted from the FileInfo table and from your local disk or Amazon S3 service as specified in **System Console > Files > Storage**.
+
+Why didn't an old file get deleted after running the deletion job?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First, check the data deletion was successful in the deletion job table in **System Console > Advanced > Data Retention Policy (Beta)**.
+
+Second, if the files were uploaded prior to Mattermost v4.2, you will need to delete the files manually from your local file storage or Amazon S3 storage. You may do this by simply deleting the relevant folders. // XXX George, any other info/help we can provide?
```
grundleborg: So prior to 4.2 the folder structure was something like `teamid/channelid/fileid/...` so the only way to delete files is either to figure out which one and manually delete the right bits, or to just delete all these folders, removing all files from pre 4.2. My advice is to wait until the data retention period you want gives all of pre-4.2 data to be past the cutoff, and then to manually delete the whole lot. In practice, I think there will be very few if any E20 customers who have pre 4.2 systems who apply data retention policy.

source/administration/data-retention.rst
```diff
@@ -0,0 +1,68 @@
...
+
+Note that these files will still be removed from the Mattermost messages if they were uploaded in Mattermost v3.5 or later, which contain the `FileInfo table <https://docs.mattermost.com/administration/changelog.html#database-changes-from-v3-4-to-v3-5>`_. 
+
+Why do I see ``Pending`` in the deletion job table with no details?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This means you haven't restarted your server after enabling the data retention policy. Restart your server and try again.
+
```
grundleborg: Only if there's no "in progress" data retention job AND they stay pending for more than ~1 minute. Otherwise, they might just be waiting for a slot to run the job in to become available.

source/administration/data-retention.rst
jasonblais: We already know of a few customers who have run Mattermost prior to v4.2 and are interested in this feature, so I think a few will hit this issue.

That being said, is the recommendation to:
 1. enable file retention policy
 2. wait until all files uploaded to the folder `teamid/channelid/fileid/...` fall into the data retention policy duration
 3. manually delete all these folders

Is there a way for the customer to know when the appropriate timing for 2 is? Like a particular time field in the Files table? Since they may not remember when they upgraded to v4.2.

grundleborg: The 3 steps are correct. I think it's actually even simpler: once the appropriate time has elapsed they can just delete the `teams/` folder in the root of their Mattermost storage directory/bucket. I'll check with HH to make sure there's nothing else in that folder, but I'm pretty sure there isn't.

As for *when* they can get away with deleting it, I can probably construct a (very slow on large installs) query they can run on a read-replica to estimate that date. But ideally, they'd be able to figure it out (at least roughly) on their own.

jasonblais: @grundleborg Updated the FAQ. 

There are some customers with ~ 1 year retention policy (already running Mattermost pre-4.2) so we might need to be prepared to give them extra support on it. But agree the number will be small so we can just handle them on a case-by-case basis

source/administration/data-retention.rst
```diff
@@ -0,0 +1,73 @@
...
+    :maxdepth: 2
+
+Set Up Guide
+--------------
+
+To set a custom data retention policy:
+
+1. Go to **System Console > Advanced > Data Retention Policy (Beta)**
+2. Set your desired retention policy for messages. Messages, including file attachments, older than the duration you set will be deleted nightly. The minimum message retention time is one day.
+3. Similarly, set your desired retention policy for file uploads. File uploads older than the duration you set will be deleted nightly from your file storage system, either from your local disk or your Amazon S3 service as specified in **System Console > Files > Storage**. The minimum file retention time is one day.
```
esethna: Does the file retention number have to be < the message retention number? Since messages include file attachments, it would be meaningless to set file retention > message retention?

source/administration/data-retention.rst
```diff
@@ -0,0 +1,73 @@
...
+
+1. Wait until all files uploaded prior to Mattermost v4.2 are past the file retention policy duration.
+2. Delete the ``teams/`` folder in the root of your  Mattermost storage directory.
+
+Note that these files will still be removed from the Mattermost user interface if they were uploaded in Mattermost v3.5 or later, which contain the `FileInfo table <https://docs.mattermost.com/administration/changelog.html#database-changes-from-v3-4-to-v3-5>`_. 
+
+Why do I see ``Pending`` in the deletion job table with no details?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This usually means another data retention job is in progress. You can verify this in the deletion job table in **System Console > Advanced > Data Retention Policy (Beta)**.
```
esethna: `another data retention job is in progress`: What about another job type? For example an LDAP sync or elasticsearch index is running?

source/administration/data-retention.rst
jasonblais: It can be longer -- If you have message retention = 7 days, and file retention = 14 days, then
 - messages are deleted from the DB after 7 days
 - messages (including files) are deleted from the UI after 7 days
 - files are deleted from file storage after 14 days.

^Let me know if this should be made more clear though

source/administration/data-retention.rst
jasonblais: I think that shouldn't affect it, but @grundleborg maybe you can help confirm?

grundleborg: That's correct - other job types don't have any effect on it.

PR #1578, opened 2017-10-26
Add quick guides to react component testing

saturninoabril:
Add quick guides to react component testing.

### DIFF
--- a/source/developer/webapp-component.rst
+++ b/source/developer/webapp-component.rst
@@ -178,3 +178,129 @@ Testing your Component
 The last required piece of building a webapp component is to test it. That can be done using the component testing framework described in this blog post:
 
 https://grundleborg.github.io/posts/react-component-testing-in-mattermost/
+
+Quick guides on how to do component testing
+~~~~~~~~~~~~~~~~~~
+
+1. Match snapshot using default or expected props. Consider global `mm_config` as necessary. Note that while snapshot is convenient, we require not to rely solely on this for every test case as this is easy to overlook by initiating `jest --updateSnapshot` without carefully inspecting the change.
+
+.. code-block:: javascript
+  :linenos:
+
+  test('should match snapshot, not send email notifications', () => {
+      global.window.mm_config.SendEmailNotifications = 'false';
+      const wrapper = shallow(<EmailNotificationSetting {...requiredProps}/>);
+
+      expect(wrapper).toMatchSnapshot();
+  });
+
+2. Add verification to important element.
+
+.. code-block:: javascript
+  :linenos:
+
+  expect(wrapper.find('#emailNotificationImmediately').exists()).toBe(true);
+
+
+3. Simulate event and verify state changes accordingly.
+
+.. code-block:: javascript
+  :linenos:
+
+  test('should pass handleChange', () => {
+      const wrapper = mountWithIntl(<EmailNotificationSetting {...requiredProps}/>);
+      wrapper.find('#emailNotificationImmediately').simulate('change');
+
+      expect(wrapper.state('enableEmail')).toBe('true');
+      expect(wrapper.state('emailInterval')).toBe(30);
+  });
+
+
+4. Ensure that all functions of a component are tested. This can be done via events, state changes or just calling it directly.
+
+.. code-block:: javascript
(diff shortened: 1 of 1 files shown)


source/developer/webapp-component.rst
```diff
@@ -178,3 +178,129 @@ Testing your Component
...
+.. code-block:: javascript
+  :linenos:
+
+  wrapper.instance().handleExpand();
+
+  expect(newUpdateSection).toBeCalled();
+  expect(newUpdateSection).toHaveBeenCalledTimes(1);
+  expect(newUpdateSection).toBeCalledWith('email');
+
+5. When a function is passed to a component via props, make sure to test it if it gets called for a particular event calls or state changes..
```
lindy65: Not sure whether this should be:
`...make sure to test it if it gets called for a particular event call or state change.`
OR
`...make sure to test it if it gets called for a particular event call or its state changes.`

saturninoabril: Thanks @lindy65! I'd be happy to know which one is correct. :)

lindy65: I'd be inclined to go for the 2nd option grammatically Saturn but let's wait for JW to weigh in from a dev point of view in case I am misunderstanding the process.

jasonblais:
@saturninoabril Thanks for writing up this guide!

Does it differ from George's blog post? https://grundleborg.github.io/posts/react-component-testing-in-mattermost/

saturninoabril:
@jasonblais I guess the blog from George discussed the overall structure of React component testing. However, it fell short in describing how to write detailed test cases that will cover as much as possible regarding the specifics and functions of a component.
The guides I provided might not be complete or might have entries that are not necessary, but I guess we could make this as a baseline, and just update or modify if needed.

source/developer/webapp-component.rst
```diff
@@ -178,3 +178,129 @@ Testing your Component
 The last required piece of building a webapp component is to test it. That can be done using the component testing framework described in this blog post:
 
 https://grundleborg.github.io/posts/react-component-testing-in-mattermost/
+
+Quick guides on how to do component testing
+~~~~~~~~~~~~~~~~~~
+
+1. Match snapshot using default or expected props. Consider global `mm_config` as necessary. Note that while snapshot is convenient, we require not to rely solely on this for every test case as this is easy to overlook by initiating `jest --updateSnapshot` without carefully inspecting the change.
```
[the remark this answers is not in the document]
saturninoabril: Snapshot is a serialized value representing the React tree generated by Jest and saved in a text file. Full details [here](http://facebook.github.io/jest/docs/en/snapshot-testing.html).

What I observed in the past is that, our React component testing usually have snapshot testing only. While it should be enough, we tend to overwrite the previous snapshot for every component change/modification without carefully inspecting what has been changed (say CSS, text, elements change, etc.). In addition, by just relying on snapshot, we tend to overlook those small but important items which ideally should have their own assertions.

What I meant in the `note` above is that we should have additional assertions on top of snapshots.

source/developer/webapp-component.rst
```diff
@@ -178,3 +178,129 @@ Testing your Component
...
+  jest.mock('react-router/es6', () => ({
+      browserHistory: {
+          push: jest.fn()
+      }
+  }));
+
+Finally, initiate the following commands:
+
+1. Run `yarn run test:watch`, select `p` and type filename, and see if the component test passed.
+2. Run `yarn run test:coverage`, and open corresponding html report at `coverage/components` folder to see the percentage covered. Update the test if necessary. Note that it's not required to meet 100% coverage of a component especially if it will take unnecessary or complicated mock up. Uncovered lines, statements, branch or functions will just be recorded so that it will be covered by next test layers like integrations or end-to-end testing.
```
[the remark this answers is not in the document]
saturninoabril: I intentionally put blank on this figure since I don't really have a recommended value. It depends whether the code is testable or not (although we should aim that the codes written should be testable).
Maybe we could start to say 80%+ and then anything below will be on a case to case basis.

jasonblais:
Thanks @saturninoabril and @lindy65 -- two minor tweaks based on your feedback. Good for dev review

PR #1586, opened 2017-10-31
Fixed 2 typos in training process

jespino:


jespino:
I only tried to fix the typo maintaining what i though was the original intention. Anyway, I like the proposal of change it to text, but the numbers have more "visual" impact (IMHO), and emphasize that we are talking about a bug that are having impact in a lot of people. Probably I would change it to 100.000. But I'm 1/5 about it, is simply an opinion :)

jasonblais:
@jespino Let's change it to `100,000`. Great catch, thanks for submitting the fixes

jespino:
Done. You are welcome. :blush:

PR #1600, opened 2017-11-03
Update bulk-loading-data-format.rst

lindy65:
Fixed up quotes on json keys

### DIFF
--- a/source/deployment/bulk-loading-data-format.rst
+++ b/source/deployment/bulk-loading-data-format.rst
@@ -43,25 +43,25 @@ The following fragment is from a file that imports two teams, each with two chan
 .. code-block:: javascript
   :linenos:
 
-  { type: "version", ... }
-  { type: "team", team: { name: "TeamA", ...} }
-  { type: "team", team: { name: "TeamB", ...} }
-  { type: "channel", channel: { team: "TeamA", name: "ChannelA1", ...} }
-  { type: "channel", channel: { team: "TeamA", name: "ChannelA2", ...} }
-  { type: "channel", channel: { team: "TeamB", name: "ChannelB1", ...} }
-  { type: "channel", channel: { team: "TeamB", name: "ChannelB1", ...} }
-  { type: "user", user: { username: "user001", ...} }
-  { type: "user", user: { username: "user002", ...} }
-  { type: "user", user: { username: "user003", ...} }
-  { type: "user", ... }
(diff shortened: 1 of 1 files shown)


source/deployment/bulk-loading-data-format.rst
```diff
@@ -79,8 +79,8 @@ For clarity, the object is shown using regular JSON formatting, but in the data
 .. code-block:: javascript
```
jasonblais: Above this code block, there is one more that doesn't have the proper quotes, below `
The following fragment is from a file that imports two teams, each with two channels, many users, and many posts.`

source/deployment/bulk-loading-data-format.rst
```diff
@@ -79,8 +79,8 @@ For clarity, the object is shown using regular JSON formatting, but in the data
...
 
   {
-    type: "version",
-    version: 1
+    "type": "version",
+    "version": "1"
```
jasonblais: Actually numbers don't have to be in quotes.

source/deployment/bulk-loading-data-format.rst
```diff
@@ -125,13 +125,13 @@ For clarity, the object is shown using regular JSON formatting, but in the data
...
-    type: "O",
-    description: "The Team Description",
-    allow_open_invite: true
+  "type": "team",
+  "team": {
+    "name": "team-name",
+    "display_name": "Team Display Name",
+    "type": "O",
+    "description": "The Team Description",
+    "allow_open_invite": "true"
```
jasonblais: @grundleborg Should `true` be in quotes or not?

source/deployment/bulk-loading-data-format.rst
lindy65: Thanks, have fixed :)

source/deployment/bulk-loading-data-format.rst
lindy65: Thanks, have fixed :)

source/deployment/bulk-loading-data-format.rst
grundleborg: None of the values (the stuff to the right of the `:` character) should be changed - they already have quotes if they need them. It's just all the keys that need fixing (the stuff to the left of the `:` character). `true` should *not* be in quotes.

PR #1604, opened 2017-11-03
Updated release-process.md

amyblais:
Added one task to the checklist - needs a PM confirmation that I put it in the right spot.

### DIFF
--- a/source/process/release-process.md
+++ b/source/process/release-process.md
@@ -365,7 +365,6 @@ Once bug fix release is ready to cut:
     - Update [ci-linux-mysql-prev](https://ci-linux-mysql-prev.mattermost.com) to the previous release version
 5. Dev:
     - Delete RCs after final version is shipped    
-    - Check if any libraries need to be updated in all repositories, including `mattermost-server`, `mattermost-webapp`, `desktop`, `mattermost-mobile` and `mattermost-redux`
     - Test the GitLab RC containing the Mattermost final bits
     - Confirm gitlab.mattermost.com is updated to final build
     - Merge changes made to release branch into `master`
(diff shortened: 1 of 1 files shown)


source/process/release-process.md
```diff
@@ -224,6 +224,7 @@ The final release is cut. If an urgent and important issue needs to be addressed
     - Review any changes made to install guides, and test if necessary
     - Run loadtests against the final release build to confirm there are no performance issues
     - Review [Security Policies](https://docs.mattermost.com/process/security.html) page is updated
```
lindy65: This wasn't part of AB's updates but wondering whether this should read:
1. Review changes made to [Security Policies](https://docs.mattermost.com/process/security.html) page
or
2. Ensure [Security Policies](https://docs.mattermost.com/process/security.html) page has been updated

source/process/release-process.md
```diff
@@ -224,6 +224,7 @@ The final release is cut. If an urgent and important issue needs to be addressed
...
+    - Update dependancies after release branch is cut
```
jasonblais: Actually, at T-0 looks like there's an item

```
Check if any libraries need to be updated in all repositories, including `mattermost-server`, `mattermost-webapp`, `desktop`, `mattermost-mobile` and `mattermost-redux`
```

but I think moving it to T-2 makes sense, and calling it dependancies. @amyblais Maybe we can just tweak to say

```
Update dependancies after release branch is cut in `mattermost-server`, `mattermost-webapp`, `desktop`, `mattermost-mobile` and `mattermost-redux`
```

source/process/release-process.md
jasonblais: @lindy65 I'd go with two, agree

jasonblais:
Thanks, @amyblais also removed the dependancy update from T-0 to avoid duplicate tasks

PR #1610, opened 2017-11-07
Adding groupmsg slash command to documentation

jespino:
Wait until groupmsg command is merged mattermost/mattermost-server#7419

### DIFF
--- a/source/developer/slash-commands.rst
+++ b/source/developer/slash-commands.rst
@@ -36,7 +36,7 @@ Each Mattermost installation comes with some built-in slash commands that are re
     "/logout", "Log out of Mattermost", "/logout"
     "/me *{message}*", "Do an action", "/me Hello World"
     "/msg *{@username}* *{message}*", "Send a Direct Message to a user", "/msg @alice hello"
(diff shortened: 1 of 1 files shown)


source/developer/slash-commands.rst
```diff
@@ -36,6 +36,7 @@ Each Mattermost installation comes with some built-in slash commands that are re
     "/logout", "Log out of Mattermost", "/logout"
     "/me *{message}*", "Do an action", "/me Hello World"
     "/msg *{@username}* *{message}*", "Send a Direct Message to a user", "/msg @alice hello"
+    "/groupmsg *{@username,...}* *{message}*", "Sends a Group Message to the specified users", "/groupmsg @alice,@bob hello"
```
lindy65: Not sure whether there should be a space between `@alice,@bob...`?

jespino: Good catch. I have changed it. Probably is more clear that you can use spaces in this way, and with the "," is already clear that you can use it without spaces.

jasonblais:
@jespino We should also update the table at https://github.com/mattermost/docs/blob/master/source/help/messaging/executing-commands.rst :)

jespino:
Ups, I didn't see it. I'll add another PR.

PR #1611, opened 2017-11-07
PLT-7623: Support questions on dealing with general sysadmin fail.

grundleborg:
https://mattermost.atlassian.net/browse/PLT-7623


grundleborg:
@lindy65 actually, systemd is correct - it's the name of the program.

lindy65:
Thanks @grundleborg, have reverted :)

PR #1619, opened 2017-11-13
Basic edit for language

seenthattinker:


source/overview/auth.rst
```diff
@@ -1,69 +1,78 @@
...
+Core Security
+-------------
 
-Security features for authentication 
--------------------------------------------
+The core security features straddle the Mattermost Team Edition and the Mattermost Enterprise Editions.
+In all versions of Mattermost users have the ability to:
 
-A core set of features is available with all authentication options to help increase security: 
+- `Set session length <https://docs.mattermost.com/administration/config-settings.html#id33>`_, allowing administrators to define how long a user can use Mattermost before needing to re-enter credentials.
```
lindy65: For readability perhaps: "....how long a user can be logged in to Mattermost..."?

seenthattinker: Yes, simple and to the point, however, it does not imply the consequence, which is that you will be kicked off your session. While it might be tacitly understood, it might be better to just state is clearly and unequivocally, "re-enter credentials."

PR #1625, opened 2017-11-15
4.4 Docs

esethna:


source/administration/config-settings.rst
```diff
@@ -3108,6 +3131,22 @@ Prefix on the Elasticsearch index name. Enables the use of Mattermost Elasticsea
 | This feature's ``config.json`` setting is ``"IndexPrefix": ""`` with string input      |
 +----------------------------------------------------------------------------------------+
 
+Live Indexing Batch Size
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+XXXXXX
```
esethna: @grundleborg please help with describing this setting

source/administration/config-settings.rst
```diff
@@ -3108,6 +3131,30 @@ Prefix on the Elasticsearch index name. Enables the use of Mattermost Elasticsea
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Timeout in seconds for Elasticseaerch calls.
+
++-------------------------------------------------------------------------------------------------------+
+| This feature's ``config.json`` setting is ``"RequestTimeoutSeconds": 30`` with whole number input     |
++-------------------------------------------------------------------------------------------------------+
+
+Bulk Indexing Time Window
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+XXXXXX
```
esethna: @grundleborg

source/administration/config-settings.rst
grundleborg: This setting controls how many new posts are batched together before they are added to the Elasticsearch Index. If it is set to `1`, all new posts are indexed immediately. If it is set to a number greater than 1, then they are batched together until that number of new posts is reached, and then they are all indexed in a batch. On most installs, it should be left as the default of `1` but on some very large installs (think multiple messages per second on average), it may be necessary to increase this value to avoid hitting the rate limit of your Elasticsearch cluster.

source/administration/config-settings.rst
grundleborg: This value sets the maximum length of the time window for a batch of posts being indexed by the Bulk Indexer. It works in conjunction with the hard coded limit of a number of posts in a batch, with the smaller of the two being the ultimate limiting factor for batch size.

It serves as a performance optimisation for very large installs (as a rough guideline, >10 million posts in the database). On smaller installs, it should not be changed from its default value.

source/help/settings/account-settings.rst
```diff
@@ -180,6 +180,10 @@ Teammate Name Display
 
 Configure how names are displayed in Mattermost: nickname, username or full name.
 
+Website Link Previews
+~~~~~~~~~~~~~~~~~~~~~~~~
+When available, the first web link in a message will show a preview of the website content below the message. This setting must be enabled by your System Admin.
```
jasonblais: Propose linking to https://docs.mattermost.com/administration/config-settings.html#link-previews

source/help/settings/account-settings.rst
```diff
@@ -215,10 +219,17 @@ Select what language Mattermost displays in the user interface. Options include:
 - 中文 (繁體) - Traditional Chinese
 - 日本語 - Japanese
 
+Sidebar
+--------
+
+Automatically Close Direct Messages
```
jasonblais: Propose mentioning this is experimental and needs to be enabled by the System Admin first

source/administration/config-settings.rst
```diff
@@ -3108,6 +3131,30 @@ Prefix on the Elasticsearch index name. Enables the use of Mattermost Elasticsea
 | This feature's ``config.json`` setting is ``"IndexPrefix": ""`` with string input      |
 +----------------------------------------------------------------------------------------+
 
+Live Indexing Batch Size
```
jasonblais: Do any of the new settings need to be mentioned in the elasticsearch docs? Or maybe we add a link to the config settings page somewhere in https://docs.mattermost.com/deployment/elasticsearch.html

source/help/settings/account-settings.rst
esethna: Done, thanks

source/administration/config-settings.rst
esethna: Added a note, thanks

PR #1635, opened 2017-11-17
Update to release process for GitLab MR

jasonblais:


hmhealey:
If it needs to be by the 7th of the month, will that mean that they're always lagging behind by a release? 

That might actually be a good thing since we can give them a dot release of the new version instead of having to do that later on, however it might also mean that we're supporting one more release back with security fixes since they'll want it for the last 3 versions of Mattermost they've shipped

jasonblais:
@hmhealey That's correct, they'll be one release behind.

And yeah, we'll aim to only backport two versions back for security fixes, but we'll have to work with them when the time comes if they want three instead.

jasonblais:
Also added GitLab-Mattermost doc MR to the process

jasonblais:
@esethna @hmhealey Is https://docs.google.com/document/d/1mbeu2XXwCpbz3qz7y_6yDIYBToyY2nW0NFZq9Gdei1E/edit#heading=h.ncq9ltn04isg still used for setting up the test environment?

source/process/release-process.md
```diff
@@ -396,6 +384,22 @@ Once bug fix release is ready to cut:
...
+    - Update to the [GitLab changelog](https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/CHANGELOG.md)
+    - Config.json updates to [gitlab.rb](https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-config-template/gitlab.rb.template#L1008-1171), [attributes default.rb](https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-cookbooks/gitlab/attributes/default.rb#L667-829) and [config.json.erb](https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-cookbooks/gitlab/templates/default/config.json.erb) with new TE config settings ([Example](https://gitlab.com/gitlab-org/omnibus-gitlab/merge_requests/1855))
+    - Update to [GitLab Mattermost documentation](https://docs.gitlab.com/omnibus/gitlab-mattermost/README.html)
+    - Summary of updates in Team Edition that are relevant to GitLab
+5. If the release contains a security update, email @marin and @briann in GitLab with a link to the MR.
```
lindy65: Just checking that "Briann" is indeed spelt with two "n's"?

jasonblais: Yeah, that's his GitLab handle https://gitlab.com/briann. Thanks for confirming :)

hmhealey:
@jasonblais Yes, that's correct for the testing process. I should get that moved into the proper docs at some point

jasonblais:
Sounds good, included a link in the GitLab MR process

PR #1640, opened 2017-11-20
Add docs for end-to-end testing

saturninoabril:
Add docs for end-to-end testing

### DIFF
--- a/source/developer/webapp-end-to-end-testing.rst
+++ b/source/developer/webapp-end-to-end-testing.rst
@@ -1,12 +1,12 @@
 End-to-end Testing
 ==================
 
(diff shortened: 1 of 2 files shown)


jasonblais:
@saturninoabril Awesome! Should we link to this doc in our contribution guidelines?

saturninoabril:
I'm 0/5 whether it should be link to contribution guidelines. I am thinking that it would be better when the time comes that the E2E is enabled at Jenkins PR.
I'll be proposing to devs to enable E2E test (using headless Chrome only) to Jenkins build in a week or two.  Though the E2E at the moment is mostly checking element IDs, I'd like to see those small test coverage run during PR build so we could at least guard breaking change to MM application.

saturninoabril:
Cool, thanks @lindy65!

PR #1642, opened 2017-11-20
Fix 404 links

kaakaa:
Fix links that returns 404 Not Found.

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -1926,7 +1926,7 @@ Enable Plugins
 Enable Plugin Uploads
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-**True**: Enables plugin uploads by System Admins at **Plugins > Management**. If you do not plan to upload a plugin, set to false to control which plugins are installed on your server. See `documentation <https://about.mattermost.com/default-plugins/#set-up-guide>`_ to learn more.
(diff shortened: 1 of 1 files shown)


kaakaa:
@jasonblais Thanks for changing the link! 
But it seems that the link you changed is incorrect...

`https://about.mattermost.com/default-plugins-uploads` ([docs](https://docs.mattermost.com/administration/config-settings.html#plugins-beta))
`https://about.mattermost.com/default-plugin-uploads` (The link you changed)

So the link in online document have returned 404 yet.

jasonblais:
@kaakaa I didn't actually make the change, just proposed that https://about.mattermost.com/default-plugin-uploads should work.

Or do you get a 404 error if you try to access it

kaakaa:
@jasonblais Oh..I misread it, sorry. And now I got it.
As you say, https://about.mattermost.com/default-plugin-uploads works correctly for me.

So I fix the link https://about.mattermost.com/default-plugins-uploads (returning 404) to https://about.mattermost.com/default-plugin-uploads?

jasonblais:
@kaakaa That's correct, that should work. We use https://about.mattermost.com/default-plugin-uploads also in the System Console, so if we ever changed the location of that help doc, we only need to change where https://about.mattermost.com/default-plugin-uploads redirects to.

jasonblais:
Thanks @kaakaa!

kaakaa:
@jasonblais Thanks!

And since I found same link (returning 404) in mattermost-webapp, I'll fix it later.
https://github.com/mattermost/mattermost-webapp/blob/ce861eaca2ec2a070ac66dec8c6b40aab88c25b0/i18n/en.json#L735

PR #1657, opened 2017-11-25
Error message in pwd

it33:


source/overview/faq.rst
```diff
@@ -187,6 +187,31 @@ Does Mattermost have an official website-based plug-in to offer anonymous chat t
 
 Not yet. You can `upvote the feature proposal online <https://mattermost.uservoice.com/forums/306457-general/suggestions/8810731-implement-a-site-chat-feature>`_ to add your support. If you create such a plug-in, we would love to see it open sourced and made available to the community. 
 
+
+Design Decisions
+----------------
+
+Why does Mattermost give an error message when they incorrectly enter a password?
```
ccbrown: You need an antecedent for "they". Also, this isn't really the right question. All sites give error messages when the user enters an incorrect password. Something like "Why does Mattermost disclose whether or not an account exists when a user enters an incorrect password?" would be more accurate.

source/overview/faq.rst
```diff
@@ -187,6 +187,31 @@ Does Mattermost have an official website-based plug-in to offer anonymous chat t
...
+
+For organizations who choose to deploy in such a configuration, please consider the following mitigations: 1) instead of username-password, use a single-sign-on provider in Mattermost Enterprise Edition like OneLogin, Okta or ADFS, use the open source GitLab SSO option available with Mattermost Team Edition, 2) per the recommended install instructions, use a VPN client to apply network security to your deployment, 3) enable monitoring and alerting from your proxy server to detect and isolate malicious behavior reaching your deployment. 
+
+Above all, make sure to subscribe to the `Mattermost Security Bulletin <https://about.mattermost.com/security-bulletin/>`_ and apply security patches as recommended.  
+
```
ccbrown: I'm just gonna go ahead and say I'm 5/5 on not including any arguments that make it sound like we're doing things this way for the sake of better security. They're highly debatable at best, and to those with different stances, may reflect poorly on the company and the individuals responsible for our security (e.g. me).

I would suggest keeping the first two paragraphs about principles / UX, saying "The possible security implications and reduced privacy are outweighed by these values and are considered acceptable risk.", and closing it with the usual security recommendations. But please don't say things like "if network security is not in effect, this makes it easier for attackers to exploit this...". This represents increased attack surface regardless of whether the deployment is behind a firewall or not. They're layered defenses, not alternative defenses.

PR #1664, opened 2017-11-30
Add info on how data retention works on mobile

jasonblais:
**Do not merge until v1.5 of the mobile apps are released**

### DIFF
--- a/source/administration/data-retention.rst
+++ b/source/administration/data-retention.rst
@@ -70,4 +70,11 @@ If you are interested in this feature, consider upvoting the `existing feature p
 How is data retention handled in the mobile apps?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-When messages or files are deleted, they are no longer searchable in the Mattermost mobile apps. However, in some cases, the message or file may be stored locally on the device itself. An improvement to locally wipe the app periodically of old data when data retention policy is enabled, is `scheduled for Q4/2017 <https://mattermost.atlassian.net/browse/RN-361>`_.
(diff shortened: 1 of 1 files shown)


source/administration/data-retention.rst
```diff
@@ -70,4 +70,11 @@ If you are interested in this feature, consider upvoting the `existing feature p
...
 
-When messages or files are deleted, they are no longer searchable in the Mattermost mobile apps. However, in some cases, the message or file may be stored locally on the device itself. An improvement to locally wipe the app periodically of old data when data retention policy is enabled, is `scheduled for Q4/2017 <https://mattermost.atlassian.net/browse/RN-361>`_.
+When messages or files are deleted, they are no longer searchable in the Mattermost mobile apps. 
+
+In v1.5 and later of the iOS and Android apps, messages and files are deleted from local storage in the following cases, if they exceed the retention policy duration:
+
+1. When the user opens the app.
+2. When the user puts the app into the background.
+
+In v1.4 and earlier of the mobile apps, messages and files are not cleared from local storage when the data retention policy is enabled.
```
lindy65: "....when the data retention policy is enabled." - Does this mean that even if data retention is enabled, there is no local storage clearance available in versions 1.4 and earlier?  If so, I would rather word this as:
"It is not possible to make use of the data retention policy in versions 1.4 and earlier of the mobile apps." or "The data retention policy will work with versions 1.5 and above."

jasonblais: The data retention does still work to some extend (e.g. the messages won't appear in Mattermost channels & you can't search for them). The only aspect that doesn't work is that they are not cleared from your local storage

PR #1665, opened 2017-12-03
Reference developer setup in mobile-compile-yourself.rst

jasonblais:


source/mobile/mobile-compile-yourself.rst
```diff
@@ -3,7 +3,7 @@ Build Your Own Version of the Mattermost Mobile Apps
 
 To deploy in production with privately-hosted mobile apps compiled by your organization in an Enterprise App Store:
 
-1. Compile your own iOS and Android mobile applications from the `open source repository <https://github.com/mattermost/mattermost-mobile>`_
+1. Compile your own iOS and Android mobile applications from the `open source repository <https://github.com/mattermost/mattermost-mobile>`_. Use `this guide <https://docs.mattermost.com/developer/mobile-developer-setup.html>`_ to help set up your environment and build the apps.
```
lfbrock: Can we switch this to a relative link? (If you switch to a relative link, it should link properly in PDF export which some customers keep asking for). 

It should be something like: 

:doc:`this guide <mobile-developer-setup>`

jasonblais: Updated

PR #1681, opened 2017-12-10
Fix typo - selectos to selectors

santos22:


lindy65:
Thanks for the PR @santos22 :+1:

PR #1682, opened 2017-12-11
PLT-8337: Docs correction.

mkraft:
https://mattermost.atlassian.net/browse/PLT-8337

### DIFF
--- a/source/developer/webhooks-incoming.rst
+++ b/source/developer/webhooks-incoming.rst
@@ -32,13 +32,15 @@ Let's learn how to create a simple incoming webhook that posts the following mes
 
 .. code-block:: text
 
-  curl -i -X POST -d 'payload={"text": "Hello, this is some text\nThis is more text. :tada:"}' http://{your-mattermost-site}/hooks/xxx-generatedkey-xxx
+  curl -i -X POST -H 'Content-Type: application/json' -d '{"text": "Hello, this is some text\nThis is more text. :tada:"}' http://{your-mattermost-site}/hooks/xxx-generatedkey-xxx
+  # or
+  curl -i -X POST --data-urlencode 'payload={"text": "Hello, this is some text\nThis is more text. :tada:"}' http://{your-mattermost-site}/hooks/xxx-generatedkey-xxx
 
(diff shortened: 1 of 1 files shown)


mkraft:
@ccbrown I'm not sure I understand why you're posting JSON as form data in the first place? Is this an actual use case?

ccbrown:
@mkraft You understand I'm not asking you to change anything functionally, right? What I'm suggesting is the equivalent of what you have here already.

This (what I'm suggesting):

`
curl -i -X POST --data-urlencode 'payload={"text": "Hello, this is some text\nThis is more text. :tada:"}' http://{your-mattermost-site}/hooks/xxx-generatedkey-xxx
`

Does exactly the same thing as this (what you have here):

`
curl -i -X POST -H 'Content-Type: application/x-www-form-urlencoded' -d 'payload=%7B%22text%22%3A%20%22Hello%2C%20this%20is%20some%20text%5CnThis%20is%20more%20text.%20%3Atada%3A%22%7D' http://{your-mattermost-site}/hooks/xxx-generatedkey-xxx
`

I don't know why Slack made the API the way it is, but we can't really change it right now.

ccbrown:
Or maybe you're suggesting we just don't talk about the weird API. I don't know if there's actually a legitimate reason for anyone to use it. Whether or not we have to acknowledge its existence in docs is probably a PM question.

source/developer/webhooks-incoming.rst
```diff
@@ -32,13 +32,15 @@ Let's learn how to create a simple incoming webhook that posts the following mes
 
 .. code-block:: text
 
-  curl -i -X POST -d 'payload={"text": "Hello, this is some text\nThis is more text. :tada:"}' http://{your-mattermost-site}/hooks/xxx-generatedkey-xxx
+  curl -i -X POST -H 'Content-Type: application/json' -d '{"text": "Hello, this is some text\nThis is more text. :tada:"}' http://{your-mattermost-site}/hooks/xxx-generatedkey-xxx
+  # or
+  curl -i -X POST -H 'Content-Type: application/x-www-form-urlencoded' --data-urlencode '{"text": "Hello, this is some text\nThis is more text. :tada:"}' http://{your-mattermost-site}/hooks/xxx-generatedkey-xxx
```
ccbrown: The `-H 'Content-Type: application/x-www-form-urlencoded'` is unnecessary. `--data-urlencode` implies it. 0/5 though.

PR #1688, opened 2017-12-13
Update instructions in step 8 on testing

cvitter:
Revise the path to test from (/opt/mattermost) and the command to test with (sudo -u mattermost ./bin/platform) based on change in 4.4.5 related to creating the client/plugin folder


hmhealey:
Looks good. We should also change this for the other setup guides as well if they do this.

@cvitter Would you be able to help us out with this if you get a chance?

cvitter:
@hmhealey I will take a look when I have a chance and see what I can do to test and update the docs.

PR #1689, opened 2017-12-14
Add note about the mattermost.service file

cvitter:
Clarified that PostgreSQL or MySQL After and Requires sections should not be included if the database is running on a different server.


jasonblais:
@cvitter Thanks! Does this also affect the Debian install guides? https://github.com/mattermost/docs/blob/master/source/install/install-debian-88-mattermost.rst

cvitter:
@jasonblais I believe that it should be the same however I haven't tested. Likely also the case with https://github.com/mattermost/docs/pull/1688.

jasonblais:
Thanks!

PR #1718, opened 2018-01-04
Updating Example Service file to move working directory up one level

KnechtionsCoding:
In order to prevent an issue where the working directory is too low, preventing the installation of plugins, changing the example service to be in the base mattermost directory. This also brings it inline with the enterprise installation docs.

issue here: https://forum.mattermost.org/t/jira-beta-plugin-not-available-on-4-5-0/4312


lindy65:
Thanks for the PR @KnechtionsCoding!

PR #1727, opened 2018-01-10
Add troubleshooting for ios xcrun

enahum:
Adds a troubleshooting for an issue reported by a Community member that came across this while following the mobile dev guide


source/developer/mobile-developer-setup.rst
```diff
@@ -855,3 +855,23 @@ Error message
 
 Solution
   The development version of the Mattermost app cannot be installed alongside a release version. Open ``android/app/build.gradle`` and change the applicationId from ``"com.mattermost.react.native"`` to a unique string for your app.
+
+Errors When Running 'make run-ios'
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+I just got a new machine and downloaded XCode. I used git to clone the React Native Mattermost project from Github. I have everything ready to go. I run ``make run-ios`` and see
```
hmhealey: This part isn't needed since the heading covers this

source/developer/mobile-developer-setup.rst
```diff
@@ -855,3 +855,23 @@ Error message
...
+
+Error message
+  .. code-block:: bash
+
+    xcrun: error: unable to find utility "instruments", not a developer tool or in PATH
+
+Solution
+  The fix for this is pretty simple but not at all intuitive.
```
hmhealey: Similarly, this line isn't needed if we're trying to keep the solution short

source/developer/mobile-developer-setup.rst
```diff
@@ -855,3 +855,23 @@ Error message
...
+
+  - Launch XCode and agree to the terms first.
+  - Go to **Preferences -> Locations** and you'll see an option for Command Line Tools, click the select box and choose the version of XCode you'll be using.
```
hmhealey: Is there a specific version they should be using, or do they just need to select any version?

source/developer/mobile-developer-setup.rst
enahum: sure I can remove this

source/developer/mobile-developer-setup.rst
enahum: and this

source/developer/mobile-developer-setup.rst
enahum: no specific version, and I don't think there is more than one version listed, at least not for me

hmhealey: I might edit it slightly to clarify that then

PR #1733, opened 2018-01-15
Created screenshots of the login screen with various sign-in options

james-mm:
Created screenshots of the login screen with various sign-in options


jasonblais:
Thanks @james-mm!

PR #1738, opened 2018-01-17
Match current api in SSL version of proxy config

alerque:
The non SSL version of the docs has these updates which match current (and future) api version stings in the URLs that need special treatment, but the SSL version was missing them. This brings over the same syntax.

See https://github.com/mattermost/mattermost-server/issues/8026 for interesting side effects of not having all relevant api endpoints proxied the same way.


esethna:
Needs dev review

source/install/config-ssl-http2-apache2.rst
```diff
@@ -21,30 +21,30 @@ When opened, edit it to look something like the following:
...
-		<Location /api/v3/users/websocket>
-		  Require all granted
-		  ProxyPass ws://127.0.0.1:8065/api/v3/users/websocket
-		  ProxyPassReverse ws://127.0.0.1:8065/api/v3/users/websocket
-		  ProxyPassReverseCookieDomain 127.0.0.1 mysubdomain.mydomain.com
-		</Location>
+		<LocationMatch "^/api/v(?<apiversion>[0-9]+)/(?<apiusers>users/)?websocket">
+        		Require all granted
+        		ProxyPass ws://127.0.0.1:8065/api/v%{env:MATCH_APIVERSION}/%{env:MATCH_APIUSERS}websocket
+        		ProxyPassReverse ws://127.0.0.1:8065/api/v%{env:MATCH_APIVERSION}/%{env:MATCH_APIUSERS}websocket
```
saturninoabril: Should it have `/` before `websocket` like `{env:MATCH_APIUSERS}/websocket`?

Do we have somewhere in our docs where we initially define how/what to setup with `MATCH_APIUSERS` and `MATCH_APIVERSION`? If none, then we might need to define it first.

alerque: @saturninoabril These variable aren't something the user needs to setup, they are auto generated by the expression in the `LocationMatch` directive. Also the slash is part of the match there, so if that segment exists in the requested path the segment (with trailing slash) will be placed in the proxy URL as well.

PR #1749, opened 2018-01-19
4.6 changelog update about Zoom plugin

jasonblais:
- Move note about supporting the on-prem version of Zoom to v4.7 and add a known issue to 4.6 changelog. If customers report it as a bug, we'll consider shipping 4.6.1 to fix the bug.


jasonblais:
@jwilander For context, the customer who requested on-prem Zoom server support is not in a rush and can wait to v4.7. (I don't think they're using the plugin yet)

PR #1757, opened 2018-01-26
Update desktop download links

amyblais:


jasonblais:
@amyblais Great! Dev review isn't needed, but I'll wait to Monday before merging.

PR #1770, opened 2018-02-01
Add information about the telemetry endpoint

Tristramg:
Users have observed that the client sends information to segment.io if the telemetry is activated.

Some can react very suspiciously, as it is documented nowhere that the information is gathered from the client and to a third party provider.


lindy65:
Thanks for the PR @Tristramg !

We'll get it through the review process soonest :)

PR #1774, opened 2018-02-06
Update the build instructions to use Go 1.9.2

kemenaran:
Mattermost now strongly recommends using Go 1.9.2—and warns if the 1.9.0 version is used.

Update the docs to reflect this.

### DIFF
--- a/source/developer/dev-setup-centos-7.rst
+++ b/source/developer/dev-setup-centos-7.rst
@@ -35,11 +35,11 @@ Set up your development environment for building, running, and testing Mattermos
 
     a. Download the Go binary.
 
-       ``wget https://storage.googleapis.com/golang/go1.9.linux-amd64.tar.gz``
+       ``wget https://storage.googleapis.com/golang/go1.9.2.linux-amd64.tar.gz``
 
     b. Install the Go binary.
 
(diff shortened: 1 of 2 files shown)


lindy65:
Thanks for the PR @kemenaran :) We'll get this through the review process soonest...

kemenaran:
@ccbrown why not.

That said the main goal of this PR was to fix the instructions for the minimal Go required version (as defined by [pre-check.sh](https://github.com/mattermost/mattermost-server/blob/034dbc07e3068c482e654b6a1a8fcbe4b01c44f3/scripts/prereq-check.sh#L38)). It can also wait for the release of Go 1.9.4 though (especially if the platform plans to make this new version mandatory soon).

ccbrown:
We won't be making 1.9.4 mandatory. 1.9.2 is only mandatory because it includes fixes that Mattermost simply won't function securely or properly without. To my knowledge, that won't be true of 1.9.4, but we'll want developers / build machines to be using it for their own security. If we're going to walk people through Go installation, let's try to keep them secure instead of instructing them to install a vulnerable toolchain. :)

ccbrown:
Go 1.9.4 is released. Build machines are now using it and so should devs.

I've gone ahead and updated your PR. Thanks for the contribution! :D

kemenaran:
@ccbrown thanks!

PR #1776, opened 2018-02-07
document sign-in method under security, not general

lieut-data:
Update the documentation for LDAP configuration to reference the sign-in method under the `Security` tab in account settings, not `General`:

image


jasonblais:
@lindy65 whoops - I merged before your review.

PR #1781, opened 2018-02-09
PLT-8726 - update cd path instruction in step 11 of upgrade

AndersonWebStudio:
update cd path instruction in step 11 of upgrade

### DIFF
--- a/source/administration/upgrade.rst
+++ b/source/administration/upgrade.rst
@@ -86,7 +86,7 @@ Owner and group of the install directory - *{owner}* and *{group}*
 
   .. code-block:: text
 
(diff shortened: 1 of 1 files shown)


AndersonWebStudio:
@jasonblais no i don't think anything changed- unless the root install path changed recently. I went through the upgrade process myself with TLS and this person is correct, that instruction needs an adjustment.

AndersonWebStudio:
@jasonblais i almost forgot there is another mistake in the list. Using a vanilla ubuntu 16.04 user the step 6 backup step also needs to be sudo'd. I will commit another change.

PR #1791, opened 2018-02-13
Document custom post types for integrations and remove usage of 'payload' in outgoing webhooks and slash commands docs

jasonblais:
@jwilander Wondering if we have a list of post types that can be used?

### DIFF
--- a/source/developer/slash-commands.rst
+++ b/source/developer/slash-commands.rst
@@ -233,7 +233,7 @@ For example, you can use the following response to override the profile picture 
 Custom post type
 ~~~~~~~~~~~~~~~~~~
 
-You can specify a custom post type `via client plugins <about.mattermost.com/default-plugins>`_ when sending a webhook message. To set the type, use the `type` parameter.
+You can specify a custom post type when sending a webhook message, for use by `plugins <about.mattermost.com/default-plugins>`_. To set the type, use the `type` parameter on the JSON response.
 
 .. code-block:: text
 
--- a/source/developer/slash-commands.rst
+++ b/source/developer/slash-commands.rst
@@ -230,6 +230,15 @@ For example, you can use the following payload to override the profile picture t
 .. note::
   `Enable integrations to override profile picture icons <https://docs.mattermost.com/administration/config-settings.html#enable-integrations-to-override-profile-picture-icons>`_ must be set to `true` in `config.json` to override usernames. Enable them from **System Console > Integrations > Custom Integrations** or ask your System Administrator. If not enabled, the icon of the creator of the webhook URL is used to post messages.
 
+Custom post type
+~~~~~~~~~~~~~~~~~~
+
+You can specify a custom post type when sending the webhook message. To set the type, use the `type` parameter of the JSON payload.
+
+.. code-block:: text
+
+  payload={"username": "webhook-bot", "text": "Hello, this is some text\nThis is more text. :tada:", "type": "custom_type_here"}
+
 Tips and Best Practices
 ------------------------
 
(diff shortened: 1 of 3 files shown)


source/developer/slash-commands.rst
```diff
@@ -230,6 +230,15 @@ For example, you can use the following payload to override the profile picture t
...
   `Enable integrations to override profile picture icons <https://docs.mattermost.com/administration/config-settings.html#enable-integrations-to-override-profile-picture-icons>`_ must be set to `true` in `config.json` to override usernames. Enable them from **System Console > Integrations > Custom Integrations** or ask your System Administrator. If not enabled, the icon of the creator of the webhook URL is used to post messages.
 
+Custom post type
+~~~~~~~~~~~~~~~~~~
+
+You can specify a custom post type when sending the webhook message. To set the type, use the `type` parameter of the JSON payload.
+
+.. code-block:: text
+
+  payload={"username": "webhook-bot", "text": "Hello, this is some text\nThis is more text. :tada:", "type": "custom_type_here"}
```
jwilander: Slash command responses should just be JSON, there shouldn't be a `payload=` at the start. Looks like we have that other places on this page and we should remove it

source/developer/slash-commands.rst
```diff
@@ -230,6 +230,15 @@ For example, you can use the following payload to override the profile picture t
 .. note::
   `Enable integrations to override profile picture icons <https://docs.mattermost.com/administration/config-settings.html#enable-integrations-to-override-profile-picture-icons>`_ must be set to `true` in `config.json` to override usernames. Enable them from **System Console > Integrations > Custom Integrations** or ask your System Administrator. If not enabled, the icon of the creator of the webhook URL is used to post messages.
 
+Custom post type
+~~~~~~~~~~~~~~~~~~
+
+You can specify a custom post type when sending the webhook message. To set the type, use the `type` parameter of the JSON payload.
```
jwilander: Might be a good idea to mention that this is mainly useful for plugins (and not much else right now)

source/developer/webhooks-outgoing.rst
```diff
@@ -173,6 +173,15 @@ For example, to create a message with a heading, and an italicized text on the n
...
 Messages with advanced formatting can be created by including an :doc:`attachment array <message-attachments>` and :doc:`interactive message buttons <interactive-message-buttons>` in the JSON payload.
 
+Custom post type
+~~~~~~~~~~~~~~~~~~
+
+You can specify a custom post type when sending the webhook message. To set the type, use the `type` parameter of the JSON payload.
+
+.. code-block:: text
+
+  payload={"username": "webhook-bot", "text": "Hello, this is some text\nThis is more text. :tada:", "type": "custom_type_here"}
```
jwilander: Same as slash commands, there's no `payload=` at the start of responses, looks like we have it incorrect on other parts of this page as well

jasonblais:
@jwilander Updated

source/developer/webhooks-incoming.rst
```diff
@@ -149,6 +149,15 @@ For example, to create a message with a heading, and an italicized text on the n
 
 Messages with advanced formatting can be created by including an :doc:`attachment array <message-attachments>` and :doc:`interactive message buttons <interactive-message-buttons>` in the JSON payload.
 
+Custom post type
+~~~~~~~~~~~~~~~~~~
+
+You can specify a custom post type `via client plugins <about.mattermost.com/default-plugins>`_ when sending a webhook message. To set the type, use the `type` parameter on the JSON payload.
```
jwilander: >via client plugins

You can set the post type via anything that's using the webhook/slash command/whatever. It's just that the plugins are the only thing that will use the post type right now

Maybe change this to `You can specify a custom post type when sending a webhook message, for use by <plugins>(somelinkhere). To set the type, use the `type` parameter on the JSON payload.`

I'd also leave out the "client" bit of plugins, since the server part of plugins can use the types too

jasonblais:
@jwilander Done.

PR #1799, opened 2018-02-16
Update instructions for configuring systemd service

kemenaran:
Starting from Mattermost 4.8, the platform server can notify systemd when it is actually started up and ready to accept network connections.

This makes the service status more reliable: systemd will now report the server as going from "stopped" to "activating" and finally "running". If the server exits before signaling it is ready, this will be correctly reported as an error to systemd. And executing `systemctl mattermost start` will return control to the shell when the server is actually ready–or return with an error exit code in case of startup failure.

(More details in https://github.com/mattermost/mattermost-server/pull/8296)

### To be merged only after Mattermost 4.8 is released

### DIFF
--- a/source/install/install-debian-88-mattermost.rst
+++ b/source/install/install-debian-88-mattermost.rst
@@ -74,8 +74,9 @@ Assume that the IP address of this server is 10.10.10.2.
     Requires=postgresql.service
 
     [Service]
-    Type=simple
+    Type=notify
     ExecStart=/opt/mattermost/bin/platform
+    TimeoutStartSec=90
     Restart=always
     RestartSec=10
     WorkingDirectory=/opt/mattermost
--- a/source/install/install-debian-88-mattermost.rst
+++ b/source/install/install-debian-88-mattermost.rst
@@ -76,7 +76,7 @@ Assume that the IP address of this server is 10.10.10.2.
     [Service]
     Type=notify
     ExecStart=/opt/mattermost/bin/platform
-    TimeoutStartSec=90
+    TimeoutStartSec=3600
     Restart=always
     RestartSec=10
     WorkingDirectory=/opt/mattermost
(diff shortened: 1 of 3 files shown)


source/install/install-debian-88-mattermost.rst
```diff
@@ -74,8 +74,9 @@ Assume that the IP address of this server is 10.10.10.2.
     Requires=postgresql.service
 
     [Service]
-    Type=simple
+    Type=notify
     ExecStart=/opt/mattermost/bin/platform
+    TimeoutStartSec=90
```
kemenaran: NB: this is the start timeout. If the server doesn't send the ready notification within 90s, systemd will consider it as stuck (which makes sense, for instance if the server is waiting for an unavailable DB connection).

90s is the default system-wide value provided by systemd. I prefer to make it explicit in this file, to make it clear that the server taking too long to start could result in the process being killed.

That said, it means that if for instance a database migration is performed on the main thread before the server is signaled as ready, it should take no more of 90s, otherwise the process will be flagged as unresponsive and killed. 

Do database migrations happen in the background, or synchronously on startup? And if so, what is the expected maximum duration of these migrations?

jasonblais:
I've added a `Merge On Hold` label until 4.8 is released

source/install/install-debian-88-mattermost.rst
ccbrown: Synchronously on startup. The duration totally depends on the size of the deployment, but 90 seconds is definitely within the realm of possibility. I would suggest something a little more extreme to be safe, like 3600.

lindy65:
@kemenaran - can you take a look at `ccbrown's` request please?

kemenaran:
Definitely! I'm traveling this week, with limited Internet access – but I'll amend the patch when I get back in a week. As this should not be merged before the 4.8 release anyway, this should be fine.

lindy65:
Perfect! Thank you 👍🏻 Safe travels!

PR #1812, opened 2018-02-19
Updated Ubuntu 16.04 unit file so service (re-)starts after automatic database updates

torgeirl:
Closes [mattermost-server issue 8239](https://github.com/mattermost/mattermost-server/issues/8239).

**Issue**
On Ubuntu, the daily apt upgrade will stop Mattermost if it's about to apply an update to the database. If you follow the current instructions the mattermost.service will continue to run, but the server will be down (nginx reports `HTTP 502 bad gateway`).

**Solution**
A slight change to the unit file ensures that automatic update of the database cause the mattermost.service to be (re-)started.

Of note, applying these change to an existing installation requires your to re-enable the service, but I opted to not go into those details as the guide assumes you haven't enabled it yet.

### DIFF
--- a/source/install/install-ubuntu-1604-mattermost.rst
+++ b/source/install/install-ubuntu-1604-mattermost.rst
@@ -90,10 +90,15 @@ Assume that the IP address of this server is 10.10.10.2.
     If you are using MySQL, replace ``postgresql.service`` with ``mysql.service`` in 2 places in the ``[Unit]`` section and 1 place in the ``[Install]`` section.
 
   .. note::
-    If you have installed MySQL or PostgreSQL on a dedicated server then you need to remove the ``After=postgresql.service`` and ``Requires=postgresql.service`` or ``After=mysql.service`` and ``Requires=mysql.service`` lines in the ``[Unit]`` section and replace the ``WantedBy=postgresql.service`` or ``WantedBy=mysql.service`` line in the ``[Install]`` section with ``WantedBy=multi-user.target`` or the Mattermost service will not start.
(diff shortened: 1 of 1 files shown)


source/install/install-ubuntu-1604-mattermost.rst
```diff
@@ -84,13 +84,16 @@ Assume that the IP address of this server is 10.10.10.2.
     LimitNOFILE=49152
 
     [Install]
-    WantedBy=multi-user.target
+    WantedBy=postgresql.service
```
crspeller: This only works for running the DB server on the same machine as the Mattermost app server. Which is not the recommended configuration. Let's have this as the generic with a note at the bottom to change this if you are running the MM server on the same machine as the DB.

torgeirl: @crspeller: the section of the documentation where I'm suggesting changes defaults to them running on the same machine, and rather notes what you should to different *if* you are running the database on a dedicated server:

> **Note**
> If you have installed MySQL or PostgreSQL on a dedicated server then you need to remove the `After=postgresql.service` and `Requires=postgresql.service` or `After=mysql.service` and `Requires=mysql.service` lines in the `[Unit]` section or the Mattermost service will not start.

Are you suggesting we keep it `WantedBy=multi-user.target`, and instead remove both`After=postgresql.service` and `Requires=postgresql.service` and flip the notes?

crspeller: Ahh. Sorry I assumed our default would be a multi machine setup. This is fine then.

PR #1817, opened 2018-02-20
v4.7.1 docs

jasonblais:
@amyblais Please help with review? This might be something you can take on next dot release.


amyblais:
+1, but one question: final was cut today so would the release date be 19th?

jasonblais:
We usually make it the day we announce the release (e.g. 4.7.0 is cut 1 or 2 days before the release day on the 16th)

PR #1826, opened 2018-02-21
Update config-settings.rst

amyblais:


source/administration/config-settings.rst
```diff
@@ -2145,7 +2145,7 @@ When false, disables file downloads on mobile apps. Users can still download fil
 Image Proxy
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Includes three configuration keys: ``ImageProxyType``, ``ImageProxyURL`` and ``ImageProxyOptions``. When these keys are configured, posts served to the client will have their markdown modified enabling all images to be loaded through a proxy. See `documentation <https://docs.mattermost.com/administration/image-proxy>`_ for more details.
+Includes three configuration keys: ``ImageProxyType``, ``ImageProxyURL`` and ``ImageProxyOptions``. When these keys are configured, posts served to the client will have their markdown modified enabling all images to be loaded through a proxy. See `documentation <https://docs.mattermost.com/administration/image-proxy.html?highlight=image%20proxy>`_ for more details.
```
jasonblais: can we remove the `?highlight=image%20proxy`? I think that's not needed.

amyblais: It was like that before and the doc link wasn't working, so I added the whole link (I tested and it works now).

jasonblais: If the URL is, https://docs.mattermost.com/administration/image-proxy.html?highlight=image%20proxy, it'll highlight all words containing `image` or `proxy`. 

So I think https://docs.mattermost.com/administration/image-proxy.html would be enough. Currently it's missing the `html` part which is probably why the link is broken.

amyblais: Cool, fixed.

PR #1827, opened 2018-02-21
Update sso-ldap.md

james-mm:
Updated FAQ: AD/LDAP sync fails to perform a scan in LDAP if there are 0 LDAP users in Mattermost. If you have at least 1 LDAP user in Mattermost, the sync will complete.


james-mm:
@lindy65 Updated per recommendation re Group Messages (GM)

PR #1831, opened 2018-02-22
Add saml auth migration command documentation

jespino:
It includes 2 python scripts that needs to be tested with OneLogin and Okta accounts, to verify that works (I'm not 100% because I haven't accounts there).

And this PR include the documentation of saml migration which will be included in the 4.8 version of mattermost, so must be merged accordingly to that.

### DIFF
--- a/source/administration/command-line-tools.rst
+++ b/source/administration/command-line-tools.rst
@@ -752,12 +752,15 @@ platform user invite
       sudo ./platform user invite user@example.com myteam
       sudo ./platform user invite user@example.com myteam1 myteam2
 
-platform user migrate_auth (to ldap)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+platform user migrate_auth
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
   Description
-    Migrates all user accounts from one authentication provider to another. For example, you can upgrade your authentication provider from email to AD/LDAP. Output will display any accounts that are not migrated successfully.
+    Migrates all user accounts from one authentication provider to another. For example, you can upgrade your authentication provider from email to AD/LDAP, or from AD/LDAP to SAML. Output will display any accounts that are not migrated successfully.
 
+**Migrate to AD/LDAP**
+
+  Parameters
     -  ``from_auth``: The authentication service from which to migrate user accounts. Supported options: ``email``, ``gitlab``, ``saml``.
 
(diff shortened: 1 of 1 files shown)


source/administration/command-line-tools.rst
```diff
@@ -752,8 +752,8 @@ platform user invite
       sudo ./platform user invite user@example.com myteam
       sudo ./platform user invite user@example.com myteam1 myteam2
 
-platform user migrate_auth
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+platform user migrate_auth (to ldap)
```
jasonblais: Maybe we keep the command name the same but have a subsection for migrating to LDAP, and another for migrating to SAML

jasonblais:
@jespino I created an Okta and OneLogin account for you. Let me know if you need additional permissions to use the scripts.

Also, should we prepare one for ADFS as well?

PS: Great work on this doc 👍

jespino:
Probably I will need some administration privileges to test my scripts, so, I will try to create my own account and create some users, to verify that the script works, at least, for a basic configuration.

jasonblais:
Thanks @jespino  - let me know if you need any help with that.

jespino:
Tested on a basic okta and on a basic onelogin configuration, and everything goes well.

About ADFS, I'm not know well ADFS, but I think is not so easy, because it is not a service like Okta or OneLogin, so there is not a clear API to manage users. I think it simply expect that you use Active Directory directly to manage accounts, and ADFS give you the layer over Active Directory to Single Sign On service.

So the answer here would be, use directly Ldap protocol to extract the users information.

source/administration/command-line-tools.rst
jespino: I though this way was clearer, but if you prefer I can try to split on sections.

jasonblais:
@jespino Sounds good for ADFS. 👍 

I had a few minor changes that I tried to push to my own branch, but for some reason I couldn't..so I ended up committing directly into yours. Hope that's okay?

Changes are here: https://github.com/mattermost/docs/pull/1831/commits/ae41a79067c73dec3c8cbb75c0f0590b4dd142b1

Feedback on the proposed changes are more than welcome. They're mostly minor wording updates, and combining the two sections under the `platform user migrate_auth` command. The reason is that we've linked this doc section elsewhere and if we renamed it, those links would break.

jasonblais:
Adding `Merge on Hold` label so we don't merge until v4.8 is ready. 

@jespino As noted above, feedback on my changes is more than welcome :)

PR #1833, opened 2018-02-22
WIP: Update config-settings.rst

amyblais:
@esethna I got feedback that the config settings for the Experimental Group Unreads Channels here is wrong since the config.json can only take two options: true and false.

From George: "the document you've linked seems to contradict itself. How can there be three options default_on, default_off and disabled when the config.json can only take two options: true and false".

A customer tried to set ExperimentalGroupUnreadChannels to either default_on or default_off in the ServiceSettings block of config.json but they had to end up setting it like this: "ExperimentalGroupUnreadChannels": true

Do you know how the config settings doc should be fixed?


jasonblais:
@amyblais Can you file a ticket for this? I think this is a bug.

The config.json setting options should be `disabled`, `default_off` and `default_on`. So `true` nor `false` should work.

amyblais:
Ticket: https://mattermost.atlassian.net/browse/MM-9684

PR #1838, opened 2018-02-25
Fix developer setup links in remaining two places

tejasbubane:
I had fixed one in https://github.com/mattermost/docs/pull/1739.

These two instances of old URL were remaining in the codebase.


jasonblais:
Thank you!!

PR #1844, opened 2018-03-02
Known issue for plugin sandboxing error message

jasonblais:


source/administration/changelog.md
```diff
@@ -170,6 +170,7 @@ Multiple setting options were added to `config.json`. Below is a list of the add
 - Numbered lists can sometimes extend beyond the normal post area.
 - Slack import through the CLI fails if email notifications are enabled.
 - Push notifications don't always clear on iOS when running Mattermost in High Availability mode.
+- `[WARN] unable to create user namespace: fork/exec /proc/self/exe: no such file or directory` log message is posted when sandboxing isn't enabled for plugins. If you don't use plugins, you can ignore this message. If you have plugins enabled, [learn how to enable sandboxing](https://developers.mattermost.com/extend/plugins/security/#sandboxing).
```
ccbrown: There are a few reasons sandboxing might not work. Maybe instead of quoting this specific warning about /proc/self/exe, it would be better to quote the more generic warning that follows it?

> [WARN] plugin sandboxing is not supported. plugins will run with the same access level as the server

jasonblais: @ccbrown Updated

source/administration/changelog.md
```diff
@@ -170,6 +170,7 @@ Multiple setting options were added to `config.json`. Below is a list of the add
 - Numbered lists can sometimes extend beyond the normal post area.
 - Slack import through the CLI fails if email notifications are enabled.
 - Push notifications don't always clear on iOS when running Mattermost in High Availability mode.
+- `[WARN] plugin sandboxing is not supported. plugins will run with the same access level as the server` log message is created when sandboxing isn't enabled for plugins. If you don't use plugins, you can ignore this message. If you have plugins enabled, [learn how to enable sandboxing](https://developers.mattermost.com/extend/plugins/security/#sandboxing).
```
lindy65: Should `plugins' at the beginning of the 2nd sentence have a capital letter?

ccbrown: The error has it in lowercase.

PR #1849, opened 2018-03-05
Updates docs to be explicit about JSONL bulk import format.

mkraft:


mkraft:
@grundleborg I think it's required for hyperlinking with that syntax.

mkraft:
@grundleborg 2 options:

```rst
this is a link_ and then some more text.

.. _link: http://www.mattermost.com/
```
or
```rst
this is a `link
<http://www.mattermost.com/>`_ and then some more text.
```
I have zero preference.

PR #1850, opened 2018-03-06
Update mobile-developer-setup.rst

CoolMoeDee:
Hey,
as mentioned in the native mobile apps channel.
Please read to verify: https://github.com/fastlane/fastlane/issues/3920
read: biolinh commented on 7 Apr 2016

worked for me to solve my issue.


source/developer/mobile-developer-setup.rst
```diff
@@ -411,7 +411,7 @@ Setting up Gradle Variables
 Setting up Environment Variables
 ++++++++++++++++++++++++++++++++
 In order to use the **make build-android** command, you'll need to set a few environment variables. In this guide, we will explain some of them. You can refer to the `env_vars_example <https://github.com/mattermost/mattermost-mobile/blob/fastlane/fastlane/env_vars_example>`_
-file under the fastlane directory to see all of them.
+file under the fastlane directory to see all of them. In order to make fastlane work with these environment variables, copy the file with your variables to ../mattermost-mobile/fastlane/.env where .env is the file name.
```
enahum: Actually I did not know you could do that, what I always do is setting the env vars in the terminal from where I’m running make build-* normally I have a `sh` file with a bunch of **exports** one for each variable and then I’ll source that file.

Perhaps we should document both ways

CoolMoeDee: ok, interesting. the reason why this didnt work for me, is probably because i left out the "source" part. i just "chmod +x" on my "env_vars" file and then "sh env_vars" in the cli where i ran "make build-*" later. ofcourse it's a good idea to document both ways.
should i submit another PR, including both options?
I'm new to github, so I'm not sure right now if I can modify this PR.

enahum:
No need for another PR just make the changes into this one and the push the commit :)

lindy65:
@CoolMoeDee, 

Thanks for the PR 👍  Could you rebase and we'll get it through PM review :)

Thanks!

PR #1860, opened 2018-03-11
Updated boolean values for short attribute

cjohannsen81:
Had to change the fields back to boolean as Strings are ignored. Changed the description of the settings as well.


cjohannsen81:
@jasonblais Hey Jason the issue is that String values are ignored and lowercase characters are not working in JavaScript. My pull request stated that we need capital first letters because lowercase will fail :)

jasonblais:
@cjohannsen81 For me it's the opposite actually. Lowercase passes but capital first letters fail..hmm. 

Which Mattermost version are you testing on?

PR #1862, opened 2018-03-14
Adding a note to config-settings about `EnableAPIV3` setting

amyblais:


amyblais:
It already appears as false when I view the preview of the doc in this PR, so somebody made the change already in the 4.8 branch. Closing.

PR #1866, opened 2018-03-16
Adding note to docs re Forward80To443 related to Let's Encrypt 

lindy65:
Added note that Forward80To443 config.json setting should also be set to true when enabling Let's Enrypt.

### DIFF
--- a/source/install/config-ssl-http2-apache2.rst
+++ b/source/install/config-ssl-http2-apache2.rst
@@ -7,7 +7,8 @@ Configuring Apache2 with SSL and HTTP/2 (Unofficial)
 
 Once you've configured Apache2 as a proxy for your Mattermost Server, the easiest way to enable SSL on Apache2 is via Let's Encrypt and `Certbot <https://certbot.eff.org/#ubuntuxenial-apache>`_.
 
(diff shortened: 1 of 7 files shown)


source/install/config-ssl-http2-apache2.rst
```diff
@@ -7,6 +7,8 @@ Configuring Apache2 with SSL and HTTP/2 (Unofficial)
 
 Once you've configured Apache2 as a proxy for your Mattermost Server, the easiest way to enable SSL on Apache2 is via Let's Encrypt and `Certbot <https://certbot.eff.org/#ubuntuxenial-apache>`_.
 
+.. note:: If Let’s Encrypt is enabled, forward port 80 through a firewall, with `Forward80To443 <https://docs.mattermost.com/administration/config-settings.html#forward-port-80-to-443>`_ ``config.json`` setting set to ``true`` to complete the Let’s Encrypt certification.
```
jasonblais: I think for the `note` to format correctly at doc.mattermost.com, this part needs to be written as 
```
.. note::
   If Let's Encrypt is enabled...
```

Similarly for the others

lindy65: Thanks @jasonblais - I'll fix these up and submit for re-review!

jasonblais:
Thanks!

PR #1870, opened 2018-03-16
4.8 Docs corrections

amyblais:


source/administration/changelog.md
```diff
@@ -107,6 +107,10 @@ Multiple setting options were added to `config.json`. Below is a list of the add
 
  - It is required that any new integrations use API v4 endpoints. For more details, and for a complete list of available endpoints, see [https://api.mattermost.com/](https://api.mattermost.com/).
  - All API v3 endpoints have been deprecated and are scheduled for removal in Mattermost v5.0.
+ 
+#### RESTful API v4 Changes
+
+ - Created `NSUrlSessionUploadTask` endpoint for the iOS share extension to upload a file as octet-stream. This enables uploading files larger than 20 Mb.
```
jasonblais: I think based on https://github.com/mattermost/mattermost-api-reference/pull/335/, and previous API change entries in the changelog, the correct API change would actually be something like

```
 - Updated `POST /files to support requests with only the `channel_id` and `filename` defined as query parameters with the contents of a single file in the body of the request.
```

(I took the above note from https://github.com/mattermost/mattermost-api-reference/pull/335/files#diff-41243d95a14d0c1e53733b273f4ad5beR10)

PR #1872, opened 2018-03-18
v4.8 Simple Correction

timconner:
Changed `users` to `user` in the Performance section of Release v4.8.


lindy65:
Thanks for the pull request @timconner 👍

PR #1884, opened 2018-03-21
MM-9859 Fix Windows CURL example for incoming webhook.

crspeller:


source/developer/webhooks-incoming.rst
```diff
@@ -40,7 +40,7 @@ If you're running cURL on Windows, you need to wrap the payload with double quot
```
jespino: This line say that you have to wrap the payload with `"` instead of `'`, and the example doesn't do that.

esethna: @crspeller ^

crspeller: Removed the bad text.

PR #1902, opened 2018-03-30
Create welcome_email.rst

james-mm:
Welcome email template for end users

### DIFF
--- a/source/getting-started/welcome_email.rst
+++ b/source/getting-started/welcome_email.rst
@@ -1,0 +1,55 @@
+Welcome Email to End Users
+===========================
+
+
+To make it easy for your end users to start using Mattermost right away, we created a sample email template that you can use.
+
+Remember to replace all the items below in bold with your information.
+
+
+Email Template
+-------------------
+
+From: **[company name]** IT Team
+
+To: End users
+
+Subject: New Communication Platform - Mattermost
+
+
+Hi all,
+ 
+As some of you already know, we are moving to Mattermost as our communication platform. Mattermost is a messaging app where you can talk, share files, and collaborate on projects or initiatives. Mattermost also integrate with many of the apps that you use everyday, like **[add apps]**.  
+
+We are moving to Mattermost because it will allow us to become more productive by being able communicate in one place instead of using email, instant chat, and a multitude of other programs.
+
+Some of the major benefits of using Mattermost are:
+
+- Direct 1:1 and group messaging
+
+- Channels for topic-based, group-based, or meeting-based chat
+
+- Streamlined collaboration on projects
+
+- Reduced email clutter
+
+- Searching across messages and channels
+
+- Sharing files
+
+To get started:
+
+1. Open a browser on your computer, go to **[Mattermost URL]** and log in with your **[LDAP/AD, SAML, Google, etc]** credentials. 
+
+2. `Download <https://about.mattermost.com/download/#mattermostApps>`_ the Mattermost apps for desktop and mobile. Refer to the `Mattermost Welcome Guide <https://docs.mattermost.com/help/getting-started/welcome-to-mattermost.html>`_ on how to get up and running quickly.
+
+
+3. Start messaging! 
+
+
+Questions?
+If you have any questions, feel free to post in the **[~Mattermost channel]** or email us at **[IT email]**.
+
+Happy messaging!
+
+**[company name]** IT Team
--- a/source/getting-started/welcome_email.rst
+++ b/source/getting-started/welcome_email.rst
@@ -19,7 +19,7 @@ Subject: New Communication Platform - Mattermost
 
 Hi all,
  
-As some of you already know, we are moving to Mattermost as our communication platform. Mattermost is a messaging app where you can talk, share files, and collaborate on projects or initiatives. Mattermost also integrate with many of the apps that you use everyday, like **[add apps]**.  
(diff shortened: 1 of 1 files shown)


source/getting-started/welcome_email.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Hi all,
+ 
+As some of you already know, we are moving to Mattermost as our communication platform. Mattermost is a messaging app where you can talk, share files, and collaborate on projects or initiatives. Mattermost also integrates with many of the apps that you use every day, like **[add apps]**.  
+
+We are moving to Mattermost because it will allow us to become more productive by being able communicate in one place instead of using email, instant chat, and a multitude of other programs.
+
+Some of the major benefits of using Mattermost are:
+
+- Direct 1:1 and group messaging
```
esethna: Propose linking each point to the appropriate section of the end user docs

source/getting-started/welcome_email.rst
```diff
@@ -0,0 +1,55 @@
...
+
+- Channels for topic-based, group-based, or meeting-based chat
```
esethna: Docs

source/getting-started/welcome_email.rst
```diff
@@ -0,0 +1,55 @@
...
+
+- Streamlined collaboration on projects
+
+- Reduced email clutter
+
+- Searching across messages and channels
+
+- Sharing files
```
esethna: Docs

source/getting-started/welcome_email.rst
```diff
@@ -0,0 +1,55 @@
...
+
+- Direct 1:1 and group messaging
+
+- Channels for topic-based, group-based, or meeting-based chat
+
+- Streamlined collaboration on projects
+
+- Reduced email clutter
+
+- Searching across messages and channels
```
esethna: Docs

source/getting-started/welcome_email.rst
```diff
@@ -0,0 +1,55 @@
...
+
+- Sharing files
+
+To get started:
+
+1. Open a browser on your computer, go to **[Mattermost URL]** and log in with your **[LDAP/AD, SAML, Google, etc]** credentials. 
+
+2. `Download <https://about.mattermost.com/download/#mattermostApps>`_ the Mattermost apps for desktop and mobile. Refer to the `Mattermost Welcome Guide <https://docs.mattermost.com/help/getting-started/welcome-to-mattermost.html>`_ on how to get up and running quickly.
```
esethna: Should this step be optional?

source/getting-started/welcome_email.rst
```diff
@@ -0,0 +1,55 @@
...
+
+- Reduced email clutter
+
+- Searching across messages and channels
+
+- Sharing files
+
+To get started:
+
+1. Open a browser on your computer, go to **[Mattermost URL]** and log in with your **[LDAP/AD, SAML, Google, etc]** credentials.
```
esethna: We should make it clear that at any time they can click the Mattermost URL to go back to the service. It's not just a one time link for first time users

esethna: Doesn't have to be part of these steps, but it should be mentioned somewhere how to get back to the Mattermost

source/getting-started/welcome_email.rst
```diff
@@ -0,0 +1,55 @@
...
+To: End users
+
+Subject: New Communication Platform - Mattermost
+
+
+Hi all,
+ 
+As some of you already know, we are moving to Mattermost as our communication platform. Mattermost is a messaging app where you can talk, share files, and collaborate on projects or initiatives. Mattermost also integrates with many of the apps that you use every day, like **[add apps]**.  
+
+We are moving to Mattermost because it will allow us to become more productive by being able communicate in one place instead of using email, instant chat, and a multitude of other programs.
```
esethna: Intro seems a bit wordy. Maybe it would be better to use our tag line. Something like,

"Mattermost will host all our communication in one place, is instantly searchable and available from all your devices."

james-mm: Updated

source/getting-started/welcome_email.rst
james-mm: The end of the email points to the end user getting started guide, which does this in more detail: https://docs.mattermost.com/help/getting-started/welcome-to-mattermost.html.

source/getting-started/welcome_email.rst
james-mm: See above

source/getting-started/welcome_email.rst
james-mm: See above

source/getting-started/welcome_email.rst
james-mm: See above

source/getting-started/welcome_email.rst
james-mm: Updated

source/getting-started/welcome_email.rst
james-mm: Desktop and mobile are a big part of our customers' every use case. They can skip if they don't want to download. The welcome guide is essential to them getting up and running quickly on MAttermost.

source/getting-started/welcome_email.rst
```diff
@@ -0,0 +1,55 @@
...
+
+- Reduced email clutter
+
+- Searching across messages and channels
+
+- Sharing files
+
+To get started:
+
+1. Open a browser on your computer, go to **[Mattermost URL]** and log in with your **[LDAP/AD, SAML, Google, etc]** credentials. Remember to bookmark the URL as that you can use it next time to log in.
```
esethna: Remember to bookmark the URL so you can use it to log in next time.

PR #1910, opened 2018-04-03
Clean up service enable commands

tomo667a:
we use the "systemctl enable" command to enable services on RHEL 7.
When we use the chkconfig command, we get the message:

  Note: Forwarding request to 'systemctl enable *.service'.


lindy65:
Thanks for the PR @tomo667a 👍 

We'll get it through the review process as soon as possible...

jasonblais:
Thanks @tomo667a! 

@lindy65 changes look okay to me, but I don't really understand them enough to approve. Propose we get a dev to help review?

PR #1918, opened 2018-04-04
Add Timezone feature to account settings

amyblais:
Question for Eric: since this is experimental, do we add this setting to the docs at this point?


amyblais:
@esethna I'm not sure what the URL will look like until the 4.9 config settings will show up in the documentation after 4.9 branch is merged. Is it ok if I open a quick correction PR on Monday to add the link? I'll make sure it gets done.

PR #1946, opened 2018-04-19
Adding a note to desktop notification settings

amyblais:
In response to: https://forum.mattermost.org/t/solved-in-any-channel-any-message-from-anyone-triggers-a-desktop-notification/4811/6?u=amy.blais

### DIFF
--- a/source/help/getting-started/configuring-notifications.rst
+++ b/source/help/getting-started/configuring-notifications.rst
@@ -67,8 +67,7 @@ keywords <https://docs.mattermost.com/help/settings/account-settings.html#words-
    Notifications**. By default, all channels use the global setting
    configured in *Account Settings*.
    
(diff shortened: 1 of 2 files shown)


source/help/getting-started/configuring-notifications.rst
```diff
@@ -67,6 +67,9 @@ keywords <https://docs.mattermost.com/help/settings/account-settings.html#words-
    Notifications**. By default, all channels use the global setting
    configured in *Account Settings*.
    
+.. note::
```
esethna: Is a note styling necessary for this? Do we have standards for when we use notes and other styling vs text?

This is only for desktop app though right? We should specify in the first few words and maybe link to the download page. 

Also specify which Mattermost icon are we talking about?

PR #1971, opened 2018-04-30
Add a step in all install guides to configure Site URL

jasonblais:


source/install/ee-prod-rhel-6.rst
```diff
@@ -366,7 +364,12 @@ Test setup and configure Mattermost Server
    ``system_admin`` role, which gives you access to the System Console.
 3. From the ``town-square`` channel click the dropdown and choose the
    ``System Console`` option
-4.  Update **Notification** > **Email** settings to setup an SMTP email service. The example below assumes AmazonSES.
+4. Set the Site URL:
+
+ - In the **GENERAL** section of the System Console, click **Configuration**.
+ - In the **Site URL** field, set the URL that users point their browsers at. For example, *https://mattermost.example.com*. If you are using HTTPS, make sure that you set up TLS, either on Mattermost Server or on a proxy.
+
+5.  Update **Notification** > **Email** settings to set up an SMTP email service. The example below assumes AmazonSES.
```
jespino: A small, not important at all, and not needed to be change, but in this line is an extra space between the `5.` and the start of the sentence. It happens in all the files. As I said, is not important at all.

PR #1973, opened 2018-05-01
Fixing broken links throughout docs

amyblais:


source/install/ee-prod-rhel-6.rst
```diff
@@ -302,7 +302,7 @@ Set up NGINX with SSL (Recommended)
   server {
      listen         80;
      server_name    mattermost.example.com;
-     return         301 https://$server_name$request_uri;
+     return         301 ``https://$server_name$request_uri``;
```
amyblais: Not 100% if this link should actually work or if it's just an example

PR #1974, opened 2018-05-01
MM-10413: Redirect EE install guides to TE guides, and make EE the default binary

jasonblais:
List of changes below. Note that each guide already had EE as the default binary, so I only had to add proper redirects.
1) Redirect Ubuntu 14.05 EE guide to TE. The guides were already identical.
2) Redirect Ubuntu 14.05 EE guide to TE. The guides were already identical.
3) Redirect RHEL 7.1 EE guide to TE. The guides were already identical.
4) Redirect RHEL 6.6 EE guide to TE. The guides were already identical.
5) Redirect Debian EE guide to TE. The guides were already identical.
6) Production Docker guide is already installing EE by default. There is an option at the bottom of the doc to download TE if someone wants to.


jasonblais:
@amyblais If you can review first. There may be mistakes. Key thing is to check:

1) All redirects work correctly.
2) Links to the EE guides are replaced everywhere in our docs and our .com website. I only found ../install/ee-install.rst file in docs, fixed with [this commit](https://github.com/mattermost/docs/compare/jasonblais-install-guides?expand=1#diff-ca0f091bc7c7aa39ba81272c13a62402), but there may be others. A quick check would be helpful.

Once reviewed, Lindsay and Corey should review and approve.

amyblais:
1. The redirects in the docs don't work when I click on the links in preview (except in Update ee-install.rst) - is there a way to test them locally?
2. Didn't find other docs / pages to update.

PR #1988, opened 2018-05-07
Update command-line-tools.rst

lfbrock:
@jasonblais can you confirm the proposed change is correct? I think we now support switching to SAML auth as well? 

(Also do we support migrating from LDAP to SAML? I wasn't sure on that, if so we probably need to add ``ldap`` to the from auth field as well).

### DIFF
--- a/source/administration/command-line-tools.rst
+++ b/source/administration/command-line-tools.rst
@@ -767,7 +767,7 @@ platform user migrate_auth
   Parameters
     -  ``from_auth``: The authentication service from which to migrate user accounts. Supported options: ``email``, ``gitlab``, ``saml``.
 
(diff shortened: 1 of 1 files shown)


jasonblais:
@lfbrock We separated this into two sections, one for migrating to LDAP and another to SAML. We should make that more clear though if it's not obvious

image

The separation was made to make the migration steps between the two systems more clear.

amyblais:
@lfbrock Are changes needed based on Jason's comments or is this ready for review?

lfbrock:
Oh cool, yeah that covers it. 

It's definitely a little hard to find in the doc since it's so big, but I'm not sure if there's a better way to handle it since it seems like they need separate instructions. I'll close this PR for now, and maybe as a long term thing we can think about if there's any improvements we can make for how we structure these long docs.

PR #2008, opened 2018-05-14
Add GitLab TE --> EE migration guide

jasonblais:


source/install/ee-install.rst
```diff
@@ -41,6 +41,17 @@ Upgrading to Enterprise Edition from Mattermost Team Edition
 
 If you need to migrate Team Edition prior to install, `please follow the migration guide. <http://docs.mattermost.com/administration/migrating.html>`_
 
+Upgrading to Enterprise Edition in GitLab Omnibus
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+GitLab Omnibus runs the open source Mattermost Team Edition. To upgrade to Mattermost Enterprise Edition, follow these steps:
+
+1) Disable the built-in Mattermost instance on // XXX Harrison: Where do they go disable it?
```
jasonblais: Harrison: If you can help with this step

hmhealey: To disable it, they go open `/etc/gitlab/gitlab.rb` and set the following line to false
```ruby
mattermost['enable'] = false
```
And then they run `sudo gitlab-ctl reconfigure` to apply the updated settings

hmhealey:
We might want to also tell them to migrate the database or (if possible) point the EE instance at GitLab's Mattermost database. I'm not 100% sure how to do that and if the database is even ran after shutting off the internal Mattermost instance, so I'd have to look more into how to do that if we want it

source/install/ee-install.rst
```diff
@@ -41,6 +41,26 @@ Upgrading to Enterprise Edition from Mattermost Team Edition
...
+
+ - Go to ``/etc/gitlab/gitlab.rb`` and set the following line to false
+
+   .. code-block:: text
+     mattermost['enable'] = false
+
+ - Run `sudo gitlab-ctl reconfigure` to apply the updated setting
+
+2. Install Mattermost using `one of the guides above <https://docs.mattermost.com/install/ee-install.html#installing-enterprise-edition>`_.
+3. Migrate the database used by GitLab Mattermost for your new Enterprise Edition instance.
```
hmhealey: I made a small wording tweak here since the sentence sounded backwards (migrating to GitLab Mattermost)

PR #2009, opened 2018-05-15
MM-10264: Adds docs for permissions export/import CLI commands.

mkraft:


amyblais:
@mkraft Is this for v5.0? If yes, I'll wait for next release before merging; if not, then I'll be merging this as soon as it's reviewed.

jespino:
Maybe makes sense to include a comment about this commands only works when you have a valid license?

Actually, if permissions stuff only will work on E20 (I'm 2/5 about it), maybe we have mention exactly that.

mkraft:
@jespino I think what you're proposing makes sense.

@amyblais Yes it's for v5.0. Is there a PM perspective on if/how to document enterprise-only CLI commands? If not I can add something adhoc.

amyblais:
@jasonblais Also Martin's question above: "Is there a PM perspective on if/how to document enterprise-only CLI commands? If not I can add something adhoc."

PR #2033, opened 2018-05-18
Add note on MySQL 8 Support

cvitter:


cvitter:
BTW - formatting on the code block example doesn't look right to me. I don't have the RST Markdown 100% yet clearly.

PR #2044, opened 2018-05-24
Proposing download links

it33:
Propose adding download links to changelog so we can add a new menu option in Desktop App for `Desktop app -> Help -> Check for newer version` which would go the changelog and allow users to notice a new version, and download it right away. 

image

I've been waiting for 4.1.1 for a long time because of the loading animation. It was very difficult for me to check for a new version. This modification to our changelog process makes it easier to check for newer versions and immediately download once one is found. 

If we do this, propose creating a redirect link from mattermost.com so we don't hardcode to a page, and instead make a link that can be redirected easily in future.

### DIFF
--- a/source/help/apps/desktop-changelog.rst
+++ b/source/help/apps/desktop-changelog.rst
@@ -4,9 +4,13 @@ Desktop Application Changelog
 Release v4.1.1
 ----------------------------
 
-Release date: May 17, 2018
-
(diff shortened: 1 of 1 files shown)


jasonblais:
Deferring to ES now as the owner of desktop app. If we want to add it, it's easy to add into product ([PR that added "Learn More" link](https://github.com/mattermost/desktop/pull/180/files#diff-30612c64fd3cc7c1d30e567a61a65eb9)).

PR #2065, opened 2018-05-30
Adding information about flow campaign

jespino:
I would like to add this campaign for flow stuff in mattermost-redux (I have to create the Help wanted ticket yet) but must be more or less easy to do with the excel already created.


amyblais:
@jespino Made some tweaks to text - one note is that I'm not able to access the spreadsheet (https://docs.google.com/spreadsheets/d/10FmRm5TgpsDIkpvqX1emwVYe33-NQShy4I7LZSZbkPA/edit#gid=0), so we want to make it public before merging this PR.

jespino:
Thanks @amyblais Now is shared properly :)

PR #2076, opened 2018-06-05
Updates to dot release process

amyblais:


amyblais:
The idea is to release in about a week. When the bug list is ready the release is started at T-7. There are no previous dates, e.g. T-15. I can reduce the dates to start maybe at T-5, but just wanted to make sure there is enough time for bug fixing and testing.

jasonblais:
@amyblais Propose we shorten the dot release cycle as much as possible. If we used this for 4.10.1, I think we'd release on or around the same date and I'd prefer it was half of that to push bug fixes out quicker.

Maybe we try shortening the proposal by half and try it out? We can always adjust it.

Other notes
 - why the 3-day separation between code complete and RC cut?
 - which day `Once bug fix release is ready to cut:` is this targeted for?

Otherwise the plan looks quite solid 👍

PR #2080, opened 2018-06-05
System notice

michaelgamble:
Added the docs update to include system notice with clarification around user visibility as part of the guideline.


amyblais:
Hey @michaelgamble - I made some grammar tweaks.

Can you help sign the CLA and then we can continue with reviews and merge this PR?

PR #2085, opened 2018-06-11
prevent installation on SAN/NAS mounts

cjohannsen81:
Added a note to prevent installation on SAN/NAS mounts as they may have a negative impact on the application server performance. Observed in customer environments.


crspeller:
You can't just blacklist SAN or NAS in general. Some people just have slow hardware. Also I know this is something we tell people we support.

PR #2087, opened 2018-06-11
Added troubleshooting for mismatched server versions in HA

icelander:


jespino:
Actually, It can be worse in some cases. With the 5.0 version we have an on-line migration of data, that means that both 4.10 and 5.0 supports the database scheme, but if the on-line migration is already done, and you still have some nodes in 4.10, some api calls can generate wrong data in the database. I don't know if we must be more "explicit" or "concerning" in the message because is really important to have consistent versions when you are using HA.

PR #2098, opened 2018-06-15
MM-10502 Update HA docs for job server

jasonblais:
Doc for https://mattermost.atlassian.net/browse/MM-10502

Ticket to print a warning message when the Job Server settings are misconfigured.

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -3851,7 +3851,7 @@ Set whether or not this Mattermost server will schedule tasks that will be compl
 
 When running Mattermost on a single machine, this setting should always be enabled.
 
-When running Mattermost in High Availablity mode, this setting should always be enabled. In a High Availability cluster, exactly one of the servers will be designated as the Scheduler at a time to ensure that duplicate tasks aren't created.
+When running Mattermost in High Availablity mode, this setting should always be disabled. In a High Availability cluster, exactly one of the servers will be designated as the Scheduler at a time to ensure that duplicate tasks aren't created.
 
 +-----------------------------------------------------------------------------------------------------------------------------------------+
 | This feature's ``config.json`` setting is ``"RunScheduler": true`` with options ``true`` and ``false`` for above settings respectively. |
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -3851,7 +3851,7 @@ Set whether or not this Mattermost server will schedule tasks that will be compl
 
 When running Mattermost on a single machine, this setting should always be enabled.
 
-When running Mattermost in High Availablity mode, this setting should always be disabled. In a High Availability cluster, exactly one of the servers will be designated as the Scheduler at a time to ensure that duplicate tasks aren't created.
+When running Mattermost in High Availablity mode, this setting should always be enabled. In a High Availability cluster, exactly one of the servers will be designated as the Scheduler at a time to ensure that duplicate tasks aren't created. See `High Availability documentation <https://docs.mattermost.com/deployment/cluster.html#job-server>`_ for more details.
 
 +-----------------------------------------------------------------------------------------------------------------------------------------+
 | This feature's ``config.json`` setting is ``"RunScheduler": true`` with options ``true`` and ``false`` for above settings respectively. |
--- a/source/deployment/cluster.rst
+++ b/source/deployment/cluster.rst
@@ -11,7 +11,6 @@ High availability in Mattermost consists of running redundant Mattermost applica
 
   This document applies to Mattermost Server version 4.0 and later. For previous versions, see :doc:`cluster-310`.
 
-
 .. contents::
   :backlinks: top
   :local:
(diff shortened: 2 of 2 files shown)


source/administration/config-settings.rst
```diff
@@ -3851,7 +3851,7 @@ Set whether or not this Mattermost server will schedule tasks that will be compl
 
 When running Mattermost on a single machine, this setting should always be enabled.
 
-When running Mattermost in High Availablity mode, this setting should always be enabled. In a High Availability cluster, exactly one of the servers will be designated as the Scheduler at a time to ensure that duplicate tasks aren't created.
+When running Mattermost in High Availablity mode, this setting should always be disabled. In a High Availability cluster, exactly one of the servers will be designated as the Scheduler at a time to ensure that duplicate tasks aren't created.
```
grundleborg: OK seeing the old docs I'm now worried something has gone badly wrong here. I'll investigate...

source/deployment/cluster.rst
```diff
@@ -265,6 +264,20 @@ In Mattermost v4.2 and later, a cluster leader election process assigns any sche
...
+^^^^^^^^^^^^^^^^
+
+Mattermost schedules and completes periodic tasks via the `job server <https://docs.mattermost.com/administration/config-settings.html#jobs>`_. These tasks include:
+
+ - LDAP sync
+ - Data retention
+ - Compliance exports
+ - Elasticsearch indexing
+
+When running in High Availablity mode, the ``JobSettings.RunScheduled`` setting should be set to ``false`` in config.json. This will make all the app servers also be Job Servers, allowing the High Availability cluster leader election process to run only a single scheduler. Otherwise, all the app servers run the scheduled jobs causing the jobs to be run multiple times (once per app server).
```
grundleborg: The assertion that it makes them become job servers is not actually implied by this setting. I need to think about how to word this better when I've investigated the above comment.

source/administration/config-settings.rst
jasonblais: @grundleborg This was recently added, so it might have also been a misunderstanding. But @hmhealey might know more about that.

hmhealey: Last we had discussed this, my understanding was that all servers would have `RunScheduler` enabled and the cluster leader would assign one of them to be the scheduler

jasonblais: @grundleborg Any updates on this one?

hmhealey: I just looked back at my chat history, and when Corey and I were discussing it, we thought that it had to be on for all servers and then the leader node would take care of figuring out which server was actually the scheduler. Perhaps we were wrong about that? It was in a DM so I can send you the chat transcript

grundleborg: I've had a look through the code and I have no idea any more. I think Corey implemented this, so it might be worth us asking him what the intended behaviour was and how it worked.

grundleborg: Discussing off-github with Corey and Harrison.

grundleborg: Update: code fix pending.

jasonblais:
@grundleborg Now that the server PR is merged, are the doc changes correct for versions 5.2 and later? https://github.com/mattermost/mattermost-server/pull/9174

source/administration/config-settings.rst
grundleborg: @jasonblais ok so now the code fix has gone in, the options are:
1. Run a single jobserver *without* the `--noschedule` flag. That jobserver will be solely responsible for scheduling recurring jobs.
2. Run all jobservers with `--noschedule` flag, and then set this config to `true` for *all* app servers in the cluster. The cluster leader at any given time will then be responsible for scheduling recurring jobs.

I think 2 is the recommended option in a properly configured HA cluster, as it provides redundancy in job scheduling capability, which would not be present if the only jobserver with scheduling responsibility were to fail.

source/deployment/cluster.rst
grundleborg: Given the comment I've made on the config setting, this paragraph now makes no sense so should be removed.

jasonblais:
Requesting another editor review given the changes.

source/deployment/cluster.rst
```diff
@@ -265,6 +264,20 @@ In Mattermost v4.2 and later, a cluster leader election process assigns any sche
...
+^^^^^^^^^^^^^^^^
+
+Mattermost schedules and completes periodic tasks via the `job server <https://docs.mattermost.com/administration/config-settings.html#jobs>`_. These tasks include:
+
+ - LDAP sync
+ - Data retention
+ - Compliance exports
+ - Elasticsearch indexing
+
+Run all job servers with ``--noschedule flag``, then set ``JobSettings.RunScheduled`` to ``true`` in config.json for all app servers in the cluster. The cluster leader at any given time will then be responsible for scheduling recurring jobs.
```
grundleborg: The setting is `JobSettings.RunScheduler`

source/deployment/cluster.rst
```diff
@@ -265,6 +264,20 @@ In Mattermost v4.2 and later, a cluster leader election process assigns any sche
 
 The process is based on a widely used `bully leader election algorithm <https://en.wikipedia.org/wiki/Bully_algorithm>`_ where the process with the lowest node ID number from amongst the non-failed processes is selected as the "leader". 
 
+Job Server
+^^^^^^^^^^^^^^^^
+
+Mattermost schedules and completes periodic tasks via the `job server <https://docs.mattermost.com/administration/config-settings.html#jobs>`_. These tasks include:
```
grundleborg: In the setup you describe below, it isn't *scheduling* periodic tasks via the jobserver, just running them on the jobserver.

source/deployment/cluster.rst
```diff
@@ -265,6 +264,20 @@ In Mattermost v4.2 and later, a cluster leader election process assigns any sche
...
+
+ - LDAP sync
+ - Data retention
+ - Compliance exports
+ - Elasticsearch indexing
+
+Run all job servers with ``--noschedule flag``, then set ``JobSettings.RunScheduled`` to ``true`` in config.json for all app servers in the cluster. The cluster leader at any given time will then be responsible for scheduling recurring jobs.
+
+For increased performance, use a standalone job server when running in High Availability mode.
```
grundleborg: This is a bit confusing - it sounds like this is an alternative to what is described above, when really it is orthogonal to it.

jasonblais:
@grundleborg how's that?

PR #2117, opened 2018-06-25
document subpath support

lieut-data:
Document subpath support. This complements the more developer focussed documentation at https://github.com/mattermost/mattermost-developer-documentation/pull/128.

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -36,17 +36,13 @@ The URL that users will use to access Mattermost. The port number is required if
 
 This field is required in Mattermost v3.8 and later.
 
+In Mattermost v5.1 and later, the URL may contain a subpath such as `https://example.com/company/mattermost`.
+
 If Site URL is not set, the following features will operate incorrectly:
 
  - email notifications will contain broken links, and email batching will not work
  - authentication via OAuth 2.0, including GitLab, Google and Office 365, will fail
  - plugins may not work as expected
-
-.. note:: Do not append a team name to the end of the site URL.
-
-Correct example: ``https://mattermost.example.com:8065``
-
-Incorrect example: ``https://mattermost.example.com/team_name``
 
 +----------------------------------------------------------------------------------------------------------------------------------------------------+
 | This feature's ``config.json`` setting is ``"SiteURL": ""`` with string input.                                                                     |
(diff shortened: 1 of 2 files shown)


source/administration/config-settings.rst
```diff
@@ -36,18 +36,14 @@ The URL that users will use to access Mattermost. The port number is required if
...
 
+In Mattermost v5.1 and later, the URL may contain a subpath such as `https://example.com/company/mattermost`.
+
 If Site URL is not set, the following features will operate incorrectly:
 
  - email notifications will contain broken links, and email batching will not work
  - authentication via OAuth 2.0, including GitLab, Google and Office 365, will fail
  - plugins may not work as expected
 
-.. note:: Do not append a team name to the end of the site URL.
```
jasonblais: This does make me wonder - what happens when someone accidentally appends the team name to the URL? Obviously email notifications will be broken, for instance, but will you be able to start the server, log in, post messages, etc?

lieut-data: I'm not sure what would happen before 5.1, but obviously that would now be treated as a subpath and the rest of the site anchored accordingly. So it would "work" now, but with a double nested team name when browsing that particular team.

lieut-data:
@jasonblais, re: install guides, I'm inclined to keep them on the "happy path" and reserve the subpath for more advanced configuration. Does that seem reasonable?

source/administration/config-settings.rst
jasonblais: @lieut-data I see - I'm wondering if we want to keep a note about not appending the team name?

Otherwise admins running a single team might attempt to enter `https://example.com/company/mattermost/teamname` which subsequently fails.

jasonblais:
@lieut-data Yeah that's reasonable to me. Probably better to keep the install guide as simple as possible.

source/administration/config-settings.rst
lieut-data: I suppose we could keep the note, but I feel like the examples explaining what we mean would be confusing. Plus who's to say that they don't want `/company/mattermost/<overall teamname within company>/subteam1`, etc.?

If they did misconfigure it, it seems like it would be obvious how to fix it if they didn't like the URL, and it's only a restart away. Thoughts?

jasonblais: > So it would "work" now, but with a double nested team name when browsing that particular team.

So everything would work, it might just be "undesirable" which the admin can easily change?

lieut-data: Yes, precisely that :)

PR #2120, opened 2018-06-26
Update important-upgrade-notes

amyblais:
In response to https://github.com/mattermost/mattermost-server/issues/8945


jasonblais:
@cpanato What happens if they escape an incorrect character? Will it fail?

In that case we'd need to list all the special characters.

cpanato:
No, will work as well

PR #2126, opened 2018-06-28
Update downgrade instructions

esethna:
@grundleborg I moved your content to a new doc. Please help review and let me know if more steps from the upgrade doc should be copied over here? Goal being to make downgrading as easy to follow as the upgrade guide (https://docs.mattermost.com/administration/upgrade.html). We should have step by step instructions

### DIFF
--- a/source/administration/downgrade.rst
+++ b/source/administration/downgrade.rst
@@ -1,0 +1,86 @@
+Downgrading Mattermost Server
+=============================
+
+In most cases you can downgrade Mattermost Server using the same steps as :doc:`upgrade`. The binaries can be found in the :doc:`version-archive`
+
+Downgrade from 5.0 to 4.10
+---------------------------
+
+During the version 5.0 upgrade process, a migration is run to enable new features of the `advanced permissions system <https://docs.mattermost.com/deployment/advanced-permissions.html>`_. This changes the database in ways that result in it no longer being compatible with Mattermost server 4.10. If you need to downgrade from 5.0 or later to 4.10, it is necessary to follow the steps below.
+
+You should gather the following information before starting the upgrade:
+
+Existing install directory - *{install-path}*
+  If you don't know where Mattermost Server is installed, use the ``whereis mattermost`` command. The output should be similar to */opt/mattermost/bin/mattermost*. The install directory is everything before the last occurrence of the string */mattermost*. In this example, the *{install-path}* is ``/opt``.
+Location of your local storage directory
+  The local storage directory contains all the files that users have attached to their messages. If you don't know its location, open the System Console and go to **Files > Storage** and read the value in **Local Storage Directory**. Relative paths are relative to the ``mattermost`` directory. For example, if the local storage directory is ``./data/`` then the absolute path is ``{install-path}/mattermost/data``.
+Owner and group of the install directory - *{owner}* and *{group}*
+  Use the ``ls -l {install-path}/mattermost/bin/mattermost`` command to get the owner and group.
+
+#. In a terminal window on the server that hosts Mattermost Server, change to your home directory. If any, delete files and directories that might still exist from a previous download.
+
+   .. code-block:: sh
+
+     cd ~
+
+#. Stop Mattermost Server.
+
+   On Ubuntu 14.04 and RHEL 6.6:
+
+   .. code-block:: sh
+
+     sudo service mattermost stop
+
+   On Ubuntu 16.04 and RHEL 7.1:
+
+   .. code-block:: sh
+
+     sudo systemctl stop mattermost
+     
+
+     
+#. Back up your data and application.
+
+   #. Back up your database using your organization’s standard procedures for backing up MySQL or PostgreSQL.
+
+   #. Back up your application by copying into an archive folder (e.g. ``mattermost-back-YYYY-MM-DD-HH-mm``).
+
+    cd {install-path}/mattermost
+    sudo setcap cap_net_bind_service=+ep ./bin/mattermost
+
+        cd {install-path}
+        sudo cp -ra mattermost/ mattermost-back-$(date +'%F-%H-%M')/
+        
+        
+#. Run the following SQL statements to revert the changes to the database made by the migration. XXXXXX GG Are there commands that have to be run before entering these SQL statements?
+
+```
+UPDATE Teams SET SchemeId = NULL;
+UPDATE Channels SET SchemeId = NULL;
+
+UPDATE TeamMembers SET Roles = CONCAT(Roles, ' team_user'), SchemeUser = NULL where SchemeUser = 1;
+UPDATE TeamMembers SET Roles = CONCAT(Roles, ' team_admin'), SchemeAdmin = NULL where SchemeAdmin = 1;
+
+UPDATE ChannelMembers SET Roles = CONCAT(Roles, ' channel_user'), SchemeUser = NULL where SchemeUser = 1;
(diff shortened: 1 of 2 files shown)


source/administration/downgrade.rst
```diff
@@ -0,0 +1,86 @@
...
+   #. Back up your application by copying into an archive folder (e.g. ``mattermost-back-YYYY-MM-DD-HH-mm``).
+
+    cd {install-path}/mattermost
+    sudo setcap cap_net_bind_service=+ep ./bin/mattermost
+
+        cd {install-path}
+        sudo cp -ra mattermost/ mattermost-back-$(date +'%F-%H-%M')/
+        
+        
+#. Run the following SQL statements to revert the changes to the database made by the migration. XXXXXX GG Are there commands that have to be run before entering these SQL statements?
```
grundleborg: well you need to be connected to your database CLI. Not sure if that's just assumed basic sysadmin knowledge though, since the way people do that is different for different databases and different deployment methods and different preferred dbadmin tools.

source/administration/downgrade.rst
```diff
@@ -0,0 +1,86 @@
...
+
+   On Ubuntu 16.04 and RHEL 7.1:
+
+   .. code-block:: sh
+
+     sudo systemctl start mattermost
+     
+     
+
+XXXXXX GG any more steps to be added or are applicable based on the upgrade guide? https://docs.mattermost.com/administration/upgrade.html
```
grundleborg: I don't think so. Maybe you could note that those SQL commands could take a while to run on large installations. Don't ask me for specific run times - just could conceivably be minutes rather than milliseconds.

esethna:
@cpanato I believe you tested the downgrade script (5.0 > 4.10) mentioned in this document? Can you please help verify the steps or add any that are missing?

source/administration/downgrade.rst
```diff
@@ -0,0 +1,83 @@
...
+
+     
+#. Back up your data and application.
+
+   #. Back up your database using your organization’s standard procedures for backing up MySQL or PostgreSQL.
+
+   #. Back up your application by copying into an archive folder (e.g. ``mattermost-back-YYYY-MM-DD-HH-mm``).
+
+    cd {install-path}/mattermost
+    sudo setcap cap_net_bind_service=+ep ./bin/mattermost
```
cpanato: Is this step really need in this stage? also the formatting is a bit strange, it is using a quote

source/administration/downgrade.rst
```diff
@@ -0,0 +1,83 @@
...
+
+        cd {install-path}
+        sudo cp -ra mattermost/ mattermost-back-$(date +'%F-%H-%M')/
+        
+        
+#. Connect to your database CLI and run the following SQL statements to revert the changes to the database made by the migration. The commands may take up to a few minutes to run on large installations.
+
+```
```
cpanato: the format here is weird as well , please review

source/administration/downgrade.rst
```diff
@@ -0,0 +1,83 @@
...
+UPDATE Teams SET SchemeId = NULL;
+UPDATE Channels SET SchemeId = NULL;
+
+UPDATE TeamMembers SET Roles = CONCAT(Roles, ' team_user'), SchemeUser = NULL where SchemeUser = 1;
```
cpanato: extra space in `' team_user'`

Got this error:
```
UPDATE TeamMembers SET Roles = CONCAT(Roles, 'team_user'), SchemeUser = NULL where SchemeUser = 1;
ERROR:  operator does not exist: boolean = integer
LINE 1: ...Roles, 'team_user'), SchemeUser = NULL where SchemeUser = 1;
```

grundleborg: That extra space is not a mistake - it's essential or else Mattermost will break.

grundleborg: Are you testing on postgres or mysql? I wonder if the treatment of booleans is different betwene them (perhaps `SchemeUser = 1` should be `SchemeUser = true` on Postgres?)

cpanato: postgres. if have specific command for each DB we need make a note.

cpanato: Ok, did the downgrade. the permission scheme still appears in the system console.

esethna: @cpanato in the about mattermost dialog does it show 4.10? @grundleborg thoughts?

source/administration/downgrade.rst
esethna: Fixed

source/administration/downgrade.rst
```diff
@@ -0,0 +1,85 @@
...
+
+      UPDATE Teams SET SchemeId = NULL;
+      UPDATE Channels SET SchemeId = NULL;
+
+      UPDATE TeamMembers SET Roles = CONCAT(Roles, ' team_user'), SchemeUser = NULL where SchemeUser = 1;
+      UPDATE TeamMembers SET Roles = CONCAT(Roles, ' team_admin'), SchemeAdmin = NULL where SchemeAdmin = 1;
+
+      UPDATE ChannelMembers SET Roles = CONCAT(Roles, ' channel_user'), SchemeUser = NULL where SchemeUser = 1;
+      UPDATE ChannelMembers SET Roles = CONCAT(Roles, ' channel_admin'), SchemeAdmin = NULL where SchemeAdmin = 1;
+
```
grundleborg: OK, have tested the `true/false` version of this and it works for me on MySQL *and* Postgres, so please replace these four lines with:

```

UPDATE TeamMembers SET Roles = CONCAT(Roles, ' team_user'), SchemeUser = NULL where SchemeUser = true;
UPDATE TeamMembers SET Roles = CONCAT(Roles, ' team_admin'), SchemeAdmin = NULL where SchemeAdmin = true;

UPDATE ChannelMembers SET Roles = CONCAT(Roles, ' channel_user'), SchemeUser = NULL where SchemeUser = true;
UPDATE ChannelMembers SET Roles = CONCAT(Roles, ' channel_admin'), SchemeAdmin = NULL where SchemeAdmin = true;
```

source/administration/downgrade.rst
```diff
@@ -0,0 +1,85 @@
...
+      DELETE from Systems WHERE Name = 'migration_advanced_permissions_phase_2';
```
grundleborg: Also, *before* these lines, please add:

```
UPDATE TeamMembers SET SchemeUser = NULL, SchemeAdmin = NULL;
UPDATE ChannelMembers SET SchemeUser = NULL, SchemeAdmin = NULL;
```

source/administration/downgrade.rst
```diff
@@ -0,0 +1,85 @@
...
+#. Back up your data and application.
+
+   #. Back up your database using your organization’s standard procedures for backing up MySQL or PostgreSQL.
+
+   #. Back up your application by copying into an archive folder (e.g. ``mattermost-back-YYYY-MM-DD-HH-mm``).
+
+   .. code-block:: sh
+    
+      cd {install-path}/mattermost
+      sudo setcap cap_net_bind_service=+ep ./bin/mattermost
```
grundleborg: I don't know what this line does, so I have no idea why this is included.

source/administration/downgrade.rst
grundleborg: If you still see a *team override scheme* in the system console then you haven't downgraded. That doesn't exist in 4.10. The system scheme configuration exists there though, so it is expected to see that still.

esethna:
Thanks for the ping. @cpanato can you please retest this script after the updates from George?

amyblais:
Is this good to merge?

PR #2131, opened 2018-07-03
Update config-settings.rst for v5.1

jasonblais:
Proposed dev reviewers:

- @lieut-data on `ConnMaxLifetimeMilliseconds` doc
- @hmhealey on `ExperimentalLimitClientConfig` doc

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -2812,7 +2812,7 @@ The number of seconds to wait for a response from the database after opening a c
 
 Maximum Connection Lifetime
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Maximum lifetime for a connection to the database in milliseconds. Use this setting to configure a maximum connection timeout for your database. Defaults to an hour (3,600,000 milliseconds).
+Maximum lifetime for a connection to the database, in milliseconds. Use this setting to configure the maximum amount of time a connection to the database may be reused. Defaults to an hour (3,600,000 milliseconds).
 
 +-------------------------------------------------------------------------------------------------------------------------+
 | This feature's ``config.json`` setting is ``"ConnMaxLifetimeMilliseconds": 3600000`` with whole number input.           |
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -3745,7 +3745,7 @@ Limit Access to Config Settings Prior to Login
 
 Enable this setting to limit the number of config settings sent to users prior to login.
 
-Supported for Mattermost server v5.1.0 and later, and Mattermost Mobile apps v1.9.0 and later.
+Supported for Mattermost server v5.1.0 and later, and Mattermost Mobile apps v1.10.0 and later.
 
 +-------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | This feature’s ``config.json`` setting is ``"ExperimentalLimitClientConfig": "false"`` with options ``true`` and ``false`` for above settings respectively.       |
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -3740,12 +3740,10 @@ Changes made when hardened mode is enabled:
 | This feature’s ``config.json`` setting is ``"ExperimentalEnableHardenedMode": false`` with options ``true`` and ``false`` for above settings respectively.          |
 +---------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
-Limit Access to Client Config Settings Prior to Login
+Limit Access to Config Settings Prior to Login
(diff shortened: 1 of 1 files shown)


source/administration/config-settings.rst
```diff
@@ -2807,7 +2807,15 @@ SQL Query Timeout
 The number of seconds to wait for a response from the database after opening a connection and sending the query. Errors that you see in the UI or in the logs as a result of a query timeout can vary depending on the type of query.
 
 +-------------------------------------------------------------------------------------------------------------------------+
-| This feature's ``config.json`` setting is ``"QueryTimeout": 30`` with whole number input, in the *SqlSettings* section. |
+| This feature's ``config.json`` setting is ``"QueryTimeout": 30`` with whole number input.                               |
++-------------------------------------------------------------------------------------------------------------------------+
+
+Maximum Connection Lifetime
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Maximum lifetime for a connection to the database in milliseconds. Use this setting to configure a maximum connection timeout for your database. Defaults to an hour (3,600,000 milliseconds).
```
lieut-data: I'd suggest:
> Maximum lifetime for a connection to the database, in milliseconds. Use this setting to configure the maximum amount of time a connection to the database may be reused. Defaults to an hour (3,600,000 milliseconds).

jasonblais:
Thanks for the review @lieut-data 👍

source/administration/config-settings.rst
```diff
@@ -3732,6 +3740,19 @@ Changes made when hardened mode is enabled:
 | This feature’s ``config.json`` setting is ``"ExperimentalEnableHardenedMode": false`` with options ``true`` and ``false`` for above settings respectively.          |
 +---------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
+Limit Access to Client Config Settings Prior to Login
```
hmhealey: Maybe drop `Client` from the title? We only really refer to it as the client config internally

source/administration/config-settings.rst
```diff
@@ -3732,6 +3740,19 @@ Changes made when hardened mode is enabled:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Enable this setting to limit the number of client config.json settings prior to login.
+
+The complete client config is pulled after logging in.
```
hmhealey: This can be left out. That just describes the implementation

source/administration/config-settings.rst
```diff
@@ -3732,6 +3740,19 @@ Changes made when hardened mode is enabled:
 | This feature’s ``config.json`` setting is ``"ExperimentalEnableHardenedMode": false`` with options ``true`` and ``false`` for above settings respectively.          |
 +---------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
+Limit Access to Client Config Settings Prior to Login
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Enable this setting to limit the number of client config.json settings prior to login.
```
hmhealey: Maybe `Enable this setting to limit the number of config settings sent to users prior to login.`? That might

jasonblais:
@hmhealey Updated. Also removed reference to `client config` and kept it as `config`. Note that the config.json setting name itself refers to it as `client config`.

hmhealey:
That's fine that it still refers to it as the client config since that's the internal name for it throughout the code and it refers to the part of the config sent to the client, but I don't think people reading the docs need to understand that

source/administration/config-settings.rst
```diff
@@ -3732,6 +3740,17 @@ Changes made when hardened mode is enabled:
...
 
+Limit Access to Config Settings Prior to Login
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Enable this setting to limit the number of config settings sent to users prior to login.
+
+Supported for Mattermost server v5.1.0 and later, and Mattermost Mobile apps v1.9.0 and later.
+
++-------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| This feature’s ``config.json`` setting is ``"ExperimentalLimitClientConfig": "false"`` with options ``true`` and ``false`` for above settings respectively.       |
```
esethna: Just curious why this is experimental and set to false by default?

jasonblais: @esethna Because this is a breaking change for older apps, this setting is disabled by default with an experimental setting to turn it on for testing. Once we get to a point where we drop support for the 1.9 mobile apps, we can remove the setting and enable it for everyone.

PR #2132, opened 2018-07-04
Community Overview Update

it33:


it33:
@lfbrock @jasonblais me again, re-wrote the Community Overview as what was posted was pretty tricky--when I read it, it sounded like we only farm out small, unimportant stuff to community, which is not accurate (e.g. Gyfcat, Docker, Desktop app, etc.). 

I've re-written it to frame community as the overarching priority and Mattermost staff as paid members of community, customers as community as well able to accelerate features, but not dictate our destiny. 

Feel free to edit and share feedback. 

I think the page was going in somewhat the right direction, but it needed acceleration.

jasonblais:
@it33 Thanks! At a quick glance the edits are definitely an improvement.

To increase discoverability, we should also consider moving or referencing this page elsewhere, given it only receives 1 or 2 views per day.

Have made a follow-up task to review more closely later. Will also review with Joram who might have feedback on the code contributor levels.

PR #2143, opened 2018-07-10
small fix mattermost user migrate_auth command

burguyd:


amyblais:
@burguyd Thank you for the PR!

PR #2171, opened 2018-07-26
Patch 1 / Scaling doc

JustinReynolds-MM:
Updated Scaling to include Sample Server Scaling guide.


source/deployment/scaling.rst
```diff
@@ -30,3 +30,33 @@ Cluster-based Deployment
...
+
+As deployments scale above 5,000 users, additional servers are added for performance load-balancing and to provide additional redundancy (see our `High Availability Cluster guide <https://docs.mattermost.com/deployment/cluster.html#mattermost-server-configuration>`_).
+
+`This spreadsheet <https://docs.google.com/spreadsheets/d/14rXqhLa2VeBFFQC2CISxMppVMTT_DGrHEV6DHEEXU4A/>`_ illustrates how much hardware you’ll need for different-sized Mattermost deployments, as well as estimates as to how much you can expect to pay. It includes links to AWS’s cost calculator for various deployment sizes. The smaller deployment examples (i.e., 1,000 users and 5,000 users) are on the conservative side, with separate servers per function that can easily be scaled out as Mattermost is rolled out.
```
lfbrock: I think this google doc link is not public

lfbrock: Minor, but can we try and simplify wording when possible? (Our documentation guideline is to write for an international audience). 

For example, I think "illustrates", "estimates as to", could maybe be changed to use simpler language.

PR #2180, opened 2018-07-30
Updating software requirements

jasonblais:
This PR updates software requirements, and adds proposed guidelines on minimum browser support.

The proposed requirements are attached below. For more context on how these were arrived at, see the new `../process/software-requirements.rst` file and discussion at https://pre-release.mattermost.com/core/pl/sb4fq6qhyfbb5xjdp7x3ud146e

image

Software requirements are also updated on webapp error page when an incompatible browser is used: https://github.com/mattermost/mattermost-webapp/pull/1946

This PR also adds a step into the release process to review software requirements are up-to-date based on the guidelines, and if not, update documentation accordingly, and note changes in the Changelog.

Finally, a ticket has been prepared to remove mobile view support on IE11: https://mattermost.atlassian.net/browse/MM-12825

### DIFF
--- a/source/process/bug-fix-release.md
+++ b/source/process/bug-fix-release.md
@@ -16,6 +16,7 @@ Pre-work for the current release begins at the code complete date of the previou
     - Post this checklist in Release Checklist channel
     - Queue a list of MVP candidates in alphabetical order to the Platform Meeting channel. [See example](https://pre-release.mattermost.com/private-core/pl/q9jdbzw7c7ribjsp78857xbomh)
     - Draft Changelog in a WIP PR with updates for known issues, compatibility updates for deprecated features, config.json, [database changes](https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/upgrade.go), [API changes](https://github.com/mattermost/mattermost-server/commits/master/model/client.go), and [WebSocket event changes](https://github.com/mattermost/mattermost-server/blob/master/model/websocket_message.go#L13); [see example](http://docs.mattermost.com/administration/changelog.html#compatibility)
+    - Review [software requirements](https://docs.mattermost.com/install/requirements.html#software-requirements) are up-to-date based on [these guidelines](https://docs.mattermost.com/process/software-requirements.html). If not, update documentation accordingly, and note changes in the Changelog
     - Update [Upgrade Guide](https://docs.mattermost.com/administration/upgrade.html#upgrade-guide) with any special notes for upgrading to the new version
     - Review submitted NOTICE.txt PRs for any new libraries added from dev, and ensure they are cherry-picked to feature release branch.
     - Start posting a daily Zero Bug Balance query (posted until zero bugs or day of release)


source/install/requirements.rst
```diff
@@ -26,12 +26,21 @@ Client Software
 PC Web Experience
 ^^^^^^^^^^^^^^^^^
 
--  PC: Windows 7, Windows 8, Windows 10 with IE 11*, Chrome 43+, Firefox 52+, and Edge 40+ (or EdgeHTML v15+)
--  Mac: OS 10 (Safari 9, Chrome 43+)
+-  PC: Windows 7+ with IE 11*, Chrome 43+, Firefox 52+, and Edge 40+ (or EdgeHTML v15+)
+-  Mac: OS 10.8+ (Safari 9, Chrome 43+)
 -  Linux: Arch 4.0.0 (Chrome 43+)
```
crspeller: I know i have mentioned this before. But Arch doesn't really have versions. We can just drop Arch and just say Linux: (Chrome 59+ Firefox 60+)

source/install/requirements.rst
```diff
@@ -26,12 +26,21 @@ Client Software
 PC Web Experience
 ^^^^^^^^^^^^^^^^^
 
--  PC: Windows 7, Windows 8, Windows 10 with IE 11*, Chrome 43+, Firefox 52+, and Edge 40+ (or EdgeHTML v15+)
--  Mac: OS 10 (Safari 9, Chrome 43+)
+-  PC: Windows 7+ with IE 11*, Chrome 43+, Firefox 52+, and Edge 40+ (or EdgeHTML v15+)
```
crspeller: We should update this to versions we actually know we support. So Chrome 59+ (which is the version electron is on). As well as Firefox 60+ (Current Firefox ESR).

crspeller: Not sure what we mean by edge 40+ because that's not a version of edge.  Looking at the caniuse. I think the earist version of Edge we can support is 16. So 16+ for edge. Same for EdgeHTML we support 16+.

source/install/requirements.rst
jasonblais: I can drop Arch, no problem

source/install/requirements.rst
jasonblais: 40+ was based off of https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history. 

Re: EdgeHTML/Chrome/Firefox versions: While what you say makes sense (we shouldn't say we support Chrome v43 if we have no idea if Mattermost still works with that version), I think there should be a broader process via QA to validate we still support version XX, and if not, make the call on dropping support for it or fixing the issue to support version XX

jasonblais: And if we drop support for older versions, we should release note it somewhere.

jasonblais: (Will post in QA to get their thoughts)

crspeller: Ah sorry I read stuff wrong. All the stuff I was looking refers to edge by the EdgeHTML version. So we support Edge 41+ (which is EdgeHTML 16+).

jasonblais: @crspeller The more time I spend looking into this, the more I feel it’s not critical enough to go through the process of updating these version numbers right now.

1 - We have a UI page that asks people to updated their browsers if they’re on an old version we no longer support.

That means a change to that page, and blocking some % of users accessing Mattermost.

image

2 - We’d need to release note it so customers know they must upgrade their users’ browsers if they’re not on one of the supported versions (or else their users will be blocked from Mattermost).

This caused a lot of trouble for at least one customer last time we did it for Firefox and decided to drop v45 ESR support, because we legitimately could no longer support it.

A lot of our customer base is in government organizations and financial institutions which are notorious for running old browser versions.

3 - It’s not clear what benefit we get out of this change. No one is hitting issues with it at the moment.

4 - It’s not clear what impact the changes would have without further investigation (e.g. I looked into Firefox usage and v52 ESR is used more than v60 ESR, at least by Mattermost users)

5 - Competitors have lenient browser version support, see Slack’s attached 

image

----

There is a separate conversation we've had with QA about this, and we're looking at options to more easily automate UI testing across browsers and versions, whether with NightWatchJS or something else.

But unless you feel really strongly about this, I don't think it's worth updating this right now.

crspeller: 1. We don't have to hard block, we can just update the numbers here. 
2. We can release note it sure. 
3. Actually having an accurate description of what we support? There isn't much point in listing support for browser versions we don't actually support. Also having a definite and updated browser support versions will allow us to improve performance by using modern browser features. 
4. It's not a hard block. So people running old versions should still work, but FF doesn't support their old ESR anymore so there is little reason for us to.
5. They have more resources. Also their support is not much more lenient then ours.

jasonblais: @crspeller I didn't realize Firefox v52ESR was no longer supported as of Sept 5th. That makes it easier to push people to v60+.

Alright, so we would change version support for Chrome (v59+), Firefox (v60+) and Edge (v41+).
 - Have touched base with a customer who might be affected and see if they'd be open to upgrading their browser versions given FFv52 is considered EOL.
 - I'll also see if we can drop IE11 mobile view support at the same time, since it's buggy and likely not used. Have touched with a customer who might be affected on that.

PS: If it's not a hard block, then we'd have people running on old unsupported browsers with potentially a poor user experience. That said, we indirectly do that already so that's less of a concern.

jasonblais: Customers I talked to are okay with these changes, next I've shared a proposal with QA: https://pre-release.mattermost.com/core/pl/sb4fq6qhyfbb5xjdp7x3ud146e

jasonblais:
I have updated PR description with the changes.

@amyblais @esethna @lindalumitchell  Please help review and let me know if there are any questions.

source/process/software-requirements.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Current software requirements are `documented here <https://docs.mattermost.com/install/requirements.html#software-requirements>`_.
+
+Desktop Apps
+---------------------------------
+
+.. csv-table::
+    :header: "Operating System", "Guideline"
+
+    "Windows", "Supported versions by Microsoft - `reference <https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions>`_"
```
esethna: Wikipedia is probably not the most official source for this info, is there a page on the windows or apple sites with this info?

source/process/software-requirements.rst
```diff
@@ -0,0 +1,38 @@
...
+
+PC Web
+---------------------------------
+
+.. csv-table::
+    :header: "Browser", "Guideline"
+
+    "Chrome", "Chromium version of latest Mattermost Desktop App"
+    "Firefox", "Supported versions by Mozilla - `reference <https://www.mozilla.org/en-US/firefox/organizations/>`_"
+    "Safari", "Supported versions by Apple - `reference <https://en.wikipedia.org/wiki/Safari_version_history>`_"
```
esethna: Same here

source/process/software-requirements.rst
```diff
@@ -0,0 +1,38 @@
...
+    "Internet Explorer", "Latest release"
+    "Edge", "Latest release"
+    
+Mobile Apps
+---------------------------------
+
+.. csv-table::
+    :header: "Operating System", "Guideline"
+
+    "iOS", "Latest and next-to-latest versions - `reference <https://en.wikipedia.org/wiki/IOS_version_history>`_"
```
esethna: and here

jasonblais:
@esethna I didn't really find other good references directly in Microsoft, Apple or Google sites that would easily list supported versions.

They generally try to get people to latest versions and hence only document those.

amyblais:
@jasonblais "software-requirements.rst" needs to be added to table of contents.

jasonblais:
@amyblais PR here https://github.com/mattermost/docs/pull/2361

PR #2182, opened 2018-07-30
copy security.md from developer docs, with tweaks

lieut-data:
This lifts `security.md` from the old developer docs, with just a small change to remove the section about only system administrators being able to install custom plugins (since this is now more obvious in context).

Thoughts? The security language as written is pretty scary, but perhaps justifiably so...


jasonblais:
Reason I'm saying it is that it might reduce how many admins install plugins

PR #2194, opened 2018-08-03
Create migration-announcement-email-template.rst

wiersgallak:


jasonblais:
PS: There is one welcome email template here: https://docs.mattermost.com/getting-started/welcome_email.html. Not sure if we want to replace it?

PR #2219, opened 2018-08-15
MM-11255: Bring bulk import docs up to date.

grundleborg:
Brings docs for Bulk Importer up to date with what actually exists in 5.2.

### DIFF
--- a/source/deployment/bulk-loading-data-format.rst
+++ b/source/deployment/bulk-loading-data-format.rst
@@ -8,8 +8,12 @@ The input data file must be a valid `JSONL
 
 Version
   Mandatory. The Version object must be the first line in the file, and must occur only once.
+Scheme
+  Optional. If present, Scheme objects must occur after the Version object but before any Team objects.
+Emoji
+  Optional. If present, Emoji objects must occur after the Version objects but before any Team objects.
 Team
-  Optional. If present, Team objects must occur after the Version object and before any Channel objects.
+  Optional. If present, Team objects must occur after any Scheme objects and before any Channel objects.
 Channel
   Optional. If present, Channel objects must occur after all Team objects and before any User objects.
 User
@@ -29,6 +33,9 @@ The identifiers for each object are listed in the following table:
   :header: Object, Unique Identifier
 
   Version, Not Applicable
+  Scheme, *name*
+  Role, *name*
+  Emoji, *name*
   Team, *name*
   Channel, "*name*, *team*"
   User, *username*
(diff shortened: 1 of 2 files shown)


jasonblais:
FYI @cvitter ^

source/deployment/bulk-loading-data-format.rst
```diff
@@ -8,8 +8,12 @@ The input data file must be a valid `JSONL
 
 Version
   Mandatory. The Version object must be the first line in the file, and must occur only once.
+Scheme
+  Optional. If present, Scheme objects must occur after the Version object but before any Team objects.
+Emoji
+  Optional. If present, Emoji objects must occur after the Version objects but before any Team objects.
```
wiersgallak: Based on order of objects: 
- Does Scheme need to occur after Version object before Emoji objects?
- Does Emoji have to occur after the Scheme object prior to Team?

grundleborg: No to both, but if it makes it simpler to explain to users, we could just say it does in the docs. There's no real downside of forcing that strict ordering.

wiersgallak: I was just making sure there was not a hard requirement and an oversight possibly in copy/paste.  Enforcing the order in documentation may make it more consistent with the other items and avoid the question that I asked.

amyblais:
@grundleborg There are some formatting issues when previewing the doc - these don't render properly:

- Example Scheme object
- Example Emoji object
- Fields of the Reply object
- Fields of the Attachment object

source/deployment/bulk-loading-data-format.rst
```diff
@@ -115,6 +123,212 @@ Fields of the Version object
...
+      <td valign="middle">name</td>
+      <td valign="middle">string</td>
+      <td>The scheme name.</td>
+      <td align="center" valign="middle">Yes</td>
+      <td align="center" valign="middle">Yes</td>
+    </tr>
+    <tr class="row-odd">
+      <td valign="middle">image</td>
+      <td valign="middle">string</td>
+      <td>The path (either absolute or relative to the ``mattermost`` binary) to the image file for this emoji.</td>
```
jespino: I think it is relative to the current directory, not to the mattermost binary.

source/deployment/bulk-loading-data-format.rst
```diff
@@ -893,6 +1136,27 @@ This object is a member of the Post/DirectPost object.
       <td align="center" valign="middle">Yes</td>
       <td align="center" valign="middle">Yes</td>
     </tr>
+tr class="row-odd">
```
jespino: Looks like here is wrong the html format.

amyblais:
@grundleborg These sections are still not rendering properly in the preview mode, not sure if I'm viewing it incorrectly?

- Example Scheme object
- Example Emoji object
- Fields of the Attachment object

grundleborg:
Fields of the attachment object appears to be rendering just fine when I look at the visual diff. As for the other two, I can't see anything wrong with them, so maybe Github doesn't know how to render those kind of blocks?

PR #2235, opened 2018-08-24
Add medium article written by Santos to CLI docs

jasonblais:


amyblais:
I know Santos is a community member but wondering if the word "excellent" should be used in docs? 0/5

jasonblais:
Good point, remove the word "excellent"

PR #2245, opened 2018-08-27
Update casestudy.rst

comharris:
Updated case study guidelines per Aug. 27th meeting with IJT, RT, and JR

### DIFF
--- a/source/process/casestudy.rst
+++ b/source/process/casestudy.rst
@@ -117,20 +117,20 @@ Written Case Study Lay Out
 
 Follow these steps to write a case study that is clear and concise:
 
-1. Title: Highlight the most compelling use case and/or business benefit from Results section. 
+1. Title: Highlight the most compelling use case and/or business benefit from Results section (see step #6 below). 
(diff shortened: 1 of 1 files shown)


source/process/casestudy.rst
```diff
@@ -106,20 +106,31 @@ Case Study Interview Process
 Writing Guidelines 
 ----------------------------
 
-1. Write so that people with English as a second language can understand every sentence. 
-2. Don't disparage other products. It's okay to factually address differences on why Mattermost was selected. 
+1. DO NOT use slang or jargon. Write so that people with English as a second language can understand every sentence. 
+2. DO NOT debase the competition or disparage other products. It's okay to factually address differences on why Mattermost was selected.
```
amyblais: Wondering if a simpler word can be used than "debase"?

comharris: How about: "DO NOT criticize the competition or disparage other products..."

Would that work?

amyblais: Sure thing! Made the edit

PR #2267, opened 2018-09-06
WIP: Create integration_contribution_guidelines

wiersgallak:


amyblais:
I made lot of initial grammar and readability edits. Also, a note for future that we don't include exclamation marks or smiley faces, etc. in docs.

Also, where will the new "integration contribution guidelines" doc link from in the documentation?

wiersgallak:
Thanks Amy.  Maybe docs is not the right location for this content.  We were trying to convey a very encouraging message.  @jasonblais Do you think we should continue on docs or do you think we should see about adding it to the website instead?

jasonblais:
Hmm, I didn't think about that. Good points @amyblais 

An alternative place could be the developer docs which are less formal than docs.mattermost.com: https://developers.mattermost.com/integrate/getting-started/

wiersgallak:
I will work with Rich to get it added to the developers site.

jasonblais:
@wiersgallak Actually, the site is hosted here, so opening a GitHub PR works :) 

https://github.com/mattermost/mattermost-developer-documentation/tree/master/site

PR #2272, opened 2018-09-10
Add minimum and target display size to UX guidelines

hmhealey:
As discussed here: https://pre-release.mattermost.com/core/pl/xns8xbfabtf6pg5tqdk498wzzr

@asaadmahmood I also added 1920x1080 as a target screen size for desktop since it's very common, and I'm not sure what size our designs are made against

### DIFF
--- a/source/developer/fx-guidelines.rst
+++ b/source/developer/fx-guidelines.rst
@@ -79,7 +79,7 @@ Screen and Content Dimensions
 Web/Desktop App
 -------------------
 
-The webapp supports two main modes of displaying the app: desktop view and mobile view. The app is primarily used in desktop view, but when the screen is 768 pixels or less wide, it switches to mobile view. The mobile view is also used in mobile browsers and the Classic App.
+The webapp supports two main modes of displaying the app: desktop view and mobile/tablet view. The app is primarily used in desktop view, but when the screen is 768 pixels or less wide, it switches to mobile view. The mobile view is also used in mobile browsers and the Classic App.
 
 For desktop view:
 
--- a/source/developer/fx-guidelines.rst
+++ b/source/developer/fx-guidelines.rst
@@ -77,22 +77,25 @@ Screen and Content Dimensions
 ****************************
 
 Web/Desktop App
----------------
+-------------------
 
-The web app supports two main modes of displaying the app: desktop view and mobile view. The primary way the app is used is in desktop view, but when the screen is 768 pixels or less wide, it switches to mobile view. This mobile view is also used by mobile browsers and the Classic App.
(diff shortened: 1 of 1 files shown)


source/developer/fx-guidelines.rst
```diff
@@ -73,15 +73,29 @@ We’re using `BEM methodology <https://en.bem.info/method/>`_  and would advise
...
+Screen and Content Dimensions
 ****************************
 
-Desktop
--------
+Web/Desktop App
+-------------------
 
-- **Minimum screen resolution:** 1280 x 800
+The webapp supports two main modes of displaying the app: desktop view and mobile view. The app is primarily used in desktop view, but when the screen is 768 pixels or less wide, it switches to mobile view. The mobile view is also used in mobile browsers and the Classic App.
```
esethna: What about tablet view?

source/developer/fx-guidelines.rst
```diff
@@ -73,15 +73,29 @@ We’re using `BEM methodology <https://en.bem.info/method/>`_  and would advise
...
 
-- **Color Depth:** 16.7M colors
+For desktop view:
+
+ - The target screen sizes are 1280 x 800 or 1920 x 1080. Note that the space available for content will be smaller than this due to the the frame of the window and system-specific components like the taskbar on Windows. For example, the desktop app on Windows 10 only has a usable height of 717 pixels when the tab bar is visible at 1280 x 800 resolution.
+ - The minimum content size is 769 x 600. When the width is 768 or less, the app switches to mobile view. When the height is below 600, UI elements may become misplaced.
```
esethna: So it's the width that determines the view right? Since we say "When the height is below 600, UI elements may become misplaced." but the condition would be that the width is also above 768? Can we clarify?

source/developer/fx-guidelines.rst
hmhealey: I don't think explicit tablet view actually exists any more. I tried resizing the window to trigger it, but I could never get the mobile send button to appear like it used to

hmhealey: To add to that, there are some widths that do trigger some different CSS behaviours, but they aren't well defined. Some stuff changes at a width of 320px, 380px, 480px, 550px, 640px, 768px, 960px, 1020px, 1024px, 1140px, 1280px, 1440px, and 1800px. I assume most of those are just slight optimizations, but I'm not sure. Asaad would know more about that

esethna:
Okay thanks Harrison, I'll wait for @asaadmahmood to review this first

source/developer/fx-guidelines.rst
asaadmahmood: I think this sounds fine. We do not have a specific tablet view, we can say. 
`The webapp supports two main modes of displaying the app: desktop view and tablet/mobile view.`

source/developer/fx-guidelines.rst
asaadmahmood: `The minimum content size is 769 x 600`.
We should change that to `768`.

PR #2277, opened 2018-09-11
Update release-process.md

jasonblais:
1. Move ownership of running loadtests from dev ops to performance team. Asking release manager's help to confirm this from the performance team, currently `joram, christopher, jesse, jason`
2. Move marketing reviews one step earlier, so Ian gets to review the blog 6 working days before the final rather than 2. I think he proposed earlier to review it one week beforehand, but having a T-5 for that item alone isn't necessary.

^feedback welcome on the above @amyblais. First one came up during ABC sprint planning


jasonblais:
I made the updates for 2) simpler

PR #2282, opened 2018-09-14
v5.4 Changelog

amyblais:


jasonblais:
@amyblais Did we remove a lot of the UX/UI improvements from the changelog? I think they're still worth keeping, just re-organized like Katie mentioned.

PR #2318, opened 2018-10-03
Edit Others Posts Permission

wiersgallak:
Please let me know if there is a better location for the documentation of this feature, as it TE only and in the UI, not config.json


wiersgallak:
Thanks @jasonblais found the reference and updated formatting accordingly.

PR #2366, opened 2018-11-06
Update config-proxy-apache2.rst

tuxfamily:
Replaced default web socket URL.
Mattermost does not use TLS per default, so the default value should be ws:// and not wss://, and a note should be added about this.


tuxfamily:
Sorry, did not notice this requirement. I filled the form, I'm now in the contributors list. Thanks.

tuxfamily:
After diging a little, it is possible to use wss:// but this requires that Mattermost uses TLS, so, per default, it should be ws:// and a note should be added about this.

amyblais:
No worries @tuxfamily, thank you for signing the form!

If you would like to make further edits to the doc to add a note about your findings that you mentioned about in your latest comment, you can follow these steps:

1. Click "Files Changed" (you'll find it near the title of this pull request).
2. On the right-hand side, click the pencil icon to make edits.

Let me know any questions.

amyblais:
Thank you @tuxfamily!

(The only edit I made was to add a period at the end of the sentence on step 4)

PR #2387, opened 2018-11-19
Updating community definition

it33:


jasonblais:
I'll keep this open while we finalize the community plan. We may do additional adjustments.

jasonblais:
I feel the existing definition is better in describing community. The proposed one is more defensive about the subject which I feel isn't needed.

I think we do need a second iteration on this page, but that can come after we've finalized the V2MOM

PR #2389, opened 2018-11-20
Update changelog.md

wiersgallak:


wiersgallak:
@jasonblais makes sense.  I have updated the text.

PR #2393, opened 2018-11-22
Newer SSL Configuration for nginx

JtheBAB:
This configuration gives a better rating at ssllabs.com

### DIFF
--- a/source/install/config-ssl-http2-nginx.rst
+++ b/source/install/config-ssl-http2-nginx.rst
@@ -73,14 +73,23 @@ You can use any certificate that you want, but these instructions show you how t
       listen 443 ssl http2;
       server_name    {domain-name} ;
 
+      #Oldest compatible clients: Firefox 27, Chrome 30, IE 11 on Windows 7, Edge, Opera 17, Safari 9, Android 5.0, and Java 8
+
       ssl on;
       ssl_certificate /etc/letsencrypt/live/{domain-name}/fullchain.pem;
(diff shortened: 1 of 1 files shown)


source/install/config-ssl-http2-nginx.rst
```diff
@@ -73,14 +73,23 @@ You can use any certificate that you want, but these instructions show you how t
       listen 443 ssl http2;
       server_name    {domain-name} ;
 
+      #Oldest compatible clients: Firefox 27, Chrome 30, IE 11 on Windows 7, Edge, Opera 17, Safari 9, Android 5.0, and Java 8
+
```
amyblais: Thank you @JtheBAB! One question is what are the compatible clients based on? We have a list of software requirements here, not sure if they should match: https://docs.mattermost.com/install/requirements.html#software-requirements.

JtheBAB: The problem is the ssl ciphers. The ciphers that i propose need at minimum this version of the browser or mobile operating system. But yes your requirements are higher so we can also just remove this comment.

amyblais: @JtheBAB Thanks for the note, you could maybe replace the clients + numbers with a link to the documentation I attached above. Just a suggestion, let me know any questions!

JtheBAB: @amyblais
As your requirements are much higher than the proposed ciphers i think it is ok to just remove the comment. I will do it tomorrow.

PR #2395, opened 2018-11-22
Interactive Message Dialogs

jasonblais:


source/developer/interactive-message-dialogs.rst
```diff
@@ -0,0 +1,272 @@
...
+          "state": "<string provided by the integration that will be echoed back with dialog submission>",
+      }
+  }
+
+Elements
+-----------------------
+
+Each dialog supports up to five elements for users to enter information. If your form needs more than five fields, such as for collecting survey responses, consider chaining dialogs together with buttons or menus to serialize workflows.
+
+// XXX Joram: I don't think we actually have this limit, do we? The sample dialog you prepared has more than 5.
```
jasonblais: Joram: This question is for you

source/developer/interactive-message-dialogs.rst
```diff
@@ -0,0 +1,272 @@
...
+-----------------------
+
+When a user submits a dialog, Mattermost will perform client-side input validation to make sure
+
+  - All required fields are filled
+  - All formats are correct (e.g. email, telephone number, etc.)
+
+Moreover, Mattermost also allows the integration itself to perform input validation.
+
+// XXX Joram can we have an example of this? Using a Go example is okay.
```
jasonblais: Joram: This question is for you

jwilander:
I'll take a look sometime early next week

source/developer/interactive-message-dialogs.rst
jwilander: I don't think a code example would be the most use here. A JSON response example like all the other sections would make most sense. For the test payload I gave you, I returned the following as the response body to the request to submit the dialog if the number wasn't 42:

```json
{"errors": {"somenumber": "Number must be 42!"}}
```

`somenumber` is the name of the number field in my test payload

I would also suggest moving the input validation to be discussed after we discuss what the submission payload looks like, since that will be the order any integration has to deal with them.

jasonblais:
Thank you @jwilander - PR updated.

source/developer/interactive-dialogs.rst
```diff
@@ -0,0 +1,403 @@
...
+      "submission": {
+          "some_element_name": "<value of that element>",
+          "some_other_element": "<value of some other element>"
+      },
+      “cancelled”: false
+  }
+
+Optionally, the dialog can send an event back to the integration if ``notify_on_cancel`` parameter is set to true. If this happens, ``cancelled`` will be set to true on the above payload, and ``submission`` will be empty.
+
+Moreover, Mattermost also allows the integration itself to perform input validation. For example, you can require the user to enter a number between 0 and 10, and return the following response body if the condition isn't satisfied:
```
jwilander:
```suggestion
Moreover, Mattermost also allows the integration itself to perform input validation. This can be done by responding to the dialog submission request with a JSON body containing an `errors` field. The `errors` field can contain a JSON object, mapping input field names to string error messages you would like to display to the user. For example, if you have a field named `num_between_0_and_10`, you can enforce the user to enter a number between 0 and 10 by returning the following response body if the condition isn't satisfied:
```

jasonblais:
Thanks! Updated

jasonblais:
All but screenshots are done so we can begin reviews

jasonblais:
@amyblais Screenshots added

amyblais:
@lfbrock For screenshot reviews

amyblais:
Added @asaadmahmood for screenshot reviews

PR #2422, opened 2018-12-10
Add Webhook delete command

wiersgallak:
Add webhook delete command: https://github.com/mattermost/mattermost-server/pull/9804 

Remove 'sudo' prior to newly added commands for 5.6 for formatting consistency when merged into master.

### DIFF
--- a/source/administration/command-line-tools.rst
+++ b/source/administration/command-line-tools.rst
@@ -384,7 +384,7 @@ mattermost command create
   Examples
     .. code-block:: none
 
-      sudo ./mattermost command create myteam --title MyCommand --description "My Command Description" --trigger-word mycommand --url http://localhost:8000/my-slash-handler --creator myusername --response-username my-bot-username --icon http://localhost:8000/my-slash-handler-bot-icon.png --autocomplete --post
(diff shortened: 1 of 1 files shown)


jespino:
I would prefer to cherry-pick the sudo removal from master to this branch, or leave it consistent, and solve the problem when it is merged. Looks weird that some commands needs sudo and others don't, actually, it give more "force" to the idea of the sudo is needed for that commands, which is not true.

wiersgallak:
After discussing with @jespino - I have taken a fresh copy of what is currently in master (with the updates to remove ```sudo``` as appropriate). I have updated with the commands that were previously approved and merged to 5.6 docs and added the webhook delete command.  

Hoping this ensures we have the cleanest version with the least conflicts with changes currently reflecting in Master.

source/administration/command-line-tools.rst
```diff
@@ -1611,8 +1612,24 @@ mattermost webhook create-outgoing
           --icon [iconURL]                        Icon URL
           --url stringArray [REQUIRED]            Callback URLs 
           --content-type string                   Content-type
-	  --h, --help				  Help for create-outgoing
-	  
+    --h, --help         Help for create-outgoing
```
jespino: Here you are channging the identation, is expected to be at the same level than the previous one in both columns

wiersgallak: Fixed - I must have accidentally hit delete.

PR #2423, opened 2018-12-10
Add an important upgrade note for public channel materialization

jasonblais:


source/administration/important-upgrade-notes.rst
```diff
@@ -9,6 +9,28 @@ Important Upgrade Notes
...
+|                                                    |       Channels c                                                                                                                                                |
+|                                                    |   WHERE                                                                                                                                                         |
+|                                                    |       c.Type = 'O';                                                                                                                                             |
+|                                                    |                                                                                                                                                                 |
+|                                                    | The queries above rebuild the materialized ``PublicChannels`` table without modifying the authoritative ``Channels`` table.                                     |
```
lieut-data:
```suggestion
|                                                    | The queries above rebuild the materialized ``PublicChannels`` table without modifying the authoritative ``Channels`` table.  
|                                                    | Note that this migration is not required if the experimental ``PublicChannels`` feature was never disabled. This feature launched in v5.4 with a temporary flag to disable should an issue arise, but nothing prompted doing so. If you did not modify this setting, there is no need to perform this migration.

jasonblais:
Thanks! @lieut-data Updated

PR #2425, opened 2018-12-11
Fix note about username override

nivex:
In the note about integration overrides, the behavior of the icon override is correct, but the username override behavior is incorrect.


amyblais:
Hi @nivex, thank you for your contribution,

Would you like to sign the CLA above and then we can proceed with reviewing and merging the PR? Let me know any questions!

PR #2438, opened 2018-12-16
Fix release date

kaakaa:
Released on 12/14?


jasonblais:
@kaakaa It was released on Dec 14th because of the weekend, but we keep the official release date as the 16th so it's easier for everyone to remember when Mattermost is released (on the 16th of every month)

PR #2445, opened 2018-12-19
Add an ADFS script to ./mattermost user migrate_auth email saml command

jasonblais:
credit to @jespino


jespino:
@jasonblais I think so. This code was executed by @mkraft against a real AD server, so... at least, it works.

jasonblais:
@jespino Thanks! Any further dev review needed, or this should be okay for PM/editor review only?

PR #2455, opened 2019-01-02
leverage aws s3 sync --delete to avoid downtime

lieut-data:
Effectively inline the `--delete` operation by not using Travis CI's `s3` deploy, but `aws s3 sync --delete` directly.

I don't know how to test this except by doing it live: suggestions?


jasonblais:
Adding `Merge On Hold` label - Jesse will merge later tonight Eastern time.

PR #2459, opened 2019-01-03
Linting fixes

lieut-data:
While perusing the Travis builds, I noticed a ton of linting warnings. Some are spurious, but others masked real problems like broken section links and such. This PR fixes a few of the more egregious problems, but there's still a ton we could do here. Ideally we get to a point where a warning fails the build instead of being ignored.

Some highlights:
* outbound links should use anonymous hyperlink references to avoid running into duplicate target names (e.g. multiple "Download" links)
* `^^^^^^` should be replaced with `~~~~~~~` for title level consistency
* underlines and overlines should be at least as long as the title itself: this may be enforced in the future


lieut-data:
I've broken things out into individual commits to help with review (hopefully!)

jasonblais:
Awesome! @amyblais please help ensure our documentation guidelines are aligned with the changes here.

source/developer/localization.rst
```diff
@@ -103,33 +103,33 @@ If you're interested in contributing to the process, please join the `Mattermost
 +------------------------------------------+-----------------------------------------------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------+
 | Language                                 | Official Reviewer(s)                                                                                                  | Maintainers                                                                                          |
 +==========================================+=======================================================================================================================+======================================================================================================+
-| Deutsch - German                         | `Christian Arnold (meilon) <https://github.com/meilon>`__                                                              | `Tim Estermann (der-test) <https://github.com/der-test>`__                                            |
```
jasonblais: Some of the linting fixes caused the tables to break, e.g. below. Curious why that might be?

image

lieut-data: Looks like a) I missed that, and b) it curiously didn't generate a warning itself. I've pushed up a change to address this, and scanned the other tables in a bid to convince myself I didn't make the same mistake elsewhere.

It's possible something else in the warning output was complaining about this, but it's still obscured by the other warnings.

PR #2469, opened 2019-01-13
shorten some windows paths within Troubleshooting Section

JtheBAB:
Now with a separate branch.

This patch shorten some paths within the Troubleshooting Section.


wget:
@amyblais I checked and proofread in context. That's ok for me. We can merge :)

PR #2473, opened 2019-01-17
Mattermost the Company

it33:


source/process/handbook.rst
```diff
@@ -34,3 +34,18 @@ Mattermost Team Edition under MIT license and the Mattermost open source project
...
+
+**Communities-centered** means we define our success in the context of the success of our `communities <https://docs.mattermost.com/process/community-overview.html>`_: users, customers, implementers, resellers, technology partners, contributors, and colleagues. The success of each community is owned by a member of the Mattermost leadership team. The plural definition of "communities" is intended to avoid unconsciously marginalizing downstream stakeholders. Referring to open source contributers as "the community", as if no other communities should be recognized, is like saying "a developer should format his code according to the style guide of the project". These statements may sound normal to the speaker, but they can unintentionally marginalize others.
```
lfbrock: > Referring to open source contributers as "the community", as if no other communities should be recognized, is like saying "a developer should format his code according to the style guide of the project". These statements may sound normal to the speaker, but they can unintentionally marginalize others.

1. Contributors is spelt wrong here 
2. I feel like this sentence is a little strange (like almost but not quite passive aggressive?)
3. I think the example used here  "a developer should format his code according to the style guide of the project" isn't clear enough to get your point across. I had to read it twice to understand the point, and I think there are many other people who will need to read it more times than that.

it33: Thanks @lfbrock, deleted the line. It's intention was to make people think, but it fails the "fast, obvious, forgiving" test, so removed it.

PR #2484, opened 2019-01-25
Update http: to https: for push.mattermost.com

cvitter:
Update http: to https: for push.mattermost.com. It appears that http://push.mattermost.com works and does not forward to https thus allowing for an insecure connection.


cpanato:
Humm, i will check what is happening and fix that

PR #2487, opened 2019-01-28
[WIP] Add Group Policy documentation

JtheBAB:
This add the documentation for the Group Policy feature of the Desktop App

cc: @wget

### DIFF
--- a/source/install/desktop.rst
+++ b/source/install/desktop.rst
@@ -35,6 +35,20 @@ Windows 10+, Windows 8.1+, Windows 7+
 
 This will start an installer for the app. Once finished, the Mattermost desktop app will open automatically.
 
+Group Policies
+~~~~~~~~~~~~~~~~~~~~~
+Copy the ``mattermost.admx`` from ``%INSTALLDIR\gpo_template`` to ``C:\Windows\PolicyDefinitions`` or if available ``\\FQDNDomain\sysvol\FQDNDomain\Policies\PolicyDefinitions``
+and ``mattermost.adml`` from ``%INSTALLDIR\gpo_template\en-US`` to ``C:\Windows\PolicyDefinitions\en-US`` or if availabe ``\\FQDNDomain\sysvol\FQDNDomain\Policies\PolicyDefinitions\en-US``
+then open the Group Policy Management Console and you should see under "Computer Configuration" --> "Administrative Templates" --> "Mattermost" the policies.
+
+Following Group Policies are available:
+
++------------------------+-----------------------------------------------------+----------------------+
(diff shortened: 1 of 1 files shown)


source/install/desktop.rst
```diff
@@ -35,6 +35,20 @@ Windows 10+, Windows 8.1+, Windows 7+
 
 This will start an installer for the app. Once finished, the Mattermost desktop app will open automatically.
 
+Group Policies
+~~~~~~~~~~~~~~~~~~~~~
+Copy the ``mattermost.admx`` from ``%INSTALLDIR\gpo_template`` to ``C:\Windows\PolicyDefinitions`` or if available ``\\FQDNDomain\sysvol\FQDNDomain\Policies\PolicyDefinitions``
```
wget: Is this normal the install directory variable is specified as `%INSTALLDIR` without a trailing `%`?

source/install/desktop.rst
```diff
@@ -35,6 +35,20 @@ Windows 10+, Windows 8.1+, Windows 7+
...
+and ``mattermost.adml`` from ``%INSTALLDIR\gpo_template\en-US`` to ``C:\Windows\PolicyDefinitions\en-US`` or if availabe ``\\FQDNDomain\sysvol\FQDNDomain\Policies\PolicyDefinitions\en-US``
```
wget: Typo `availabe` -> `available`

source/install/desktop.rst
```diff
@@ -35,6 +35,20 @@ Windows 10+, Windows 8.1+, Windows 7+
...
+then open the Group Policy Management Console and you should see under "Computer Configuration" --> "Administrative Templates" --> "Mattermost" the policies.
```
wget: Could we add `gpedit.msc` in parentheses after (Group Policy Management Console), that way power users will remember what it is about. This seems weird, but if we are talking to me about Group Policy Management Console, I don't know what it is about, `gpedit.msc` is clear though. Matter of use case maybe because I'm more used to use the Run command rather than clicking everywhere in a GUI.

source/install/desktop.rst
JtheBAB: Fixed

source/install/desktop.rst
JtheBAB: Fixed :/

source/install/desktop.rst
JtheBAB: gpesit.msc is the local group policy. A admin want to deploy it for the domain. See 
https://searchwindowsserver.techtarget.com/definition/Group-Policy-Management-Console

PR #2496, opened 2019-02-04
Upgrade - simpler way for copy paste

JtheBAB:
This is a suggestion to make the upgrade more copy paste.

/tmp is for such downloads perfect.
mattermost as user and group is defined as standard in the installation documentation.

I would also suggest that there is a link https://releases.mattermost.com/latest/mattermost-latest-linux-amd64.tar.gz instead of https://releases.mattermost.com/5.7.1/mattermost-5.7.1-linux-amd64.tar.gz

This would avoid wrong links or changes everywhere when a new version is released.

### DIFF
--- a/source/administration/upgrade.rst
+++ b/source/administration/upgrade.rst
@@ -105,6 +105,8 @@ Location of your local storage directory
    .. code-block:: sh
 
      sudo chown -hR mattermost:mattermost /tmp/mattermost-upgrade/
+     
+If you have used a different *{owner}* and *{group}* while setting up the server adjust the command accordingly.
 
 #. Copy the new files to your install directory and remove the temporary files.
 
--- a/source/administration/upgrade.rst
+++ b/source/administration/upgrade.rst
@@ -106,9 +106,10 @@ Location of your local storage directory
 
      sudo chown -hR mattermost:mattermost /tmp/mattermost-upgrade/
      
-   .. note:: If you didn't use `mattermost` as the owner and group of the install directory, run ``sudo chown -hR {owner}:{group} tmp/mattermost-upgrade/``
+   .. note::
+     If you didn't use ``mattermost`` as the owner and group of the install directory, run ``sudo chown -hR {owner}:{group} tmp/mattermost-upgrade/``
 
-   If you're uncertain what owner or group was defined, use the ``ls -l {install-path}/mattermost/bin/mattermost`` command to obtain them.
+     If you're uncertain what owner or group was defined, use the ``ls -l {install-path}/mattermost/bin/mattermost`` command to obtain them.
 
 #. Copy the new files to your install directory and remove the temporary files.
 
--- a/source/administration/upgrade.rst
+++ b/source/administration/upgrade.rst
@@ -24,8 +24,6 @@ Existing install directory - *{install-path}*
   If that command does not produce any results because your version is older, try ``whereis platform`` instead.
 Location of your local storage directory
   The local storage directory contains all the files that users have attached to their messages. If you don't know its location, open the System Console and go to **Files > Storage** and read the value in **Local Storage Directory**. Relative paths are relative to the ``mattermost`` directory. For example, if the local storage directory is ``./data/`` then the absolute path is ``{install-path}/mattermost/data``.
-Owner and group of the install directory - *{owner}* and *{group}*
-  Use the ``ls -l {install-path}/mattermost/bin/mattermost`` command to get the owner and group.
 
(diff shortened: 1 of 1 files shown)


jasonblais:
Thank you @JtheBAB!  Couple of initial questions:

1. Are there any potential side effects of this change?
2. Where do you see the reference for `https://releases.mattermost.com/5.7.1/mattermost-5.7.1-linux-amd64.tar.gz` ?

svelle:
@jasonblais I think he might be referring to the server install docs. There's no direct copy paste link inside the documentation because you'd have to specify the version you want.  
Adding a `latest` link would partially solve this because you could add a copy paste box for the latest version and then explain below that different versions can be downloaded the way it's currently described.

jasonblais:
Makes sense, looks like there's actually a ticket for that already here: https://mattermost.atlassian.net/browse/MM-11815

JtheBAB:
I don't think there is a side effect. Perhaps that /tmp is too small for the tgz and the extracted folder. The user:group change of course is just easier when you install it as Mattermost recommends. But i think we can tune the docs to make it easier for users that follow all the docs.

Yes, ticket MM-11815 would be great.

jasonblais:
@JtheBAB @svelle One last question - for those who don't use `mattermost` as user and group, will these upgrade steps work? I'm not sure if `mattermost` was always the user and group.

The updated steps sound good, I just want to make sure there won't be issues for some admins.

JtheBAB:
They would need to adjust the command.

source/administration/upgrade.rst
```diff
@@ -106,16 +104,16 @@ Owner and group of the install directory - *{owner}* and *{group}*
 
    .. code-block:: sh
 
-     sudo chown -hR {owner}:{group} {path-to}/mattermost-upgrade/
+     sudo chown -hR mattermost:mattermost /tmp/mattermost-upgrade/
```
svelle:
```suggestion
     sudo chown -hR mattermost:mattermost /tmp/mattermost-upgrade/
     
     If you have used different *{owner}* and *{group}* while setting up the server adjust the command accordingly.
```

source/administration/upgrade.rst
```diff
@@ -24,8 +24,6 @@ Existing install directory - *{install-path}*
   If that command does not produce any results because your version is older, try ``whereis platform`` instead.
 Location of your local storage directory
   The local storage directory contains all the files that users have attached to their messages. If you don't know its location, open the System Console and go to **Files > Storage** and read the value in **Local Storage Directory**. Relative paths are relative to the ``mattermost`` directory. For example, if the local storage directory is ``./data/`` then the absolute path is ``{install-path}/mattermost/data``.
-Owner and group of the install directory - *{owner}* and *{group}*
-  Use the ``ls -l {install-path}/mattermost/bin/mattermost`` command to get the owner and group.
```
svelle: I'd suggest not removing this line just incase anyone uses a differing setup. 

Maybe add something along the lines of: `If you have followed the server setup from the documentation *{owner}* and *{group}* should both be mattermost.

svelle:
@jasonblais I've added some comments to the changes. The wording is of course subject to change but I think this should cover all the cases.

source/administration/upgrade.rst
JtheBAB: I think this would be too much. I mean user that does it manually should know there owner:group that they give. All others could profit from #2502

JtheBAB:
Dont know if you get a notification for the comment i made ^

source/administration/upgrade.rst
jasonblais: > I mean user that does it manually should know there owner:group that they give

That's true, but sometimes IT admin roles change and the information gets lost.

My worry is that if the new IT admin then follows the admin doc (with owner=group=mattermost) and it fails, they have no idea why and they panic.

I like the suggestion of having a note after step 9. Admins can skip it if it doesn't apply to them. Maybe we can format it as a note like below

```
.. note:: If you used a different owner and group of the install directory, run ``sudo chown -hR {owner}:{group} tmp/mattermost-upgrade/``

If you're uncertain what owner or group was defined, use the ``ls -l {install-path}/mattermost/bin/mattermost`` command to get the owner and group.
```

Thoughts @svelle or @JtheBAB

svelle: Good proposal @jasonblais I'd go with this one.

JtheBAB:
I added the suggestion from @jasonblais 

When #2502 gets more love ( ; ) ) the majority will (hopefully) use the script.

PR #2501, opened 2019-02-06
Change heading for "0/5" subsection to properly link by autolink plugin

jasonblais:
@lieut-data To fix the autolink for "0/5, 1/5, .." (https://github.com/mattermost/platform-private/pull/146) - I didn't look into it in depth, but I don't think there is an easy way to make #id1 or #id8 static -- they are dynamic anchors by nature.

So to fix it properly for long-term, we should make a static section header for the term. I've included a proposal with this PR. The anchor used for the autolink plugin would be #out-of-5

### DIFF
--- a/source/process/training.rst
+++ b/source/process/training.rst
@@ -407,11 +407,6 @@ Naming specific repeated mistake helps us find patterns, avoid repeated mistakes
 List of Terms
 ---------------------------
 
-0/5, 1/5, 2/5, 3/5, 4/5, 5/5
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


jasonblais:
I tried that, but when I try `https://github.com/mattermost/docs/blob/43231fe691895bef311575e9fb03a33c59e14bfb/source/process/training.rst#id8`, it doesn't seem to link to the correct section.

I updated my PR to use ref so we can try it out, in case the update I tried in the past was incorrect.

Also, added `_out-of-5` reference.

lieut-data:
I /think/ that's because GitHub doesn't fully process reST like Sphinx does -- if you compare the generated code GitHub makes:
image
with what's on docs.mattermost.com:
image

It appears that GitHub makes different ids. I tried the above change locally and it /seemed/ to work fine.

jasonblais:
Interesting. Thanks for sharing! I learned something new today :)

PR #2502, opened 2019-02-06
Script for Autoupdating Mattermost

JtheBAB:
I wrote a small script that automates the process of updating Mattermost with a simple shell script instead of doing all the commands manually.


jasonblais:
@JtheBAB First, I love the idea of being able to upgrade Mattermost in 3 simple steps.

I'm wondering
1) How much work would it be to add an auto-update option in the Mattermost product itself, with a script like you have (that would be AWESOME but might be more complicated  )
2) Any issues admins might hit because of how they installed Mattermost?

JtheBAB:
>     1. How much work would it be to add an auto-update option in the Mattermost product itself, with a script like you have (that would be AWESOME but might be more complicated  )

When the script would be delivered with mattermost and mattermost could do a sh command and this command is still executed when mattermost itself is no longer running then it should work.
The question is more if it is not worth to think about creating rpm's (RedHat/CentOS/...) and deb (Debian/Ubuntu) files and distribute updates then with yum/apt. Then the admin could just add the mattermost repository and update it the same way as a linux kernel.

>     2. Any issues admins might hit because of how they installed Mattermost?

As the script is customizable i think not. It supports the OS that Mattermost supports and the path for Mattermost can be somewhere. For example my Mattermost instance is /usr/mattermost.

But yes testing on different systems would be good. I will for sure use it again to update to 5.8.

source/administration/update_mattermost.sh
```diff
@@ -0,0 +1,101 @@
...
+
+#delete old backup folder
+if [ -d "$backupdir/mattermost-backup" ]; then
+        echo Deleting old mattermost backup
+        rm -rf $backupdir/mattermost-backup
+fi
+
+###Get the file
+echo Downloading Mattermost $edition $version
+wget $downloadlocation -P $downloaddir -q
```
saturninoabril: Should we download first before deleting the old backup so that in case the download fails, the old back up is still available?

source/administration/update_mattermost.sh
```diff
@@ -0,0 +1,101 @@
...
+if pgrep mattermost > /dev/null
+then
+        echo Mattermost still running. Update not possible.
+        echo cleaning
+        rm -rf $downloaddir/mattermost-upgrade/
+        rm -f $downloaddir/mattermost*.gz
+        exit 1
+fi
+
+echo Create backup of mattermost directory
```
saturninoabril: Does it make sense to add notice/warning to back up the database separately according to own's standard procedures?

source/administration/update_mattermost.sh
```diff
@@ -0,0 +1,101 @@
...
+cp -ra $mattermostdir/ $backupdir/mattermost-backup/
```
saturninoabril: Maybe align back up directory naming convention similar to manual where it includes date/time as name suffix?

source/administration/upgrade.rst
```diff
@@ -34,6 +34,62 @@ Owner and group of the install directory - *{owner}* and *{group}*
 
 #. Review the :doc:`important-upgrade-notes` to make sure you are aware of any actions you need to take before or after upgrading from your particular version.
 
+**With a script**
```
saturninoabril: I feel like the manual upgrade steps should come first before this "with script".

saturninoabril:
I'm requesting for another review from @cpanato as he's expert in this domain :)

source/administration/update_mattermost.sh
```diff
@@ -0,0 +1,101 @@
...
+USER=$(stat -c '%U' $mattermostdir/bin/mattermost)
+GROUP=$(stat -c '%G' $mattermostdir/bin/mattermost)
+chown -hR $USER:$GROUP $downloaddir/mattermost-upgrade/
+#clean up mattermost directory
+find $mattermostdir/ -mindepth 1 -maxdepth 1 \! \( -type d \( -path $mattermostdir/config -o -path $mattermostdir/logs -o -path $mattermostdir/plugins -o -path $mattermostdir/data \) -prune \) | sudo xargs rm -r
+
+#rename plugin directory
+if [ $plugins -eq 0 ];  then
+        echo Renaming plugin folder
+        mv $mattermostdir/plugins/ $mattermostdir/plugins~
```
cpanato: dont need to do the same for the `client/plugins` ?

JtheBAB: I sticked to the manual steps. So if this is needed, this step would also be needed within the manual steps

source/administration/upgrade.rst
JtheBAB: I put it there because the script (should) be easier then the manual steps. What do you think @jasonblais?

source/administration/update_mattermost.sh
JtheBAB: I made it fix as it simplified the script and the backup from 2 version ago doesn’t matter right? Because when you need to rollback the Binary and the DB you are anyway screwed because of loosing data

source/administration/update_mattermost.sh
JtheBAB: Could add a confirmation question like „Did you backup your Mattermost Database?“. And of course an override for that question.

source/administration/update_mattermost.sh
JtheBAB: Makes sense. I will update accordingly

source/administration/update_mattermost.sh
saturninoabril: Sounds good to me, thanks.

source/administration/update_mattermost.sh
saturninoabril: Each version is different (folder might change, JS files/maps are different).  So if the admin, didn't clear up the back up directory before upgrade, it will end up with mixed folders/files from old and new version, and might not be usable in case the admin needs to revert from the recent upgrade. I think that's the reason why it ends up with backing up to new folder (with date/time) so that admin don't need to worry about that and may attend to those back up folders separately.

source/administration/upgrade.rst
saturninoabril: My reason is for the admin to see the manual upgrade first (obviously) and hopefully try it so that the admin may control/feel the entire upgrade process before trying with script.  I think the first part of "with script" - configuration - is key to success of the upgrade process but also easy to make mistake if you're doing the upgrade for the first time.

source/administration/update_mattermost.sh
JtheBAB: @cpanato 

I check'd my client/plugins and i have "just" one plugin (Spoiler Alert by moustec) that put some data there. For what is that folder?

source/administration/update_mattermost.sh
JtheBAB: ? Thats the reason why the backup folder gets first deleted. So it the admin has always the recent backup.

Example:
- current version is 5.7.1
- admin uses the script to update to 5.8.0
- backup contains 5.7.1
- admin uses the script to update to 5.9.0
- backup of 5.7.1 get's deleted
- backup of 5.8.0 is created
- update to 5.9.0 happen

source/administration/update_mattermost.sh
cpanato: plugins that have the client side will be unpacked to the client/plugins folder.
for plugins with server code that will be unpacked to the plugins folder in the root

cpanato: i will test out this script

wget:
I'll be again considered the Ugly Duckling here, but I'm now used to be 😁

I'm strongly against such a script:
* First the script is vulnerable and could lead to really bad things on the system (like losing the whole server data). Please check with `shellcheck` for more pieces of info (doubles quotes, shell expansion, etc.).
* The user might not have installed Mattermost to these specific locations written in the script.
* Like said by jthebab, sysadmins are usually updating software using the package manager of the GNU/Linux distribution.
* Concentrate on upstreaming a .deb that will be auto updated and pushed to the official Debian repo is IMHO a better idea. I have several friends of mine who are Debian developers and could commits our patches easily.

cpanato:
@wget you have a good valid points which makes a lot of sense

JtheBAB:
@wget 

> First, the script is vulnerable and could lead to really bad things on the system (like losing the whole server data). Please check with `shellcheck` for more pieces of info (doubles quotes, shell expansion, etc.).

Shellcheck is "only" complaing about the double quotes - which is fixable - and https://github.com/koalaman/shellcheck/wiki/SC2181 --> everything that can be fixed.

> The user might not have installed Mattermost to these specific locations written in the script.

Thats why the user can specify the directory at the top of the script..... As i wrote, my mattermost is in /usr instead of the recommended /opt. I can add a check if the Mattermost file is really within the path. Then this is checked.

> Concentrate on upstreaming a .deb that will be auto updated and pushed to the official Debian repo is IMHO a better idea. I have several friends of mine who are Debian developers and could commit our patches easily.

rpms or debs are always better. I would offer it as separate rpm/deb location. As far as i know Debian supports not the fast Mattermost cycle on official repos?

JtheBAB:
I updated the script to add more checks and asking if the admin made a database backup.

@saturninoabril @jasonblais @cpanato

JtheBAB:
I just want to add that i successfully updated my production mattermost instance from 5.7.1 to 5.8.0 with this script :)

source/administration/update_mattermost.sh
saturninoabril: Oh right, I missed that delete backup block above, maybe move it near this create block?
But still I'm in the impression that this "with script" and manual upgrades should at least be doing the same, especially when saving backup directory. Non-blocking, just need to give more thought about it.

wget:
Hello guys. I updated the script to make it more robust.
- Added proper quotations
- Using the PATH env to check dependency rather than the hardcoded path values

@JtheBAB A few remarks:
- Pay attention at the fix of spaces, tabs and your code indentation
- GNU/Linux distributions might not have `wget` installed (this is the case of Arch Linux for example). Same applies with `curl`.
- Make sure you use simpler syntax for the command in the future: regex replaces etc. in tar and nested spaces for `find`  was a bit ugly.

Could you please retest the script with the modifications I made?

Otherwise, I still don't see the purpose of such a script, since this is very specific to one environment (even if we can change config values at the start of the script). Indeed, RPM package has landed during the week-end, and official .deb are in the process to be merged upstream as soon I got this damn msi/new CI fully working ;) This is why I would have preferred to have such a script kept on your own repo rather than merged here, but I'm not the only one to decide :)

hanzei:
Site note: There is HW ticket for creating a PPA for Mattermost: https://github.com/mattermost/mattermost-server/issues/4176

source/administration/upgrade.rst
```diff
@@ -34,6 +34,62 @@ Owner and group of the install directory - *{owner}* and *{group}*
 
 #. Review the :doc:`important-upgrade-notes` to make sure you are aware of any actions you need to take before or after upgrading from your particular version.
 
+**With a script**
+
+*Preparing the script*
+
+`Save the script<https://docs.mattermost.com/administration/update_mattermost.sh>`__ to your mattermost server.
```
amyblais: Change to this so that the link works:

_`Save the script <https://docs.mattermost.com/administration/update_mattermost.sh>`__ to your Mattermost server.

source/administration/upgrade.rst
```diff
@@ -34,6 +34,62 @@ Owner and group of the install directory - *{owner}* and *{group}*
...
+
+.. code-block:: sh
+
+     chmod +x ./update_mattermost.sh
+    
+Please adjust the parameters at the beginning of the script according to your environment.
+
+*Start the script*
+
+To start the update process start the script and add the desired version number as argument.
```
amyblais: Add comma after "process".

Add "an" before the word "argument"

source/administration/upgrade.rst
```diff
@@ -34,6 +34,62 @@ Owner and group of the install directory - *{owner}* and *{group}*
...
+
+     sh update_mattermost.sh %VERSION%
+
+Example:
+
+.. code-block:: sh
+
+     sh update_mattermost.sh 5.7.1
+     
+After the update the following additional steps are required:
```
amyblais: Add comma after "update".

source/administration/upgrade.rst
```diff
@@ -34,6 +34,62 @@ Owner and group of the install directory - *{owner}* and *{group}*
...
+     cd {install-path}/mattermost
+     sudo setcap cap_net_bind_service=+ep ./bin/mattermost
+
+#. Upgrade your ``config.json`` schema:
+
+   #. Open the System Console and change a setting, then revert it. This should enable the Save button for that page.
+   #. Click **Save**.
+   #. Refresh the page.
+
+   Your current settings are preserved, and new settings are added with default values.
```
amyblais: Remove comma.

source/administration/upgrade.rst
```diff
@@ -34,6 +34,62 @@ Owner and group of the install directory - *{owner}* and *{group}*
...
+
+After the server is upgraded, users might need to refresh their browsers to experience any new features.
+
+3. Re-instate, if needed the ``plugins`` directory, then restart the mattermost service.
```
amyblais: Change "Re-instate to "Reinstate".

Also, this sentence seems incomplete / hard to read.

source/administration/update_mattermost.sh
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env bash
```
amyblais: If this is a new doc it should be in the table of contents: https://docs.mattermost.com/guides/administrator.html.

Also, I'm not able to get a proper preview of this page when clicking "View file".

source/administration/upgrade.rst
wget: @saturninoabril Completely agree with you here. Doing the change.

source/administration/update_mattermost.sh
wget: The doc is already in the table of content.

The file that you are talking about (which is not proper previewed) is not a documentation text (a restructured fie) but rather a Bash script. This is code, this is thus normal you don't see any proper preview.

wget:
This PR is pending until the discussion about which documentation guidelines to adopt hasn't been solved:
https://community.mattermost.com/core/pl/rnnfufeud7dkdyd1ynncnsd36o

hanzei:
@JtheBAB Would you please merge `master` into your branch?

JtheBAB:
@hanzei 

I will try my best as soon as the Reviews are done. Perhaps this get never merged.

hanzei:
Hey @JtheBAB, gently reminder about the above :point_up:

JtheBAB:
@hanzei 
Done. I hope...

hanzei:
@wget @svelle Would you please provide guidance on the next steps? Is the discussion you linked above finished?

JtheBAB:
We can also close this PR when the script is not what Mattermost want. I have my script and use it monthly :D

uhlhosting:
I also have made a small .sh to upgrade mattermost, will happy try this one, my version have issues on updating plugins side.

uhlhosting:
Any updates on this?

justinegeffen:
Following up on this to confirm the status and whether the next steps have been agreed-on. :)

JtheBAB:
At the end Mattermost needs to decide if they want to promote such script or not

uhlhosting:
Wether script or not mattermost need one click upgrades. Common folks were speaking of web technologies here. Security? Manual upgrades are much more prone to human errors than a well coded module would.

justinegeffen:
I'm adding @jasonblais to this for feedback regarding the promotion of the script and plans for one-click upgrades.

jasonblais:
Thank you everyone! It's great to see so much activity from the community here.

Regarding the promotion of the script - @justinegeffen we've been considering a "community" section with the new information architecture that's under discussion - and at one point saw a "Community Guides" or "Unofficial Guides" subsection there. I'm wondering if this script would be a good example of such a community guide - when the Mattermost team doesn't have the time to officially test and maintain a piece of documentation, the community can still find them under community/unofficial guides. cc @JtheBAB for thoughts as well

As for one-click upgrades - thanks for the feedback! There's an [existing feature proposal for auto-updates](https://mattermost.uservoice.com/forums/306457-general/suggestions/12593031-should-implement-an-admin-interface-to-the-system) - @uhlhosting Would you be open to upvote the feature proposal?

JtheBAB:
Sounds good. Vote for Community Guides.

justinegeffen:
That sounds like a great fit, @jasonblais. I'm more than happy to help get that off the ground if needed.

uhlhosting:
Voted +1

justinegeffen:
Hi @JtheBAB. We have a new [Community-Managed Documentation]()https://docs.mattermost.com/guides/administrator.html#community-managed-documentation) section in the Admin Guide. If this is still an applicable solution, I propose adding it there. @jasonblais, what are your thoughts?

PR #2512, opened 2019-02-11
Add FAQ on how to add a message attachment via the API

jasonblais:
Not part of this PR, but we should at some point move the dev-focused integration documentation from https://docs.mattermost.com/guides/integration.html to developers.mattermost.com

### DIFF
--- a/source/developer/message-attachments.rst
+++ b/source/developer/message-attachments.rst
@@ -143,6 +143,12 @@ You need to add an "attachment" key to the post's props JSON field. The value is
 
 .. code-block:: bash
 
+  curl -i -X POST -H 'Content-Type: application/json' -d '{"channel_id":"qmd5oqtwoibz8cuzxzg5ekshgr", "message":"Test message #testing", "props":{"attachments": [{"pretext": "This is the attachment pretext.","text": "This is the attachment text."}]}}' http://{your-mattermost-site}/posts
+
+Below is an example HTTP request:
+
+.. code-block:: http
+
   POST /posts HTTP/1.1
   Host: {your-mattermost-site}
   User-Agent: curl/7.63.0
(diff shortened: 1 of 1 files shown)


source/developer/message-attachments.rst
```diff
@@ -124,10 +124,33 @@ And here is how it renders in Mattermost:
...
+---------------------------
+
+Can I send a message attachment via the API?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Yes, you can use the `create post RESTful API <https://api.mattermost.com/#tag/posts%2Fpaths%2F~1posts%2Fpost>`_.
+
+You need to add an "attachment" key to the post's props JSON field. The value is an array of message attachments you want attached to the post. See below for an example curl command.
+
+.. code-block:: bash
```
crspeller: Why are there two similar examples?

jasonblais: One with bash, another with an HTTP request, like in our incoming webhook dev doc example: https://developers.mattermost.com/integrate/incoming-webhooks/#basic-usage

crspeller: The second example is not an HTTP request the proper example for the curl command is:
```http
POST /posts HTTP/1.1
Host: {your-mattermost-site}
User-Agent: curl/7.63.0
Accept: */*
Content-Type: application/json
Content-Length: 192

{"channel_id":"qmd5oqtwoibz8cuzxzg5ekshgr", "message":"Test message #testing", "props":{"attachments": [{"pretext": "This is the attachment pretext.","text": "This is the attachment text."}]}}
```

jasonblais: Thanks Chris! Updated doc

crspeller: I was unclear, sorry. You want to keep the old curl example and replace the HTTP request example with the above.

jasonblais: That makes more sense--like so?

crspeller: Yes!

PR #2514, opened 2019-02-12
Update image proxy documentation

jasonblais:
For https://github.com/mattermost/mattermost-server/pull/9967


hmhealey:
I've made a few changes to:
1. Clarify the differences between the local and atmos/camo proxies
2. Remove some unnecessary details about the local image proxy like that it uses the willnorris/imageproxy and about some features that we didn't actually support.

I re-added Amy for a second editor review in case I messed something up.

PR #2533, opened 2019-02-20
Differentiate /join from /open

staabm:
Better describe the difference of this 2 commands


staabm:
Just signed the CLA

amyblais:
Thank you for the pull request @staabm,

I think in order to change the documentation we would first need to edit the help text for the slash commands themselves (the help text on slash commands match the text in this documentation). @jasonblais is that accurate?

jasonblais:
Thanks @staabm! The two commands actually work the same way - for instance, both let you join a channel. So I'm curious if we'd keep the command help texts the same, to signify they are the same commands (with different trigger words).

We also have the same with `/remove` and `/kick` that do the same thing.

staabm:
maybe it should be more explicit then, like `/join` is an alias for `/open` or vice-versa

jasonblais:
@staabm Yeah that could work, with something like `/join (or /open) {channel-name}`?

There may be better ways of displaying the alias.

staabm:
I would suggest

```
    "/join *{channel-name}*", "Join the given channel", "/join off-topic"
    "/open *{channel-name}*", "Alias for /join", "/open off-topic"
```

instead of

```
    "/join *{channel-name}*", "Join the given channel", "/join off-topic"
    "/open *{channel-name}*", "Open the given channel", "/open off-topic"
```

what we have today

jasonblais:
Thanks @staabm - stepping back, I'm curious what problem we're trying to solve. Is the current list confusing when the two help texts are identical for `/join` and `/open`?

staabm:
While reading the docs the text felt like it contains a copy&paste error because both lines are similar

JtheBAB:
Join and open is for me something else.

The first time i join the channel. Afterwards i open the channel as i am already joined the first time. 

Didn’t tried it but what happened when you /open channel before you have joined? And what happen when you /open a channel where you dont have the permission? Same behavior as /join?

jasonblais:
/open and /join are exactly the same functionality now. They may even use the same code path.

JtheBAB:
Ok. Then why not remove /join as Help wanted?

jasonblais:
> Ok. Then why not remove /join as Help wanted?

/join and /open do the same by design, in fact. It was done to follow Slack's model at the time.

jasonblais:
> Ok. Then why not remove /join as Help wanted?

/join and /open do the same by design, in fact. It was done to follow Slack's model at the time.

staabm:
I will change the docs to whatever you like most. Just tell me how things should read

source/developer/slash-commands.rst
```diff
@@ -41,8 +41,8 @@ Each Mattermost installation comes with some built-in slash commands that are re
     "/invite_people *{name@domain.com ...}*", "Send an email invite to your Mattermost team","/invite_people john@example.com"
     "/kick *{@username}*", "Remove a member from a public or private channel", "/kick @alice"
     "/remove *{@username}*", "Remove a member from a public or private channel", "/remove @alice"
-    "/join *{channel-name}*", "Join the open channel", "/join off-topic"
-    "/open *{channel-name}*", "Join the open channel", "/open off-topic"
+    "/join *{channel-name}*", "Join the given channel", "/join off-topic"
```
jasonblais: `"/join (or /open) *{channel-name}*", "Join the given channel", "/join off-topic"`

source/developer/slash-commands.rst
```diff
@@ -41,8 +41,8 @@ Each Mattermost installation comes with some built-in slash commands that are re
     "/invite_people *{name@domain.com ...}*", "Send an email invite to your Mattermost team","/invite_people john@example.com"
     "/kick *{@username}*", "Remove a member from a public or private channel", "/kick @alice"
```
jasonblais: `"/kick (or /remove) *{@username}*", "Remove a member from a public or private channel", "/kick @alice"`

staabm:
Adjusted per feedback

Issue #2474, opened 2019-01-18
Update Zoom plugin instructions


icelander:
Zoom has made changes in how they create webhooks, so the [Zoom plugin guide](https://docs.mattermost.com/integrations/zoom.html) needs updated.

From Zoom support: 

> The username and password fields are for webhook apps setup using the V1 webhook. The new marketplace webhooks don't have that option. You can recreate the webhook using the marketplace to circumvent entering the credentials into those fields.

svelle:
I'd suggest waiting till Feb 11th, because then Zoom will deprecate the old developer portal and move to the new site. 
After that new screenshots can be made for the docs because right now the dev portal is full of deprecation warnings and in the set up process you're switching between old and new site all the time.

svelle:
This can be closed now I guess. @amyblais @jasonblais

PR #2538, opened 2019-02-22
bring zoom plugin docs up to date with zoom v2 api

svelle:
This is in reference to:  
#2474   

I removed step for because it seems that the new api has made this step redundant. Feel free to correct me though.  

I tried all the steps on my current instance and the plugin works just fine. I can create and join meetings from within Mattermost and so can others.


svelle:
just saw that I messed up the branch to be pulled from. Will reopen with correct branch

PR #2540, opened 2019-02-22
Fix contributors guidelines link

svelle:
Currently the link in the CONTRIBUTORS.md leads to the old docs page which in turn just has a link to the new docs page. This PR fixes this mild annoyance ;)


svelle:
Quickly cleaned up the commit history, don't know how that second commit came in there. All fixed now :)

PR #2545, opened 2019-02-23
Update links for pricing and product pages in release checklist

jasonblais:
Also, let's confirm the pages are updated, e.g. MFA is still listed as E10, and there is no LDAP group sync.

(I think you posted about this to marketing team, but maybe we need to update our process to make sure they get done)


amyblais:
- I don't see features listed at https://mattermost.com/product/ so I think it can be removed
 - Do we add experimental features to the pricing page (LDAP group sync)?

jasonblais:
We have features listed at /product, e.g.

image

(not at the top, but a bit further down)

> Do we add experimental features to the pricing page (LDAP group sync)?

I'm not sure, but a good point. My guess is that probably not.

PR #2550, opened 2019-02-25
Adding import-bulk command to the documentation

jespino:


source/administration/command-line-tools.rst
```diff
@@ -659,6 +659,29 @@ mattermost import
   Child Command
```
amyblais: Please include this also up here in the table of contents section.

jespino: Done, thanks! :)

wiersgallak:
@jespino I see you have a dev review label on this and George was on it and then removed. Does this still need dev review?

jespino:
@wiersgallak I think so, I think was removed by accident.

PR #2559, opened 2019-02-27
Document customer obsession meeting

jasonblais:


source/process/training.rst
```diff
@@ -106,7 +106,66 @@ Mattermost Staff Meetings
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Attendees:
+
+- All Mattermost Staff
+
+Objectives:
+
+- Share important company-wide updates with staff across the organization.
+- Educate staff on key customer use cases and feature requests.
```
it33: Propose instead of "feature requests" we call them "needs", it's not about customers filing issues, it's about us understanding what they need whether or not they ask for it.

source/process/training.rst
```diff
@@ -106,7 +106,66 @@ Mattermost Staff Meetings
...
+
+Time:
+
+- Weekly meeting on Wednesdays from 8:00am to 8:30am Palo Alto time.
```
it33: 8:02am? @comharris can you help update the meeting invite as well?

source/process/training.rst
jasonblais: Updated

source/process/training.rst
```diff
@@ -106,7 +106,66 @@ Mattermost Staff Meetings
...
+
+Procedure:
+
+**Before meeting:**
+
+1. (Chair) Schedules items for the weekly agenda, includes company update from CEO or CTO, and customer updates.
```
it33: The job of CEO is to lead and empower the company's leaders. CEO should do the introduction and unless there's something special, department heads should share the updates.

source/process/training.rst
it33: Actually, @comharris @jasonblais thinking more about this, why do we want to make it a norm for people to show up late to meetings? 

If we had an office with a printer, would it be okay for someone to steal the printer? 

The value of 2 minutes of time from everyone in the company is worth a lot more than a printer. I don't understand why we would set the norm for the company for people to show up late to meetings.

jasonblais: @it33 It's not about letting them come late to the meeting. But people have meetings that end right before, so it takes a moment for them to get to the next Zoom link and join the next call.

If we were in an office, it would be the equivalent of walking from one meeting room to the next.

That said, instead of defaulting to 30/60 minute meetings and starting the Cust Obs call at 8:02am, maybe a better resolution is to set a norm that meetings last 25/55 minutes. That way people make it on time for the next meeting (whether it's Cust Obs or another meeting). 

Thoughts?

jasonblais: https://meetingking.com/how-long-should-a-meeting-be/

> Usually meetings start late because people have the “excuse” that they came from a previous meeting. 

...

> To make sure this does not happen again in your organization, schedule your meetings to end 5 minutes before the bottom and the top of the hour. So your meetings will now be 25 and 55 minutes instead of 30 and 60 minutes. This gives participants time to go to the next meeting or dial in and start the next meeting on time.

source/process/training.rst
jasonblais: You are also giving an update tomorrow (about YC). So how about company updates are from "leads"?

0/5, "leads" may not be the correct term here.

source/process/training.rst
jasonblais: As discussed, proposal is to start meetings on time, updated doc accordingly. Resolving

jasonblais:
Merging to be ready for a call in a couple of hours. Adding editor review label to make edits separately.

Thanks Ian for the feedback!

PR #2564, opened 2019-02-27
Update people-ops.rst

comharris:
Updated email address from people@ to hr@


comharris:
Understood @cpanato . Since these letters often include confidential info (e.g. salary), it's best the request goes directly to HR rather than "people" (non-confidential asks). Hope this helps clarify some.

PR #2567, opened 2019-02-28
Restore text about invitations expire time to 48h, and removed relation with the InviteId.

jespino:
The email invitations were bound to the InviteId, but we change it, and now are
independent.

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -1502,14 +1502,6 @@ Require Email Verification
 | This feature's ``config.json`` setting is ``"RequireEmailVerification": false`` with options ``true`` and ``false`` for above settings respectively.                 |
 +----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
-Email Invite Salt
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-32-character (to be randomly generated via System Console) salt added to signing of email invitation links. Click **Regenerate** to create new salt.
-
-+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-| This feature's ``config.json`` setting is ``"InviteSalt": ""`` with string input.                                                                                    |
-+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-
 Enable Open Server
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 **True**: Users can sign up to the server from the root page without an invite.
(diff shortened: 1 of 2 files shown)


source/administration/config-settings.rst
```diff
@@ -1504,7 +1504,7 @@ Require Email Verification
 
 Email Invite Salt
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-32-character (to be randomly generated via System Console) salt added to signing of email invitation links. Email invitation links expire after 24 hours except if the salt is regenerated, then existing email invitation links invalidate immediately.  Click **Regenerate** to create new salt.
+32-character (to be randomly generated via System Console) salt added to signing of email invitation links. Click **Regenerate** to create new salt.
```
esethna: Can we keep the text as is (unless it's inaccurate?), but just change the 24 > 48?

source/help/getting-started/managing-members.rst
```diff
@@ -11,7 +11,7 @@ If enabled by your System Admin, you can add people to your team in one of three
 Direct Invites
 ~~~~~~~~~~~~~~
 
-Direct invites are invitation emails sent from your team's server directly to the invited person's email address. A link within the invitation directs them to an account creation page. Invitation links sent by email expire after 24 hours or invalidate immediately if the System Admin regenerates the `Email Invite Salt <https://docs.mattermost.com/administration/config-settings.html#email-invite-salt>`_.
+Direct invites are invitation emails sent from your team's server directly to the invited person's email address. A link within the invitation directs them to an account creation page. Invitation links sent by email expire after 48 hours.
```
esethna: Same here

source/administration/config-settings.rst
jespino: The thing is exactly that, is no longer true. Some time ago we separate entirely the concept of InvitationId and the invitation by email. So, now, if you change the invitation id, it doesn't revoke the current existing invitations for the team. I going to check it manually, but I'm almost sure.

wiersgallak:
@esethna After conversation with @jespino  - the validation of email invites have changed.  The new way involves creating a new token for each user that is invited.  This token is rejected if a user tries to use it after the 48 hours.  Once it is used, it is deleted from the server and not available to be used by another user.   He will be reviewing the ability for admins to invalidate the token immediately if they need to and confirming exactly what the email salt is used for in this new model. We will update the docs accordingly to specify the single use and how email salt is affected.

jespino:
Confirmed that the InviteSalt is no longer used. So we can remove it entirely from the application code.
For now is not possible to invalidate the email invites, but we can implement it if needed, we can add a button to invalidate all email invites, or maybe one for all the invites from an specific team, or we can create a new interface to manage the current existing invitations of the team and invalidate each one independently.

jespino:
Cleaning up InviteSalt here mattermost/mattermost-webapp#2500 and here mattermost/mattermost-server#10446

jespino:
@wiersgallak I added some changes, can you re-review it?

PR #2571, opened 2019-03-01
v5.9 Docs - documenting breaking change

amyblais:
Please edit / add missing details.

I'll also add this to changelog and let Gitlab know.

https://github.com/mattermost/mattermost-server/pull/10356

(Also needs config.json setting added to docs)


amyblais:
Also added the config setting to this same PR, but it needs the details filled out and reviewed if it's in the correct place.

amyblais:
@hmhealey / @jasonblais wanted to confirm that I should inform Gitlab about this since MFA was moved to TE recently?

https://github.com/mattermost/mattermost-server/pull/10356#pullrequestreview-209118855

source/administration/config-settings.rst
```diff
@@ -4095,6 +4095,17 @@ Supported for Mattermost server v5.1.0 and later, and Mattermost Mobile apps v1.
 | This feature’s ``config.json`` setting is ``"ExperimentalLimitClientConfig": "false"`` with options ``true`` and ``false`` for above settings respectively.       |
 +-------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
+Disable Legacy MFA
+^^^^^^^^^^^^^^^^^^^^
+
+**True**:
```
DSchalla:
```suggestion
**True**:
Disables the legacy `checkMfa` endpoint which is only required to support mobile clients before the version 1.17 when you are using Multifactor-Authentication. Recommended to set to `true` if upgrading the mobile application in your environment is possible. 
```

source/administration/config-settings.rst
```diff
@@ -4095,6 +4095,17 @@ Supported for Mattermost server v5.1.0 and later, and Mattermost Mobile apps v1.
...
+
+**False**:
```
DSchalla:
```suggestion
**False**: 
Keeps the legacy `checkMfa` endpoint enabled to support mobile versions older than version 1.17. Please note that doing so will open up an information disclosure about the MFA status of system users.
```

DSchalla:
@crspeller To double check the wording / information provided of the configuration setting

jasonblais:
@amyblais If this is a breaking change to MFA, we should let GitLab know as it's shipped in GitLab. I wish there was a way this wasn't a breaking change, though

source/administration/important-upgrade-notes.rst
```diff
@@ -7,6 +7,9 @@ Important Upgrade Notes
 +----------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | If you’re upgrading from a version earlier than... | Then...                                                                                                                                                          |
 +====================================================+==================================================================================================================================================================+
+| v5.9.0                                             | Users with an outdated mobile version (such as a custom built app) might experience issues if the config.json is based on the new default.json                   |
```
jasonblais: What's an outdated mobile version? If I'm not using a custom built app, can I have an outdated mobile app?

What action does the admin have to make? It says ```if the config.json is based on the new default.json ``DisableLegacyMFA`` ``` but I don't what to do about it.

DSchalla:
```suggestion
| v5.9.0                                             | If you are using the latest `default.json` that introduced the ``DisableLegacyMfa`` configuration setting defaulting to ``true``, users might face issues when they aren't running the latest mobile application (v1.17) together with multifactor-authentication. If the setting is not defined in the config file, the default value is `false` to avoid breaking current installations. If possible, ensure all users update to the latest mobile version and change the setting to ``true`` for additional security hardening. Otherwise change the setting to `false` if you are constrained to previous versions of the mobile application of Mattermost.            |
```
@jasonblais This should make it more clear, although the wording for sure can be enhanced. This is not a breaking change by default, I've discussed it in-depth with Katie how the roll out is planned. New installations will set this flag to true by default, old ones for now to false to avoid any breaking changes. For Gitlab I am not involved so far how the configuration is set, paging @hmhealey.

DSchalla: As another note, if you are rolling out Mattermost via an Enterprise AppStore / MDM, you could be on a non up-to-date mobile version, as an example. Also if you are using a rooted phone, or build the Mattermost mobile app yourself.

This change does only affect MFA we handle internally, aka not any SSO providers. It also does not change the MFA functionality, but the way we check if a user has MFA enabled or not.

hmhealey: GitLab configuration now works as it does for a regular server for most of the config settings, so there will be nothing special here for it.

Also, users may have an outdated mobile app if they have auto updates disabled or if they're on one of the versions of Android that we dropped support for

amyblais:
@jasonblais Can you help edit the Important upgrade notes section for readability? I was going to but got confused.

jasonblais:
@amyblais Done

PR #2574, opened 2019-03-04
Upgrade: correct path step 14

JtheBAB:
Step 14 had unfortunately a wrong path. This PR fix this.

### DIFF
--- a/source/administration/upgrade.rst
+++ b/source/administration/upgrade.rst
@@ -158,7 +158,7 @@ After the server is upgraded, users might need to refresh their browsers to expe
 
       cd {install-path}/mattermost
       sudo mv plugins~/ plugins
-      sudo mv client/plugins~/ mattermost/client/plugins
+      sudo mv client/plugins~/ client/plugins
 
 Upgrading Team Edition to Enterprise Edition
 --------------------------------------------


source/administration/upgrade.rst
```diff
@@ -158,7 +158,7 @@ After the server is upgraded, users might need to refresh their browsers to expe
 
       cd {install-path}/mattermost
       sudo mv plugins~/ plugins
-      sudo mv client/plugins~/ mattermost/client/plugins
+      sudo mv client/plugins~/ client/plugins
```
jasonblais: Should we instead make this step consistent with step 8, i.e.

```
cd {install-path}/mattermost
sudo mv plugins~/ plugins
sudo mv client/plugins~/ client/plugins
```

becomes

```
sudo mv mattermost/plugins~/ mattermost/plugins
sudo mv mattermost/client/plugins~/ mattermost/client/plugins
```

OR alternatively, if this is cleaner, should we make similar changes for step 8 by cd:ing into the mattermost directory first?

JtheBAB: Sure. Up to you :)

jasonblais: I was looking for a consistent pattern in the upgrade guide, but each step does something different..some cd to `{install-path}`, some cd to `{install-path}/mattermost` and some don't cd at all..need to fix this..

So let's go with your proposal here, which is most consistent for steps 11 onward.

PR #2577, opened 2019-03-04
Add a note about pinned/flagged messages

amyblais:
Adding a reply to a community question to documentation https://forum.mattermost.org/t/limitation-about-pinned-message-and-flagging-messages/6738/4.

### DIFF
--- a/source/help/messaging/flagging-messages.rst
+++ b/source/help/messaging/flagging-messages.rst
@@ -1,7 +1,7 @@
 Flagging Messages
 -------------------
 
-Flags are a way to mark messages for follow up. Your flags are personal, and cannot be seen by other users.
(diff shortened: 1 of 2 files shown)


amyblais:
@esethna This is a quick review

source/help/messaging/flagging-messages.rst
```diff
@@ -1,7 +1,7 @@
 Flagging Messages
 -------------------
 
-Flags are a way to mark messages for follow up. Your flags are personal, and cannot be seen by other users.
+Flags are a way to mark messages for follow up. Your flags are personal, and cannot be seen by other users. There is no limit to the number of flagged posts in a channel.
```
esethna: Sorry for the nits. Flagged posts don't live in a channel so this doesn't make sense. We should say "There is no limit to the number of posts you can flag."

PR #2590, opened 2019-03-13
MM-14478 Update documentation on local image proxy and AllowUntrustedInternalConnections

hmhealey:
These are the changes we discussed to make the AllowUntrustedInternalConnections setting easier to understand. It also mentions possible issues caused by that setting and the local image proxy in more places.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-14478


source/administration/config-settings.rst
```diff
@@ -3135,13 +3135,23 @@ Enable Developer Mode
...
 
-By default, user-supplied URLs such as those used for Open Graph metadata, webhooks or slash commands will not be allowed to connect to reserved IP addresses including loopback or link-local addresses used for internal networks. Push notification and OAuth 2.0 and other URLs entered via the System Console are trusted and not affected by this setting.
+Requests that can only be configured by admins are considered trusted and will not be affected by this setting. Trusted URLs include ones used for OAuth login or sending push notifications.
 
-Separate two or more domains with spaces instead of commas, for example: ``webhooks.internal.example.com 127.0.0.1 10.0.16.0/28``.
+.. warning:: This setting is intended to prevent users located outside your local network from using the Mattermost server to request confidential data from inside your network. Care should be used when configuring this setting to prevent unintended access to your local network.
+
+Some examples of when you may want to modify this setting include:
```
jasonblais: Are there specific examples we can share on how an admin might want to modify the setting for these cases?

source/administration/config-settings.rst
```diff
@@ -3135,13 +3135,23 @@ Enable Developer Mode
...
+
+- When installing a plugin that includes its own images, such as `Matterpoll <https://github.com/matterpoll/matterpoll>`__, you will need to add the Mattermost server's domain name to this list.
```
jasonblais: Would there be an example of an internal plugin?

hmhealey: Do you mean one hosted by us? I picked Matterpoll since it's basically official, and I know it has issues with this setting.

source/administration/config-settings.rst
hmhealey: These seem pretty specific to me. Are you looking for something more like "If you have X plugin installed on Y server, add Y to this list?"

jasonblais: > I struggled to get that exactly since it depends very much on how their network is set up

Got it, if that's the case, we can keep the examples as-is. We can always update later, if admins are looking for more specific examples.

source/administration/config-settings.rst
hmhealey: Oh yeah, I can see that. At the same time, it might be nice to promote that other people are writing plugins for Mattermost as well.

PR #2594, opened 2019-03-13
Update link to share an integration

jasonblais:
This is in preparation for apps.mattermost.com launch this week.

While we're still doing a couple of bug fixes on the site in general, it's safe to link to the new form. We get fewer than 1 submission through the form per week, so impact is low.


jasonblais:
Actually, looks like the Zapier may not be hooked correctly that posts submissions to community.mattermost.com, so will investigate that and then queue for review,

jasonblais:
We did the redirection via website so no need to update docs. Closing.

PR #2601, opened 2019-03-17
Use OPTION in place of ALT for mac shortcuts

bradjcoughlin:
In testing the Mac shortcuts, I found that the following shortcuts work with the `option` key as opposed to the `command` key which is the equivalent of `alt` on Windows. More importantly, I think it makes sense to call the `option` button what it is 😄

### DIFF
--- a/source/help/messaging/keyboard-shortcuts.rst
+++ b/source/help/messaging/keyboard-shortcuts.rst
@@ -11,13 +11,13 @@ Navigation
 +----------------------------------------+----------------------------------------+------------------------------------------------------------------------------+
 | On Windows                             | On Mac                                 | Description                                                                  |
 +========================================+========================================+==============================================================================+
-| ALT+UP                                 | ALT+UP                                 | Previous channel or direct message in left hand sidebar                      |
+| ALT+UP                                 | OPTION+UP                              | Previous channel or direct message in left hand sidebar                      |
 +----------------------------------------+----------------------------------------+------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


svelle:
> In testing the Mac shortcuts, I found that the following shortcuts work with the `option` key as opposed to the `command` key which is the equivalent of `alt` on Windows. More importantly, I think it makes sense to call the `option` button what it is

tbf I have a mac keyboard here in front of me that says alt instead of option.  
Could it be that Apple gradually changed their keyboards to say option instead of alt throught the past few years?

bradjcoughlin:
@svelle when I did some quick research I saw articles saying command is the equivalent as alt but i just looked at my MacBook Air circa 2011 that has “alt” printed above “option”. Interesting. In any case, I think “option” is the predominant Mac term.

svelle:
> @svelle when I did some quick research I saw articles saying command is the equivalent as alt but i just looked at my MacBook Air circa 2011 that has “alt” printed above “option”. Interesting. In any case, I think “option” is the predominant Mac term.

I double checked the keyboard I have here and its from 2007. Also most new models have alt and option printed on them and literally the first line on google when searching for alt key mac is `On a Macintosh, the Alt key is called the Option key. `. So I think changing it to option should be fine :D

bradjcoughlin:
@svell but my 2018 MacBook has removed “alt” and replaced it with the “alt” (⌥) symbol, keeping the “option” label underneath. I see wikipedia shows the “alt” above “option” so not sure what source I was reading 😆.

PR #2606, opened 2019-03-23
Update prod-docker.rst to add Git installation step.

bcalik:
Git must be installed to run `git clone` command.

`wget` is also forced to install in the first step, but I recommend creating a pre-step to install all these requirements before installation steps begin.


amyblais:
@pichouk Would you be able to take a look if the changes made here look correct?

pichouk:
LGTM :+1:

PR #2608, opened 2019-03-26
Updates for Apache2 proxy configuration

jasonblais:
@thawn @MikeDaniel18 @prixone 

We received some proposed updates for Apache2 proxy configuration, but are not experts of it. Wondering if you were open to helping with reviews?


thawn:
@jasonblais Could you please motivate the changes?
I put some comments to the actual changes in the code.

source/install/config-ssl-http2-apache2.rst
```diff
@@ -51,6 +51,7 @@ When opened, edit it to look something like the following:
 		</Location>
 
 		# Generated by Certbot
+		SSLEngine On
```
thawn: On my system, SSL Engine On is enabled in `/etc/letsencrypt/options-ssl-apache.conf`, which is included three lines below.

MikeDaniel18:
@jasonblais Don't think I'm qualified. Apache configuration really isn't my area of expertise. Looks like you've got a contributor who's providing strong input, however. One thing to consider, though, Apache varies considerably between environments, OS's etc. etc., so suggesting something does something else, or acts in a specific way *"on my system"* is by no means a strong case for changing it.

thawn:
@MikeDaniel I fully agree, that Apache varies between systems. Therefore, I always emphasized that my recommendation is based on one system only. 
However, on that one system I did quite extensive and thorough tests, which confirmed that most of the suggested changes are redundant.

source/install/config-ssl-http2-apache2.rst
thawn: However, since this is the ssl configuration documentation, I does make sense to explicitly add `SSLEngine On`

jasonblais:
@thawn @MikeDaniel18 Excellent feedback - From my understanding, only `SSLEngine On` makes sense to specify for the SSL docs, but otherwise I'll follow up with the user who shared the updated configuration files to better understand the suggested changes. Is that correct?

thawn:
@jasonblais sounds good to me.

jasonblais:
I'll close this for now (but leave the branch open) until I hear back from the original reporter

PR #2620, opened 2019-04-03
Support for interactive ephemeral messages

jasonblais:
For https://github.com/mattermost/mattermost-webapp/pull/2351

### DIFF
--- a/source/developer/interactive-messages.rst
+++ b/source/developer/interactive-messages.rst
@@ -312,13 +312,6 @@ However, the schema for these objects is slightly different given Slack requires
 Troubleshooting
 --------------------
 
-Message buttons and menus don't show up for slash commands
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-Make sure the `response type <https://docs.mattermost.com/developer/slash-commands.html#message-type>`__ of your slash command is set to ``in_channel``, not ``ephemeral``.
-
-Ephemeral messages do not have a state, and therefore do not support interactive message buttons at this time.
-
 Message buttons and menus do not trigger an action and return a 400 error
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


hanzei:
Please note, that ephemeral messages due not support updating themself via the `update` response parameter. Only posting another ephemeral message via `ephemeral_text` is possible.

cc @levb as we talked about this.

levb:
> ephemeral messages due not support updating themself via the update response parameter. 
Correct. The semantics of processing the response were/are that if Update is set, it is used to call p.API.UpdatePost, with some special logic on top of it to protect certain metadata; if EphemeralText is set, a new simple ephemeral message was created, with the text. https://github.com/mattermost/mattermost-server/blob/abfe0a38d33825f6311b9a7ddc499d76f2e62e08/model/integration_action.go#L98

> Only posting another ephemeral message via ephemeral_text is possible.
With regards to what's possible with the use of the response struct, that is correct. To update an ephemeral message one can use a new, experimental UpdateEphemeralMessage API. This is what the Giphy plugin does. The rationale was that since ephemeral messages carry little state, updating them is tricky business, and should be left to finer tools (APIs and custom code in the plugin), at least for the time being.

jasonblais:
@levb Is the UpdateEphemeralMessage API documented at mattermost/mattermost-api-reference?

@levb @hanzei I started a draft of an FAQ for buttons/menus in ephemeral messages. Let me know if I should add other relevant information.

levb:
@jasonblais https://developers.mattermost.com/extend/plugins/server/reference/ it's in the plugin API, we do not have HTTP APIs for these (experimental). I am not sure if we need these APIs, can see it either way. If you or @hanzei think we do, please file a separate ticket.

jasonblais:
@levb Confirming that interactive ephemeral messages are only supported for plugins?

Then, I feel this doesn't really meet the use case since >90% of integrations today are not plugins.

That's okay, as it's still an improvement, but I'll update the doc and changelog entries accordingly.

jasonblais:
Or did you mean the ability to update ephemeral message itself is only for plugins, but interactive ephemeral messages work for all integration types?

levb:
> Or did you mean the ability to update ephemeral message itself is only for plugins, but interactive ephemeral messages work for all integration types?

Correct. `CreateEphemeralPost` http API accepts an entire model.Post struct with little filtering, so it should totally work for creating interactive messages, with a slack attachment property set. I did not test this. However, we did not expose the `Update/DeleteEphemeralPost` method due to a still unresolved discussion as to whether we should use separate APIs for ephemeral vs regular, or rework the main `Create/Update/DeletePost` to better support the ephemeral use case. Ditto for CommandResponse.

I did not _test_ this capability with integrations, only with a plugin.

jasonblais:
Okay, sounds like that is fine. Also updated the FAQ accordingly.

PR #2626, opened 2019-04-10
Update inAppPinCode row

ahmaddanialmohd:
Included minor detail as per discussion with Elias on the behavior after session expires.


amyblais:
@ahmaddanialmohd Can you help choose "no wrap" when you edit the doc and help make sure the table is correctly aligned/formatted? Let me know any questions,

Screenshot at Apr 10 09-11-48

hanzei:
@ahmaddanialmohd Do you have questions about the above provided feedback? https://github.com/mattermost/docs/pull/2626#issuecomment-481685488

ahmaddanialmohd:
@hanzei @amyblais Sorry missing out on this one. I think the alignment of the table is messy due to the length of the Description I updated. May I know if it makes sense to break it down into point form? For example:

-  Require users to authenticate as the owner of the phone before using the app
- Prompts for fingerprint or passcode when the app first opens and when the app has been in the background for more than 5 minutes
- Allows access to the app and the user will need to re-login if the session expires

amyblais:
@ahmaddanialmohd You can add a second line to fit all the text, I've made the changes for you as an example:
Screenshot at Apr 26 08-36-33

amyblais:
@ahmaddanialmohd Any thoughts on the above question?

ahmaddanialmohd:
Hi, @esethna and @amyblais . Sorry I missed out on this. Specific to this one:

> Do you mean that a fingerprint is required again after session expiry?

Yes, that is what I mean. It requires fingerprint and password after session expiry. What's your thought on that? Does it make sense to include that information in the table?

amyblais:
Hi @ahmaddanialmohd, maybe you can clarify the note you added to say that a fingerprint is required again after session expiry.

amyblais:
Closing for now due to inactivity, this can be re-opened when you have time to work on this.

PR #2644, opened 2019-04-18
Document for engineer onboarding timeline & expectations

jwilander:
>This is the general timeline and expectations for new engineers starting at Mattermost. It is meant to guide new engineers, help them focus on the right tasks, and set expectations to help them be successful at Mattermost. It is not a hard timeline and circumstances may shorten, lengthen or re-order the timeline.

This has already been reviewed by Corey, most of the dev leads and some other devs.

### DIFF
--- a/source/process/engineer-expectations.md
+++ b/source/process/engineer-expectations.md
@@ -2,7 +2,7 @@
 
 This is the general timeline and expectations for new engineers starting at Mattermost. It is meant to guide new engineers, help them focus on the right tasks, and set expectations to help them be successful at Mattermost. It is not a hard timeline and circumstances may shorten, lengthen or re-order the timeline.
 
-##### Week 1: Focus on environment setup and introductions
+#### Week 1: Focus on environment setup and introductions
 - Set up equipment and environment (e.g. complete the day 1 checklist)
   - Get laptop, set up developer environment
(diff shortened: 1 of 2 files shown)


jwilander:
FYI @coreyhulen

source/process/engineer-expectations.md
```diff
@@ -0,0 +1,91 @@
...
+##### Week 1: Focus on environment setup and introductions
+- Set up equipment and environment (e.g. complete the day 1 checklist)
+  - Get laptop, set up developer environment
+  - Clone enterprise repo
+  - Log in to all accounts (Gmail, GitHub, JIRA, OneLogin, Community, Community Daily)
+  - Expense equipment
+  - HR set up (benefits, health insurance, etc.)
+- Meet the team
+  - Meet with a different team member for 30min each day
+- Meet with your mentor (dev lead)
```
cpanato: The new hired will always have the dev lead as a mentor? Or we will give a change to others assume this position and also grown :)

jwilander: This is only for "mentor for the week", so the first week will always be the dev lead then we'll rotate through the next few weeks with different mentors

jwilander: Updated it to be more clear

PR #2662, opened 2019-04-28
Add typing library to pipfile packages and remove Pipfile.lock

evan-a-a:
The typing library is required for building docs.

Remove unecessary inclusion of Pipfile.lock file. Add it
to the ignore list so that it's not comitted in the future.


amyblais:
@evan-a-a Do you know the answer to the above question?

grundleborg:
@amyblais d'oh, sorry, for some reason I thought you were the submitter of the PR. @evan-a-a sorry - just wondering if you can explain a bit more what the reason is for this change, thanks.

jasonblais:
Hey @evan-a-a, thank you for your contribution!

Would you be open to sharing more about the changes here? In particular

> is there a reason for not checking in the Pipfile.lock? In my experience the convention is generally to check it in to ensure all users have the same versions of all the libraries, but wondering if there's a particular reason why this is not desirable here?

hanzei:
Hey @evan-a-a,

I'm closing this PR due to inactivity. Still, thanks for your contribution :+1:

Issue #2665, opened 2019-04-30
Extended Support Release (ESR) docs


torgeirl:
On April 14, 2019 I received the following email from Mattermost Support:
> Extended Support Releases or ESRs are intended for organizations who value stability over having the newest features and improvements, or for organizations who have a long internal testing and certification process to undergo when upgrading.
> 
> We have been maintaining 4.10 as our current ESR.  This version is coming to the end of its lifecycle and we will be implementing version 5.9 as a new ESR. Starting 7/15/2019 we will no longer be backporting fixes to 4.10.  Please upgrade your systems to 5.9 l as soon as possible to continue to receive important bug and security updates.
> 
> To receive future updates about Extended Support Releases, sign up for our mailing list here.

**tl;dr:** 5.9 is the new ESR, support for 4.10 ends on July 15.

These updates are currently not included in the documentation, and I haven't seen this announced anywhere else either (official Twitter, official blog). At an absolute minimum the docs should be updated to prevent someone from setting up a server now using 4.10.

More generally, I would encourage the team to make an ESR roadmap that details the planned ESR release cycle going forward. A good example can be found at the [Django Project version page](https://www.djangoproject.com/download/).

amyblais:
Twitter and blog announces sound good too, @wiersgallak is that something you can help coordinate with marketing or let me know if you need any help with that?

torgeirl:
My suggested changes got approved and have been merged to `master` so the immediate nature of this issue has been resolved. :+1: 

> Twitter and blog announces sound good too

@amyblais: great! Should we keep this issue open while that's been prepared or for any broader   discussion a roadmap/improved ESR comms in general, or would you like to keep that internally?

justinegeffen:
@amyblais, imo this has been addressed and resolved. Wanted to confirm with you before I close the issue though.

torgeirl:
@justinegeffen: the roadmap etc found in [current ESR docs](https://docs.mattermost.com/administration/extended-support-release.html) is exactly what I had in mind when I opened this issue. :heart_eyes:

justinegeffen:
Thanks, @torgeirl! May I close this as resolved?

torgeirl:
@justinegeffen: I assume @amyblais has any future work on this tracked elsewhere, and that this now can be closed. :+1:

PR #2667, opened 2019-04-30
Added new ESR release to extended-support-release.rst

torgeirl:
as explained in #2665 the current ESR version information is outdated


amyblais:
Hi @torgeirl,

Agree the docs should be updated, however would it make sense to add v5.9 to a new section, e.g. "What are upcoming supported ESR versions" with the start date since the v5.9 is not an ESR yet until July. What do you think?

torgeirl:
@amyblais: the email says "please upgrade your systems to 5.9 l as soon as possible", so I assumed the coming three month's is just a overlap period. It also sounds strange to have something that won't receive updates in less than three Months as the current ESR. That's why I listed them both and added the `End of Support` field.

PR #2681, opened 2019-05-03
Clean up toc

amyblais:


source/guides/administrator.rst
```diff
@@ -57,7 +58,6 @@ Learn how to host Mattermost to meet your networking requirements.
 
    /deployment/deployment.md 
    /deployment/desktop-app-deployment*
-   AppConfig for EMM Solutions </deployment/mobile-appconfig.rst>
```
jasonblais: @amyblais why did we remove this one?

source/guides/administrator.rst
```diff
@@ -190,6 +189,7 @@ Learn how to support growth within Mattermost.
    
    /deployment/scaling*
    /deployment/cluster*
+   /deployment/cluster-310*
```
jasonblais: This one we don't actually want to add to TOC because it's an old guide. It's linked to in https://docs.mattermost.com/deployment/cluster.html

source/guides/administrator.rst
amyblais: Broken link, changed it.
It's in the Mobile section too.

PR #2682, opened 2019-05-04
MM-15312 - Updating tooltip UX guidelines

asaadmahmood:
@jasonblais Have a look.
I also notice that we are not doing things properly in some places, like "Add Reaction" should be changed to "Add reaction" etc, but that can be looked into in as a separate ticket.

Ticket: https://mattermost.atlassian.net/browse/MM-15312

### DIFF
--- a/source/developer/fx-guidelines.rst
+++ b/source/developer/fx-guidelines.rst
@@ -254,7 +254,7 @@ Display: Tooltips
 Tooltips text can be title case or sentence case.
 
 - Tooltips will be title case if the tooltip is a noun, e.g. Pinned Posts, or Flagged Posts.
-- Tooltips will be sentence case if they are announcing an action, e.g. Flag for follow up, or Create new private channel.
+- Tooltips should be sentence case if they are declaring an action, e.g. "Flag for follow up", or "Create new private channel".
 
 |
 ---------------------------
--- a/source/developer/fx-guidelines.rst
+++ b/source/developer/fx-guidelines.rst
@@ -300,7 +300,7 @@ Instructions containing a link to an external resource (ie: a blog article or do
 Display: Tooltips
 ---------------------------
 
-Tooltips text can be title case or sentence case.
+Tooltip text may be title case or sentence case depending on its context.
 
 - Tooltips should be title case if the tooltip is a noun, e.g. "Pinned Posts", or "Flagged Posts".
 - Tooltips should be sentence case if they are declaring an action, e.g. "Flag for follow up", or "Create new private channel".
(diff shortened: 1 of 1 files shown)


source/developer/fx-guidelines.rst
```diff
@@ -248,6 +248,16 @@ The color of the on_hover state should be the same as in active state, with opac
 |
 ---------------------------
 
+Display: Tooltips
```
jasonblais: Propose moving below `Display: Instructions` to keep the design guidelines section in alphabetical order.

source/developer/fx-guidelines.rst
```diff
@@ -248,6 +248,16 @@ The color of the on_hover state should be the same as in active state, with opac
...
+---------------------------
+
+Tooltips text can be title case or sentence case.
```
jasonblais:
```suggestion
Tooltip text may be title case or sentence case depending on their context.
```

source/developer/fx-guidelines.rst
```diff
@@ -248,6 +248,16 @@ The color of the on_hover state should be the same as in active state, with opac
...
+
+- Tooltips will be title case if the tooltip is a noun, e.g. Pinned Posts, or Flagged Posts.
```
jasonblais:
```suggestion
- Tooltips should be title case if the tooltip is a noun, e.g. "Pinned Posts", or "Flagged Posts".
```

source/developer/fx-guidelines.rst
```diff
@@ -248,6 +248,16 @@ The color of the on_hover state should be the same as in active state, with opac
...
+- Tooltips will be sentence case if they are announcing an action, e.g. Flag for follow up, or Create new private channel.
```
jasonblais:
```suggestion
- Tooltips should be sentence case if they are declaring an action, e.g. "Flag for follow up", or "Create new private channel".
```

source/developer/fx-guidelines.rst
asaadmahmood: Shouldn't this be, depending on its context, as a "Tooltip" is singular.
Or we can use "Tooltips" and then use "their".

source/developer/fx-guidelines.rst
asaadmahmood: Done.

source/developer/fx-guidelines.rst
asaadmahmood: @jasonblais ^

jasonblais: @asaadmahmood Yes, good catch. 

```
Tooltip text may be title case or sentence case depending on its context.
```

asaadmahmood: Makes sense.

asaadmahmood:
@jasonblais Done.

PR #2683, opened 2019-05-05
Modified the /etc/security/limits.conf

johnthompson365:
In various articles I see guidance to edit the limits.conf but it always advises defining the domain as well as type, item, value as per the file guidance. Is * or the mattermost user required in domain? Does the guidance vary for Ubuntu vs other linux distributions?
https://docs.datastax.com/en/archived/cassandra/2.0/cassandra/install/installRecommendSettings.html

### DIFF
--- a/source/deployment/cluster.rst
+++ b/source/deployment/cluster.rst
@@ -96,10 +96,10 @@ Configuration Settings
 
   .. code-block:: none
 
-    soft nofile 65536
-    hard nofile 65536
-    soft nproc 8192
-    hard nproc 8192
+    * soft nofile 65536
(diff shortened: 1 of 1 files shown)


amyblais:
Merged too fast...

lieut-data:
I don' think there's an issue with this, but adding @crspeller from security for a second set of developer eyes.

crspeller:
Looks fine to me. Since we don't recommend multi tenant machines and that's not really common any more, specifying a specific user here isn't really necessary. Someone running in that situation can modify to add the MM user there instead of `*`.

PR #2708, opened 2019-05-17
Add warning when upgrading to 5.10 with read-only config.json

halfa:
I encountered this issue because we used a configmap (which is a read-only file) in Kubernetes to hold the configuration, and 5.10 refused to start.

### DIFF
--- a/source/administration/important-upgrade-notes.rst
+++ b/source/administration/important-upgrade-notes.rst
@@ -11,6 +11,8 @@ Important Upgrade Notes
 |                                                    |                                                                                                                                                                  |
 |                                                    | This change was made because ``Update.Props == nil`` unintentionally cleared all ``Props``, such as the profile picture, instead of preserving them.             |
 +----------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| v5.10.0                                            |  Following support added to store configuration in the database, it is now required for the ``config.json`` file to be writable for the server to start          |
++----------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | v5.9.0                                             | If **DisableLegacyMfa** setting in ``config.json`` is set to ``true`` and `multi-factor authentication <https://docs.mattermost.com/deployment/auth.html>`_ is   |
 |                                                    | enabled, ensure your users have upgraded to mobile app version 1.17 or later. Otherwise, users who have MFA enabled may not be able to log in successfully.      |
 |                                                    |                                                                                                                                                                  |


halfa:
Signed the CLA

source/administration/important-upgrade-notes.rst
```diff
@@ -11,6 +11,8 @@ Important Upgrade Notes
 |                                                    |                                                                                                                                                                  |
 |                                                    | This change was made because ``Update.Props == nil`` unintentionally cleared all ``Props``, such as the profile picture, instead of preserving them.             |
 +----------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| v5.10.0                                            |  Following support added to store configuration in the database, it is now required for the ``config.json`` file to be writable for the server to start          |
```
amyblais: Just a period missing at the end of the sentence.

lieut-data:
@halfa, can you provide some context into the difficulties you encountered? Your description of the problem wasn't intentional on our part -- config.json was always assumed to be writable, even if it might get clobbered by some other deployment step -- so I'm wondering if we should dig deeper here to reverse some unexpected change.

halfa:
From my experience, before 5.10 (and since 4.x at least) I had no issue running `config.json` as a configmap, the only drawback was that some settings changed at runtime in the admin panel were reset when the app was restarted, so we had to edit the configmap ourself to make it persistent. When I upgraded from 5.9.x to 5.10.0, the server plainly crashed at startup. Maybe it is linked to the bizarre way configmap files are handled by Kubernetes. In the end we created a persistent volume to hold the writable config and upgraded without issue.

lieut-data:
@halfa, interesting. Do you happen to have the crash logs you could share?

I believe the config map results in environment variables being exported for the mattermost binary to pick up, and I'm wondering if a missing setting might have explained the crash for some reason. @cpanato, any insight you might have to share?

cpanato:
@halfa @lieut-data our current community is using the configmap and we did not have issues when we did the upgrade. We mount the confimap as a volume and then when the server starts the config is there.

I will try to reproduce the issue that you are describing when i have some time next week. This week I'm out

jasonblais:
@cpanato Let me know if you've had a chance to try reproduce the issue, or if you'd prefer a ticket assigned to a sprint?

@halfa If you have any additional data to share, that would help us investigate the underlying problem

halfa:
@jasonblais sorry for the laggy feedback, here's [a gist with the expunged config I was using](https://gist.github.com/halfa/7fc94d027a95334d990baaebf7eb7f36) when the error occurred

jasonblais:
Great, thank you @halfa! FYI @lieut-data / @cpanato on the config file above

hanzei:
@cpanato Just want to remind you about this issue.

halfa:
(if this isn't reproducible and nobody except me ever encountered it, feel free to drop the PR, the error message was pretty clear and it wasn't difficult to fix)

lieut-data:
Sounds good, @halfa, and apologies that we didn't have a chance to verify the precise behaviour here. Appreciate the contribution nevertheless :)

PR #2754, opened 2019-06-04
Update config-settings.rst

wiersgallak:
Docs for: 
 - [Updated password defaults](https://github.com/mattermost/mattermost-server/pull/10844)
- [Trusted IP Headers](https://github.com/mattermost/mattermost-server/pull/10907)

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -3349,7 +3349,9 @@ Trusted Proxy IP Header
 
 Specified headers that will be checked one by one for IP addresses (order is important). All other headers are ignored.
 
-New configs after v5.12 will have this set by default to ``[]``, meaning that no header will be trusted. Configs prior to v5.12 without the config entry will have it set to ``X-Forwarded-By``, ``X-Real-Ip``.
+New configs after v5.12 will have this set by default to ``[]``, meaning that no header will be trusted. Configs prior to v5.12 without the config entry will have it set to ``X-Forwarded-By``, ``X-Real-Ip``to maintain backwards compatibility as an authority to what the client's IP address is. 
+
+We recommend setting this to explicitly define the headers when Mattermost is running without a proxy, to avoid the client sending the headers and bypassing rate limiting and/or the audit log. For environments that use a reverse proxy this problem does not exist, if the headers are set by NGINX itself. However, if X-Real-Ip is used and X-Forwarded-By is not filtered by the proxy it will essentially be overwritten.
 
 +---------------------------------------------------------------------------------------------------+
 | This feature's ``config.json`` setting is ``TrustedProxyIPHeader`` with string array input.       |
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -3351,7 +3351,7 @@ Specified headers that will be checked one by one for IP addresses (order is imp
 
 New configs after v5.12 will have this set by default to ``[]``, meaning that no header will be trusted. Configs prior to v5.12 without the config entry will have it set to ``X-Forwarded-By``, ``X-Real-Ip`` to maintain backwards compatibility as an authority to what the client's IP address is. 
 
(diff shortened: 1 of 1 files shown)


source/administration/config-settings.rst
```diff
@@ -3344,6 +3344,17 @@ This setting only takes effect if you are using the built-in server binary direc
 | This feature's ``config.json`` setting is ``"TLSMinVer": "1.2"`` with string input. |
 +-------------------------------------------------------------------------------------+
 
+Trusted Proxy IP Header
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Specified headers that will be checked one by one for IP addresses (order is important). All other headers are ignored.
+
+New configs after v5.12 will have this set by default to [], meaning that no header will be trusted. Configs prior to v5.12 without the config entry will have it set to ``X-Forwarded-By``, ``X-Real-Ip``.
```
amyblais:
```suggestion
New configs after v5.12 will have this set by default to ``[]``, meaning that no header will be trusted. Configs prior to v5.12 without the config entry will have it set to ``X-Forwarded-By``, ``X-Real-Ip``.
```

source/administration/config-settings.rst
```diff
@@ -3344,6 +3344,17 @@ This setting only takes effect if you are using the built-in server binary direc
 | This feature's ``config.json`` setting is ``"TLSMinVer": "1.2"`` with string input. |
 +-------------------------------------------------------------------------------------+
 
+Trusted Proxy IP Header
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Specified headers that will be checked one by one for IP addresses (order is important). All other headers are ignored.
+
+New configs after v5.12 will have this set by default to ``[]``, meaning that no header will be trusted. Configs prior to v5.12 without the config entry will have it set to ``X-Forwarded-By``, ``X-Real-Ip``.
```
jasonblais: Why? Are we now automatically trusting ``X-Forwarded-By``, ``X-Real-Ip`` now?

wiersgallak: Added more content based on the original PR.
@DSchalla can you please confirm that I did not lose any of the original meaning in my edits?

source/administration/config-settings.rst
```diff
@@ -3344,6 +3344,19 @@ This setting only takes effect if you are using the built-in server binary direc
...
 +-------------------------------------------------------------------------------------+
 
+Trusted Proxy IP Header
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Specified headers that will be checked one by one for IP addresses (order is important). All other headers are ignored.
+
+New configs after v5.12 will have this set by default to ``[]``, meaning that no header will be trusted. Configs prior to v5.12 without the config entry will have it set to ``X-Forwarded-By``, ``X-Real-Ip`` to maintain backwards compatibility as an authority to what the client's IP address is. 
+
+We recommend setting this to explicitly define the headers when Mattermost is running without a proxy, to avoid the client sending the headers and bypassing rate limiting and/or the audit log. For environments that use a reverse proxy this problem does not exist, if the headers are set by NGINX itself. However, if X-Real-Ip is used and X-Forwarded-By is not filtered by the proxy it will essentially be overwritten.
```
DSchalla:
```suggestion
We recommend keeping the default setting when Mattermost is running without a proxy, to avoid the client sending the headers and bypassing rate limiting and/or the audit log. For environments that use a reverse proxy this problem does not exist, if the headers are set by NGINX itself. In those environments only explicitly whitelist the header that is set by the reverse proxy and nothing else.
```

PR #2788, opened 2019-06-14
v5.13 Changelog

amyblais:


source/administration/changelog.md
```diff
@@ -4,6 +4,66 @@ This changelog summarizes updates to [Mattermost Team Edition](http://www.matter
...
+## Release v5.13 - [Quality Release](https://docs.mattermost.com/process/release-faq.html#release-overview)
+
+Release Date 2019-07-16
+
+### Compatibility
+
+#### Removed and Deprecated Features
+
+We are removing support for Internet Explorer (IE11) in Mattermost v5.16.0, which releases on October 16, 2019. Learn more in our [forum post](https://forum.mattermost.org/t/mattermost-is-dropping-support-for-internet-explorer-ie11-in-v5-16/7575).
+
```
wiersgallak: We could also add a reminder that end of life for 4.10 ESR is July 16th as well; new ESR is 5.9.

amyblais: Added.

amyblais:
Since Carlos is off, wondering if another dev could help check if there are any Database Changes, API Changes or Websocket Event Changes in v5.13 release? /cc @lieut-data @stylianosrigas

lieut-data:
@amyblais:
* Database changes: `plugins` type entries in the `Jobs` table will be purged on upgrade. This job was incorrectly configured to run every minute, spamming the table with mostly useless records. All old records will be removed on upgrade, and the job will run daily instead.
* API: There are no differences in the `api4` package between `5.12` and `5.13`.
* Websockets: There are no differences in the `wsapi` package between `5.12` and `5.13`.

PR #2790, opened 2019-06-15
Fix wrong word (personify -> impersonate)

ThiefMaster:
Apparently even CEOs and HR coordinators are people! ;)


comharris:
Good catch @ThiefMaster !

PR #2795, opened 2019-06-18
Update ldap-group-constrained-team-channel.rst

wiersgallak:
Added a note about removals.


source/deployment/ldap-group-constrained-team-channel.rst
```diff
@@ -60,6 +60,8 @@ Users are automatically removed from the team or private channel when removed fr
 
 A user can remove themselves from the team or from the private channel when it is managed by synchronized groups.  They can be added back by users who have permission to manage members for a team or private channel by using the ``/invite`` slash command or by mentioning the user in a channel.  
 
+If the user is removed from a synchronized group and later readded to the group, they can be manually added back to the team or private channel as noted above.  They will not be automatically added back by the AD/LDAP synchronization job.
```
jasonblais:
```suggestion
If the user is removed from a synchronized group and later readded to the group, they can be manually added back to the team or private channel as noted above. They will not be automatically added back by the AD/LDAP synchronization.   
```

PR #2799, opened 2019-06-18
Move NPS out of Integration Guide

jasonblais:
I'm 4/5 NPS plugin doesn't belong in the Integration's Guide because it is not about integrating other systems into Mattermost.

image

That said, I'm on the fence about whether this goes into an end user guide or admin guide. I opted for "Administration" section of the Admin Guide, but I'm 0/5 on that to be honest. 

Another option is creating a new section in TOC for telemetry and NPS survey (under Administrator's Guide), but that feels like a lot of emphasis for that topic.

Looking for thoughts.

### DIFF
--- a/source/guides/administrator.rst
+++ b/source/guides/administrator.rst
@@ -139,6 +139,7 @@ Learn how to maintain your Mattermost system.
    /administration/health-check*
    /administration/announcement-banner.rst
    /administration/bulk-export.rst
+   /integrations/net-promoter-score*
  
 Upgrade Mattermost
 ----------------------------
--- a/source/guides/administrator.rst
+++ b/source/guides/administrator.rst
@@ -136,10 +136,10 @@ Learn how to maintain your Mattermost system.
    /administration/command*
    /administration/scripts*
    /administration/statistics.md
+   /integrations/net-promoter-score*
    /administration/health-check*
    /administration/announcement-banner.rst
    /administration/bulk-export.rst
-   /integrations/net-promoter-score*
  
 Upgrade Mattermost
 ----------------------------
--- a/source/guides/integration.rst
+++ b/source/guides/integration.rst
@@ -11,7 +11,6 @@ See one of the sample integrations below or visit the `Mattermost Integrations D
    
    /integrations/jira*
    /integrations/zoom*
-   /integrations/net-promoter-score*
    /integrations/zapier*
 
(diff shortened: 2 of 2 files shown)


JtheBAB:
Why not adding a new section „prepacked plugins“ where all prepacked plugins are documented? At the moment every prepacked plugin has the documentation on the corresponding github repo. Combined into one place should help to find all information easy and fast.

jasonblais:
@JtheBAB Thanks for the suggestion! That may work, but we may be shipping end user features like DND Scheduling as pre-packaged plugins.

So I'm wondering if we'd have a single page of pre-packaged plugins, which link to other relevant docs. Let me know what you think from an administrator's perspective?

JtheBAB:
I dont understand the difference. Why can’t be all plugins in this doc? I mean you could split it one level deeper into Enduser and Admin relevant docu, what i think would make sense. Or you open a new section in Enduser and Administration Guide and fill in there the relevant docs. 

I would prefer to have everything documentation related on one place. Of course all github readme‘s needs to be migrated and at the end the github readme has only the link to the docs page.

jasonblais:
> Why can’t be all plugins in this doc

To confirm, your concern is that plugins are not documented at docs.mattermost.com? That makes sense. I thought you meant having a single section for all pre-packaged plugins.

We've done that with Zoom, which links back to docs.mattermost.com: https://github.com/mattermost/mattermost-plugin-zoom#usage-and-configuration

There was already existing documentation for plugins in the plugin repos, and the plugins themselves link back to it for issue reports, so I just updated it for now. But it may make sense to bring that documentation to docs.mattermost.com instead.

JtheBAB:
I think it is better to have all the docs at one place and not across 1000 repos :)

How you do it is up to the team. Either a new section with all documentation (Enduser and Admin) in a new section. Or split in User’s Guide (for how to use it) and Administrator’s Guide (for how to configure it). 

I guess the welcomebot would just be inside the Administrator's Guide. The Jira Plugin split by how to use it and how to configure it

jasonblais:
Thanks @JtheBAB, good feedback, really appreciate sharing it 👍

jasonblais:
@wiersgallak Thanks! Good suggestion. Moved under "statistics" since NPS is semi-related to user activity/health whereas "health check" is more server activity/health.

PR #2816, opened 2019-06-25
Added clarification on not supporting app wrapping

icelander:
Added clarification that we don't support app wrapping, and suggestion that users contact their app wrapping provider to see if they offer EMM.

Signed-off-by: Paul Rothrock <paul@movetoiceland.com>

### DIFF
--- a/source/mobile/mobile-overview.rst
+++ b/source/mobile/mobile-overview.rst
@@ -32,6 +32,6 @@ To meet other security and compliance policies, you can :doc:`deploy the Matterm
 This option is recommended for organizations that typically use EMM solutions to deploy mobile apps to meet security and compliance policies.
 
 .. note::
-    App wrapping via an EMM provider is not supported by Mattermost. The wrapping process requires changes to the underlying app that cannot be maintained and are untested. Instead, we recommend using an EMM provider that supports Managed App Configuration.  
+    App wrapping via an EMM provider is untested, and not supported by Mattermost. Instead, we recommend using an EMM provider that supports Managed App Configuration.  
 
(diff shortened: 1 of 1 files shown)


source/mobile/mobile-overview.rst
```diff
@@ -32,3 +32,8 @@ To meet other security and compliance policies, you can :doc:`deploy the Matterm
 This option is recommended for:
 
 1. Organizations that typically use EMM solutions to deploy mobile apps to meet security and compliance policies
+
+EMM vs. App Wrapping
+--------------------
+
+Using an app wrapper, like the `Citrix MDX Toolkit <https://docs.citrix.com/en-us/mdx-toolkit/about-mdx-toolkit.html>`_ is *not supported by Mattermost* because it injects code into the app that we cannot control. Many app wrapping toolkits also provide EMM, so contact your provider for more information about how to deploy Mattermost with EMM.
```
amyblais:
```suggestion
Using an app wrapper, like the `Citrix MDX Toolkit <https://docs.citrix.com/en-us/mdx-toolkit/about-mdx-toolkit.html>`_, is not supported by Mattermost because it injects code into the app that we cannot control. Many app wrapping toolkits also provide EMM, so contact your provider for more information about how to deploy Mattermost with EMM.
```

lfbrock: @esethna - can you review the line on why we don't support it? I think it's untested and some EMM solutions may require changes to the app to make wrapping work. 

Also "EMM" is just a general term and applies to both methods of deploying, so we should update to App Wrapping vs Managed App Config and maybe add some explanation on what each is. 

Citrix is an example of an EMM provider, and I believe they have both [app wrapping](https://docs.citrix.com/en-us/mdx-toolkit/about-mdx-toolkit.html) and [managed app config](https://docs.citrix.com/en-us/xenmobile/server/policies/app-configuration-policy.html).

esethna: Updated and committed to this branch if you want to review.

PR #2823, opened 2019-06-29
Fix link to Enterprise upgrade feature comparison chart

flaki:
Apparently the link to the feature comparison of Team/Enterprise editions are broken in the docs, it took me a good 5 minutes to find the correct link on the site.


flaki:
Signed CLA.

PR #2840, opened 2019-07-13
Add Feature Documentation Template

tapaswenipathak:


source/administration/feature-documentation-template.rst
```diff
@@ -0,0 +1,37 @@
...
+This template defines how a feature should be documented improving presentation
+and retrieval of information of product features. The document is user focused.
+
+Enabling The Feature
+--------------------
+
+Write how the user can enable the feature. The user should utilize the information
+present about the feature and run it.
+
+Defining Use Case
```
wiersgallak: Wondering why Use Cases would come after enabling the feature?  Seems like use case would be first, describing to the user why they would want to use the feature.

source/administration/feature-documentation-template.rst
```diff
@@ -0,0 +1,37 @@
+Feature Documentation Template
+===============================
+
+
+This template defines how a feature should be documented improving presentation
+and retrieval of information of product features. The document is user focused.
+
+Enabling The Feature
```
wiersgallak: We should also include any prerequisites as well.  Do you think that would be it's own section or part of the "Enabling the feature" section?

wiersgallak:
Closing this PR per conversation with Tapasweni Pathak, that this project likely involves more input and collaboration with our internal team as we work through this process.

PR #2841, opened 2019-07-13
Add Documentation Template

tapaswenipathak:


wiersgallak:
@tapaswenipathak Per our conversation this morning, thank you again for your work on this.  We recognize that this project was a bit vague and think it is best to have someone a little more immersed in our processes help build out this documentation template so that we can ensure it meets all our requirements and we are able to successfully incorporate it into our processes.

tapaswenipathak:
@wiersgallak As discussed I would suggest if you can write on the ticket as well, just for people who might think about writing for the ticket.

tapaswenipathak:
@wiersgallak As discussed I would suggest if you can write _internal_ on the ticket as well, just for people who might think about writing for the ticket.

PR #2846, opened 2019-07-15

amyblais:
Customers are reporting this issue but it seems to be expected.


lfbrock:
@jasonblais @wiersgallak @esethna - not sure which team made this change, but it seems to be "expected" from a dev perspective but not from a customer perspective. Can whoever's team made the change dig into if we need to take any action beyond updating documentation here? 

(ie what is the impact to customers, who does this affect, what is the recommended action -- is there a way for them to maintain the setting they had previously, or is this a breaking change so they can't get the old behaviour back. If so, is that something that will cause problems for customers, or are they okay following whatever our recommended action is).

esethna:
Following up with apps team, sounds like it was introduced with post metadata in v5.8, but only surfacing now due to backwards compatibility removal.

source/administration/important-upgrade-notes.rst
```diff
@@ -9,6 +9,8 @@ Important Upgrade Notes
 +====================================================+==================================================================================================================================================================+
 | v5.12.0                                            | If your plugin uses the ``DeleteEphemeralMessage`` plugin API, update it to accept a ``postId string`` parameter.                                                |
 |                                                    | See `documentation <https://developers.mattermost.com/extend/plugins/server/reference/#API.DeleteEphemeralPost>`_ to learn more.                                 |
+|                                                    +------------------------------------------------------------------------------------------------------------------------------------------------------------------+                               
+|                                                    | Image links and Youtube previews do not display unless **System Console > Enable Link Previews** is enabled.                                                     |
```
esethna: Propose: 

```
Image link and Youtube previews do not display unless **System Console > Enable Link Previews** is enabled. Please ensure that your Mattermost server is connected to the internet and has network access to the websites from which previews are expected to appear. [Learn more here](https://docs.mattermost.com/administration/config-settings.html#enable-link-previews).
```

source/administration/changelog.md
```diff
@@ -25,6 +25,7 @@ Mattermost v5.12.0 contains low to medium level security fixes. [Upgrading](http
 ### Breaking changes since last release
 
  - If your plugin uses the ``DeleteEphemeralMessage`` plugin API, update it to accept a ``postId string`` parameter. See [documentation](https://developers.mattermost.com/extend/plugins/server/reference/#API.DeleteEphemeralPost) to learn more.
+ - Image links and Youtube previews do not display unless **System Console > Enable Link Previews** is enabled.
```
esethna: Same text as below

PR #2862, opened 2019-07-19
Some updates to changelog

jasonblais:
@amyblais propose for upcoming fixes, we clarify what exactly is the consequence for users or admins, so an admin can decide whether they should upgrade or not.

I've also added the 5.13 bug fixes as known issues in 5.12, since it's not clear that they are also issues in 5.12


source/administration/changelog.md
```diff
@@ -223,6 +223,10 @@ Multiple setting options were added to `config.json`. Below is a list of the add
  - Added ``KVCompareAndSet(key string, old []byte, new []byte)`` to Plugin API to add support for transactional semantics with KV Store in plugin framework.
 
 ### Known Issues
+  - Channels with large message history may have performance issues. This is scheduled to be resolved in v5.13.1.
```
jasonblais: Nevermind, these were added to 5.12.5 so maybe it's not a big deal to add as known issues. Will let you decide. For some reason I thought they were in 5.13 only

PR #2874, opened 2019-07-29
Clarified that an API call needs to be made for ephemeral posts

aaronrothschild:
This was prompted by a conversation in the forum: https://forum.mattermost.org/t/response-from-interactive-button-post-is-ignored/7757/6

### DIFF
--- a/source/developer/interactive-messages.rst
+++ b/source/developer/interactive-messages.rst
@@ -120,7 +120,7 @@ The following payload gives an example that uses message menus.
     ]
   }
 
-The integration can respond with an update to the original post, or with an ephemeral message:
+The integration can then perform a separate API call with an update to the original post, or with an ephemeral message:
 
 .. code-block:: text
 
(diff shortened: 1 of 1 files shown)


aaronrothschild:
@lieut-data Could you take a look and perhaps add some more context to when the response bodies should be considered.  I don't have enough background on this to add much to the documentation.  This PR is in response to a community member encountering some frustrations while building a bot.

lieut-data:
@aaronrothschild, apologies for the delay on this -- I too need to develop some background knowledge to be helpful here. It's on my radar, but I'm likely not going to get to that this sprint.

lieut-data:
>     "POSTs from interactive controls", "No - Response body is ignored"

Thanks for your patience on this one, @aaronrothschild. It's not obvious to me why we don't just /start/ supporting parsing the response body in this case. In the reported case, at least, it seems like this was expected, and certainly seems to fit within the larger user experience.

justinegeffen:
Hi @aaronrothschild. I wanted to find out whether you've had a chance to review the feedback and if you need any other reviews for this PR. :)

justinegeffen:
Hey @aaronrothschild, following up on this to find out whether I can assist with anything to get this PR merged. :)

PR #2885, opened 2019-08-03
Update config-tls-mattermost.html

variablenix:
Add a note about it being required to set `TLS` to **None** when using a reverse proxy even when `Use Let's Encrypt` and `Forward port 80 to 443` are already set to **false**; otherwise user's will likely run into a message similar to `Client sent an HTTP request to an HTTPS server.` that will no longer make it possible to simply specify the Mattermost domain without a port number such as `8065`.

Context: See https://forum.mattermost.org/t/apache-client-sent-an-http-request-to-an-https-server/7763

### DIFF
--- a/source/install/config-tls-mattermost.rst
+++ b/source/install/config-tls-mattermost.rst
@@ -20,6 +20,9 @@ The easiest option is to set up TLS on the Mattermost Server, but if you expect 
 
 3. Install the security certificate. You can use `Let's Encrypt <https://letsencrypt.org/>`__ to automatically install and setup the certificate, or you can specify your own certificate.
 
+.. note::
+   When using NGINX or Apache as a reverse proxy ensure ``TLS`` is set to **None** otherwise a message similar to `Client sent an HTTP request to an HTTPS server.` may be encountered. This is required when options `Forward port 80 to 443` and `Use Let’s Encrypt` are set to **false**.
+
   **To use a Let's Encrypt certificate**:
 
     The certificate is retrieved the first time that a client tries to connect to the Mattermost server. Certificates are retrieved for any hostname a client tries to reach the server at.


jasonblais:
Thank you @variablenix for the contribution! Would you be open to signing the CLA referenced above? 

This is a standard procedure for many open source projects. Please let us know if you have any questions.

variablenix:
@jasonblais sure thing. I've gone ahead and signed the referenced CLA. Thanks!

jasonblais:
Thanks @variablenix!

source/install/config-tls-mattermost.rst
```diff
@@ -20,6 +20,9 @@ The easiest option is to set up TLS on the Mattermost Server, but if you expect
 
 3. Install the security certificate. You can use `Let's Encrypt <https://letsencrypt.org/>`__ to automatically install and setup the certificate, or you can specify your own certificate.
 
+.. note::
+   When using NGINX or Apache as a reverse proxy ensure ``TLS`` is set to **None** otherwise a message similar to `Client sent an HTTP request to an HTTPS server.` may be encountered. This is required when options `Forward port 80 to 443` and `Use Let’s Encrypt` are set to **false**.
```
amyblais:
```suggestion
   When using NGINX or Apache as a reverse proxy, ensure ``TLS`` is set to **None**. Otherwise a message similar to "Client sent an HTTP request to an HTTPS server." may be encountered. This is required when options ``Forward port 80 to 443`` and ``Use Let’s Encrypt`` are set to **false**.
```

crspeller:
@variablenix This isn't entirely true. You could setup TLS between your reverse proxy and the MM server if you wanted to. And the second part about this being required is not true because you can setup your own certs instead of getting them from Let's Encrypt.

variablenix:
@crspeller The interesting thing is I do have Let's Encrypt certs specified in the `TLS Certificate File` and `TLS Key File` fields, but was running into the described snag with the MM host not resolving without specifying a port. Switching `Connection Security` back to TLS and saving so far seems to not exhibit the reported behavior. I'm testing this from a completely different box that has not visited the MM domain before. While TLS is enabled, both options **Forward port 80 to 443** and **Use Let's Encrypt** are still disabled. So far so good, but I did have it set up this way prior to the issue occurring, so I'm not entirely sure why it seems fine at this time.

Is it then safe to say (or expected behavior) that when using MM with a reverse proxy MM should **not** be running into any trouble with TLS enabled while the two security options are disabled? I will do some further testing to validate a few more things.

crspeller:
@variablenix Correct. If you have specified valid certificates and TLS is enabled you should be able to connect directly to the app server. 

Not sure about your port issue. What port do you have specified in the config? You should be able to connect using that port. (keep in mind this is a direct connection to the app server, not though the proxy, if you want connect though the proxy it has to be configured to correctly connect to the app server)

jasonblais:
Hi @variablenix, touching base on the latest question from Chris, and if you had any questions on it?

> Not sure about your port issue. What port do you have specified in the config? You should be able to connect using that port. (keep in mind this is a direct connection to the app server, not though the proxy, if you want connect though the proxy it has to be configured to correctly connect to the app server)

variablenix:
Hi @jasonblais,

Interestingly on my instance after updating to v5.14.0, I ended up up hitting the same message after the upgrade, **Client sent an HTTP request to an HTTPS server** with options _Let's Encrypt_ and _Forward port 80 to 443:_ disabled. According to Chris it is expected to work without specifying the port when you have TLS enabled in the Console settings. At the time I had TLS enabled following Chris' explanation, but the other two mentioned options disabled/false, yet after the upgrade I still ran into the message. Before the upgrade I tested the issue by restarting the Mattermost instance, but it continued to work even with a private/incognito window. By "not working" I mean not needing to specify the Mattermost port 8065 (in a reverse proxy setup).

The solution for me at least was to again disable the TLS option. I'm not sure why the expected behavior was different than what Chris confirmed -- in that Mattermost will still continue to work with a reverse proxy with TLS enabled (rather than hitting HTTP to HTTPS server message). In any case, I think it seems a bit redundant to have TLS enabled within Mattermost since the TLS side of things would be taken care of by the reverse proxy configuration.

Maybe the PR for this can be reworded in a way that states something like, "if you run into this message with a reverse proxy setup then try setting TLS to None in the Console to let the backend web server take care of TLS negotiations". Thoughts?

jasonblais:
Thanks @variablenix for the response!

@crspeller I'll let you chime in as you're more knowledgeable in this area than I am.

crspeller:
@jasonblais @variablenix If you are using the reverse proxy and you don't want TLS between the mm server and the proxy, then all the TLS related settings should be disabled.  However the statement in the PR still doesn't apply to the case where you might want TLS between your proxy and the mm server.
In your case I suspect you have the proxy and mm server running on the same box. So not specifying the port sends you though the proxy while bypassing it with a specified port number will reach the app server directly. Which creates the behavior described assuming the proxy can't properly connect to the app server when it has TLS enabled for some reason.

variablenix:
Indeed in this setup both Mattermost and the reverse proxy are on the same box. As it seems like this is expected behavior with respect to TLS, I think it should be fine to close out this PR. Thanks for all the clarification, @crspeller ! Cheers

PR #2939, opened 2019-08-21
Clarify client needs to reach the image proxy + code formatting update

jasonblais:
For context, see https://community.mattermost.com/core/pl/j8wgjni6if839ycj9a7hrzoyhc

### DIFF
--- a/source/administration/image-proxy.rst
+++ b/source/administration/image-proxy.rst
@@ -11,7 +11,7 @@ third-party servers.
 Proxy servers also provide a layer of caching, and can be made faster and more reliable than third-party sites. This caching 
 also helps preserve posts by protecting them from dead images.
 
-When enabled, the image proxy needs to be publicly accessible to the Mattermost client, which requests the image directly from the image proxy.
+When enabled, the image proxy needs to be publicly accessible to both the Mattermost client and server.
 
 You may alternatively use `atmos/camo <https://github.com/atmos/camo>`_ http proxy to route images through SSL:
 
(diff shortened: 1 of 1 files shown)


source/administration/image-proxy.rst
```diff
@@ -38,7 +40,11 @@ configuration in **System Console > Files > Storage** in prior versions or **Sys
 
 .. image:: ../images/image-proxy.png
 
-The URL will be replaced with something similar to the following: https://image-proxy.mattermost.com/d7b4022717e8d015440cd70183b81196298b9453/687474703a2f2f692e726564642e69742f36636f687964636b6b363530312e6a7067 (See `https://github.com/atmos/camo <https://github.com/atmos/camo>`__).
+The URL will be replaced with something similar to the following (ee `https://github.com/atmos/camo <https://github.com/atmos/camo>`__ for details):
```
amyblais:
```suggestion
The URL will be replaced with something similar to the following (see `https://github.com/atmos/camo <https://github.com/atmos/camo>`__ for details):
```

jasonblais: @amyblais Thanks! Updated

source/administration/image-proxy.rst
```diff
@@ -11,6 +11,8 @@ third-party servers.
 Proxy servers also provide a layer of caching, and can be made faster and more reliable than third-party sites. This caching 
 also helps preserve posts by protecting them from dead images.
 
+When enabled, the image proxy needs to be publicly accessible to the Mattermost client, which requests the image directly from the image proxy.
```
hmhealey: From a followup to that conversation, we should probably mention that it needs to be accessible by both clients and the server.

```suggestion
When enabled, the image proxy needs to be publicly accessible to both the Mattermost client and server.
```

I'm not sure how to make the second part of the sentence work if we mention both of those though

hanzei:
@esethna Would you please review this when you have the time?

PR #2941, opened 2019-08-21
Slash command response url updated to match developer docs

sadohert:
A customer was confused by the "not-yet-supported" string in our docs.  I noticed we had something more understandable in our developer-docs version of this content.  I simply copy-pasted this over into our normal docs repol


sadohert:
Sorry, ignore this.  COmbined some historical junk in my local clone.

PR #2947, opened 2019-08-24
Replace New York time with Toronto time

jasonblais:
We first use New York as an example timezone reference, but then in the next paragraph say we should use Toronto instead of New York given we have a mailing address there.

So, changing the example to Toronto

@comharris please review/confirm, @amyblais for editor review


jasonblais:
@amyblais I'll need one approver from the MM team before I can merge 🙏

PR #2952, opened 2019-08-27
Clear out Interactive Messages Buttons in example

hanzei:
See https://community.mattermost.com/core/pl/6gukujq85ifzxn8sgek6gj78bc


hanzei:
There is a notice for this in https://docs.mattermost.com/administration/important-upgrade-notes.html
Screenshot from 2019-08-27 14-57-12

hanzei:
Interactive Buttons are part of `Props`. Hence, they are effected by this change. While I see that the text is not ideal, I would argue that it's not worth to update it by now. 1/5

hanzei:
> IMO this was a pretty substantial breaking change to how interactive buttons behave, and we should make that clear.

Updated the notice text.

> Would have to investigate it more, including how it compares against how Slack's message buttons work today (they've migrated to use chat.Update method to update posts).

That seams importent to look into. Would you please open a ticket for this?

hanzei:
Thanks for summing this up. This looks fine to me :+1: 

> Making the attachments disappear is technically possible now, but if the integration action wants to preserve the other fields (e.g. username override, icon override), it must explicitly re-send them. This is very hard in a stateless service and wasn't the intended workflow.

One nit pick here: There are some preserved fields currently (like `override_username`) that are copied when a new list of props is provided.

PR #2958, opened 2019-08-30
Update hiring process to better reflect current practice.

grundleborg:


source/process/developer.rst
```diff
@@ -75,10 +75,11 @@ Both declined and accepted candidates will be invited to share feedback on their
...
-- **Video call - CEO interview** - Finally, candidates will have a 45-minute interview with our CEO.
+- **Video call - Discussion of Audition and Soft Skills** - On completion of the Audition, candidates will be invited to schedule a 60-minute interview with a core committer to discuss the experience and outcomes of participating in the audition and further assess soft skills.
+- **Video call - CTO interview** - Candidates are invited to a 45-minute interview with our CTO and co-creator of the Mattermost open source project. The interview may include technical questions along with a discussion of either past work or results of the audition, the candidate's interests, their career aspirations, and how being a core committer at Mattermost could align with those interests and aspirations.
```
grundleborg: @coreyhulen does this accurately reflect what you do in the process now?

source/process/developer.rst
```diff
@@ -75,10 +75,11 @@ Both declined and accepted candidates will be invited to share feedback on their
 - **Application/Email - Review of code samples** - In `your application <https://jobs.lever.co/mattermost/>`__ we welcome you to include links to any public open source work you've done, including GitHub, GitLab or BitBucket profiles. If you'd like to share privately we can send you our ID for GitHub, GitLab or BitBucket, or we can review code you send in via email. If you really want to impress us, `make a contribution to our open source project by completing a Help Wanted ticket. <../developer/contribution-guide.html>`__
 - **Email - Follow-up questions** - If your code samples look good, we'll email you some questions specific to the role.
```
lieut-data: I find a lot of candidates aren't submitting code samples -- should we emphasize this point, or relax it?

source/process/developer.rst
coreyhulen: yup, looks good

source/process/developer.rst
grundleborg: @lieut-data good to hear this - suggests I'm the not the only person who thinks this isn't particularly realistic here. I'll update to be more in line with what *actually* happens in this stage too.

grundleborg:
@lieut-data added an extra commit fixing the mention of "code samples".

PR #2960, opened 2019-08-30
Remove color and footer known issues

nfriend:
Updates [this section of the developer documentation](https://docs.mattermost.com/developer/message-attachments.html?highlight=footer#known-issues) to align with the changes proposed in https://github.com/mattermost/mattermost-webapp/pull/3569.

Specifically:
1. Removes the note that the `footer` and `footer_icon` properties are not supported
1. Removes the note that the `good`/`warning`/`danger` `color` values are not supported

As a side note, <span>#</span>2 above was already fixed before https://github.com/mattermost/mattermost-webapp/pull/3569.


hanzei:
Would you be open on adding short descriptions about `footer` and `footer_icon` to the list of supported fields?

nfriend:
@hanzei Good catch! I've updated my branch with some details about the `footer` and `footer_icon` fields as well as a screenshot.

source/developer/message-attachments.rst
```diff
@@ -124,12 +124,20 @@ And here is how it renders in Mattermost:
 
 .. image:: ../images/attachments-example.png
 
+Footer
+~~~~~~~
+
+``footer``: An optional line of text that will be displayed at the bottom of the attachment.
```
hanzei: Please add a note that this will be truncated if it's longer then 300 characters.

nfriend: @hanzei Done 👍

PR #2968, opened 2019-09-05
update migrating.rst

svelle:
update slack migration steps with note to avoid slack migration issues as per ticket https://github.com/mattermost/mattermost-server/issues/11927

### DIFF
--- a/source/administration/migrating.rst
+++ b/source/administration/migrating.rst
@@ -190,6 +190,10 @@ Migrating from Slack using the Mattermost CLI
 2. Run the following Mattermost CLI command, with the name of a team you have already created:
 
    ``$ mattermost import slack team_name /path/to/your-slack-export.zip``
+   
+.. note:: Please make sure to run the import command as the mattermost user and from the directory you have installed the Mattermost server to (e.g. /opt/mattermost) to avoid issues with file ownership and paths for message attachments.
+
+
 
 Using the Imported Team
 ````````````````````````
--- a/source/administration/migrating.rst
+++ b/source/administration/migrating.rst
@@ -191,7 +191,8 @@ Migrating from Slack using the Mattermost CLI
 
    ``$ mattermost import slack team_name /path/to/your-slack-export.zip``
    
-.. note:: Please make sure to run the import command as the mattermost user and from the directory you have installed the Mattermost server to (e.g. /opt/mattermost) to avoid issues with file ownership and paths for message attachments.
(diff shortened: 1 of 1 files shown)


source/administration/migrating.rst
```diff
@@ -190,6 +190,10 @@ Migrating from Slack using the Mattermost CLI
 2. Run the following Mattermost CLI command, with the name of a team you have already created:
 
    ``$ mattermost import slack team_name /path/to/your-slack-export.zip``
+   
+.. note:: Please make sure to run the import command as the mattermost user and from the directory you have installed the Mattermost server to (e.g. /opt/mattermost) to avoid issues with file ownership and paths for message attachments.
```
jasonblais: Propose (for consistency with https://docs.mattermost.com/administration/command-line-tools.html#using-the-cli)

```
.. note:: To run the CLI command, you must be in the directory that contains the Mattermost executable. On a default install of Mattermost, the directory is ``/opt/mattermost/bin``. Also, if you followed our `installation process <../guides/administrator.html#installing-mattermost>`__, you must run the commands as the user *mattermost*. The name of the executable is ``mattermost``.
```

hanzei: @svelle Gentle reminder about this comment

svelle: Thanks for the reminder. Actually I don't understand why it is recommended and I'd like to propose to also change it in the command line tools docs as well. 
Running the CLI from the bin folder causes several errors on startup due to some paths working relative in the config. (for example the /client/plugins/ directory will not be found when starting from the cli). These are confusing to a lot of sysadmins running the server and we got quite a few support requests in the past because of this.

Also in the service files we provide in the install documentation we set the `WorkingDirectory` for the Mattermost server to be `/opt/mattermost` and not `/opt/mattermost/bin`.

jasonblais: Would the recommendation then be to say

```
.. note:: To run the CLI command, you must be in the directory that contains the Mattermost executable. On a default install of Mattermost, the directory is ``/opt/mattermost``. Also, if you followed our `installation process <../guides/administrator.html#installing-mattermost>`__, you must run the command as the user *mattermost*. The name of the executable is ``mattermost``.
```
^where we reference `/opt/mattermost` instead?

svelle: Hey Jason,
not quite right. The executable still sits in `/opt/mattermost/bin` however the run directory should be `/opt/mattermost` so when you're in `/opt/mattermost` execute `bin/mattermost <command>` to run the desired command. Otherwise there'll be a bunch of errors in the CLI that could be avoided by choosing the correct rundir.

jasonblais: Thanks @svelle, would you be open to adjusting as such?

svelle: @jasonblais just did, could you review this again?

PR #2986, opened 2019-09-13
Plugin command corrections

MatthewDorner:
There are inaccuracies in the documentation for `mattermost plugin` commands. It sometimes shows `mattermost plugins` plural, when the actual command is singular `plugin`. Then, some of the command examples show tar.gz filenames being used as arguments when the correct argument for those commands is the plugin ID, not the filename. (These are the commands that apply to a plugin that's already installed. The filename is only used for `plugin add`.)

I updated the examples by copying them from mattermost-server/cmd/mattermost/commands/plugin.go since the example text is in both places and since the version in mattermost-server appears correct.


MatthewDorner:
Do I need to do something to make the CircleCI build pass?

hanzei:
@MatthewDorner Would you please merge `master` into your branch? That should fix the CI.

PR #3015, opened 2019-09-26
Update changelog.md

amyblais:
Adding notes about upcoming dot releases to make users aware.

### DIFF
--- a/source/administration/changelog.md
+++ b/source/administration/changelog.md
@@ -6,7 +6,13 @@ Also see [changelog in progress](http://bit.ly/2nK3cVf) for the next release.
 
 ## Release v5.15 - [Quality Release](https://docs.mattermost.com/process/release-faq.html#release-overview)
 
-**Release Day: 2019-09-16**
+- **v5.15.2, release day TBD**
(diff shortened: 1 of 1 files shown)


source/administration/changelog.md
```diff
@@ -77,6 +83,11 @@ Mattermost v5.15.0 contains low level security fixes. [Upgrading](http://docs.ma
 
 ## Release v5.14 - [Feature Release](https://docs.mattermost.com/process/release-faq.html#release-overview)
 
+- **v5.14.5, released 2019-10-16**
```
lieut-data: Will we actually have a `v5.14.5` or merging the two together? The order here seems off as well if we've released one before the other.

amyblais: @lieut-data The notification issue is caused by the upcoming v5.16 / desktop v4.3 release so the backports for that issue won't be released before October 16th.

The two issues you're working on are existing issues so I'm assuming we'll have a dot release for those before v5.16 release?

amyblais: Probably need clarification from @esethna if we can release the notification issue backports before October 16th.

lieut-data: Ah, that makes sense!

esethna: Checking with dev. I think we may be able to release them earlier than v4.3 assuming there are no compatibility issues with notifications if using the backported server versions with v4.2

amyblais: From Dean's notes, it seems it's ok to have all of these in the same dot release, so I made that change here. The changelog can be edited later if anything changes

PR #3029, opened 2019-10-04
Add license step to k8s install instructions

jwilander:
Adding this quickly because I forgot to include it before.

### DIFF
--- a/source/install/install-kubernetes-mattermost.rst
+++ b/source/install/install-kubernetes-mattermost.rst
@@ -25,7 +25,7 @@ Apply it with:
 
 .. code-block:: sh
 
-  $ kubectl apply -n mattermost -f /path/to/mattermost-license-secret.yaml
+  $ kubectl apply -f /path/to/mattermost-license-secret.yaml
 
 **2. Create your installation manifest file**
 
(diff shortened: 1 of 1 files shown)


source/install/install-kubernetes-mattermost.rst
```diff
@@ -5,7 +5,29 @@ Deploy a Mattermost Installation
 
 Deploying a Mattermost installation is very simple and just requires creating and applying a single manifest.
 
-**1. Create your installation manifest file**
+**1. (Enterprise Only) Create your Mattermost license secret**
```
justinegeffen: Could this be an independent point (not numbered) to clarify that it's only applicable to Enterprise customers?

justinegeffen:
Other than my suggestion above, this looks good to me.

source/install/install-kubernetes-mattermost.rst
amyblais:
```suggestion
**1. (Enterprise only) Create your Mattermost license secret**
```

jwilander: Not exactly sure how that would work but feel free to make a PR afterwards to update it :) I'm just going to merge this as is to get something out there

PR #3048, opened 2019-10-11
Update searching.rst

justinegeffen:
https://github.com/mattermost/mattermost-server/pull/11196


justinegeffen:
I wanted to confirm whether the statement under "PostgreSQL" ("Terms containing a dash return incorrect results as dashes are ignored in the search engine.") is still correct? It seems to contradict the addition of the "-" feature but I may be interpreting it incorrectly.

wiersgallak:
@justinegeffen I just tested the exclusion on a server with Postgres and can confirm it is working as expected.  I also tested searching returning the word "sign-up" and there still seems to be some unexpected behavior - ie: "sign up" returns when search term is "sign-up".  Probably best to keep that section in the docs.

PR #3056, opened 2019-10-12
Document API perf monitoring metrics (for 5.18)

jasonblais:
@jespino Wondering if you'd be open to help add the two new metrics (DB store time, and API time) to our sample Grafana dashboard? https://github.com/mattermost/docs/blob/master/source/samples/grafana-dashboards/server-perf-monitoring-set2.json

We should also consider merging https://github.com/mattermost/docs/blob/master/source/samples/grafana-dashboards/server-perf-monitoring-set1.json with https://github.com/mattermost/docs/blob/master/source/samples/grafana-dashboards/server-perf-monitoring-set2.json because the two dashboards have different metrics, but it's not clear why not just have them in one (I made the decision 3 years ago to separate the metrics into two based on the importance of metrics, but disagree with that decision now in favor of consolidation :) )

### DIFF
--- a/source/deployment/metrics.rst
+++ b/source/deployment/metrics.rst
@@ -115,6 +115,10 @@ Custom Mattermost Metrics
 
 The following is a list of custom Mattermost metrics that can be used to monitor your system's performance:
 
+API Metrics:
+
+    - ``mattermost_api_total``: The total time in seconds to execute a given API handler.
+
 Caching Metrics:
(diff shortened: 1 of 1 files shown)


jespino:
@jasonblais sure I'm open to that, and I think should be easy because we already have it in our grafana, so, it should be easy to export and add to the existing grafana json

PR #3059, opened 2019-10-13
Added user ID location and actions on grid

thefactremains:
Provided information on where to find the user ID (now visible in users list as well). I also added details of the read only list of teams each user is a member of and the new "add teams" button on user profile page.


wiersgallak:
Closing

PR #3067, opened 2019-10-15
v5.16 docs

esethna:
IE11 removal, desktop away status, MSI installer to beta, plus miscellaneous others


source/install/desktop.rst
```diff
@@ -28,24 +28,28 @@ Windows 10+, Windows 8.1+, Windows 7+
 
 1. Download latest version of the Mattermost desktop app:
```
esethna: @amyblais can we please verify all these download links once final is cut?

PR #3068, opened 2019-10-16
Update to zoom account info

aaronrothschild:
Clarified the use case of having a mix of free and paid accounts


aaronrothschild:
Moved docs, this isn't needed anymore

PR #3089, opened 2019-10-18
Remove desktop changelog from docs

amyblais:


amyblais:
@esethna We keep the one for mobile on the mobile app repo. Would it then make sense to also move the mobile changelog to the docs?

esethna:
Works for me ^

Issue #3065, opened 2019-10-15
Request for Documentation: Document supported ES versions.


matterdoc:
Mattermost user `george` from https://community-daily.mattermost.com has requested the following be documented:

```
Great, thanks @george 
I think we should not this down in the docs, if we have specific version which we tested compatibility for.
```

See the original post [here](https://community-daily.mattermost.com/_redirect/pl/ccfpn8bk73y4bep6gg6tx66q8a).

_This issue was generated from [Mattermost](https://mattermost.com) using the [Doc Up](https://github.com/jwilander/mattermost-plugin-docup) plugin._

svelle:
The page were we should document it: https://docs.mattermost.com/deployment/elasticsearch.html

PR #3103, opened 2019-10-23
Update elasticsearch.rst

justinegeffen:
https://github.com/mattermost/docs/issues/3065

### DIFF
--- a/source/deployment/elasticsearch.rst
+++ b/source/deployment/elasticsearch.rst
@@ -8,13 +8,12 @@ Elasticsearch (E20)
 .. toctree::
     :maxdepth: 2
 
-Deployment Guide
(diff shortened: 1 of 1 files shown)


source/deployment/elasticsearch.rst
```diff
@@ -8,13 +8,12 @@ Elasticsearch (E20)
 .. toctree::
     :maxdepth: 2
 
-Deployment Guide
```
amyblais: Looks good, expect I think this heading doesn't necessarily need to be removed. 0/5

justinegeffen: Re-reading the document now and I think it should be re-added and moved to just below the first note.

Issue #3111, opened 2019-10-25
Request for Documentation: ElasticSearch FAQs


matterdoc:
Mattermost user `katie.wiersgalla` from https://community.mattermost.com has requested the following be documented:

```
##### What indexes do we create?

Right now we create three types of indexes: posts, users and channels.  Users and channels only have one index each, but for posts we create multiple, aggregating the posts by date.

##### We don't store attachments in ES, correct?

Yes, we don't store files in ES.

##### What form is the data we send into ES?

The communication with ES happens through its REST API, so we send JSON messages to index entities and query for them.

##### Do we have any estimates on "Queries/user/day" or something like that?

It is difficult to estimate, but it depends on what they use ES for. Right now you can use ES for the search feature and from 5.12 on it will be possible to use it for autocompletion of users/channels. Every time a message is published, a channel is created or a user changes, either because their properties change (e.g.: change of the first name) or because they join/leave a channel, that post/channel/user is sent to ES.

As for queries, if the search feature through ES is enabled, every search will generate a query, and if the autocompletion feature is enabled, every user or channel autocompletion when writting a message or user search will generate a query.
```

See the original post [here](https://community.mattermost.com/_redirect/pl/iaypyggwm3ga9fbnjmcmus8m7w).

_This issue was generated from [Mattermost](https://mattermost.com) using the [Doc Up](https://github.com/jwilander/mattermost-plugin-docup) plugin._

justinegeffen:
I've added three of the four questions to the ES content. 

For question four, I'd like to clarify whether the customer is asking for a search string to account for that or whether they're confirming that it's possible to access logs that indicate the number of queries per user per day.

PR #3121, opened 2019-10-31
Update elasticsearch.rst

justinegeffen:
https://github.com/mattermost/docs/issues/3111

### DIFF
--- a/source/deployment/elasticsearch.rst
+++ b/source/deployment/elasticsearch.rst
@@ -109,8 +109,7 @@ What form of data is sent to Elasticsearch?
 Mattermost communicates with Elasticsearch through its REST API using JSON messages for indexing and querying entities.
 
 How much data is sent to Elasticsearch and when?
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


hanzei:
@justinegeffen Is this ready for review?

justinegeffen:
I commented on the original issue but am reposting here. I've added three of the four questions to the ES content.

For question four, I'd like to clarify whether the customer is asking for a search string to account for that or whether they're confirming that it's possible to access logs that indicate the number of queries per user per day.

I'd like to include use cases and/or examples if they'll benefit customers and decrease the need to reach out to us. It's just a case of what would be most useful to a larger audience.

justinegeffen:
Thanks, @wiersgallak! Before this is merged, I wanted to follow up on my question around point 4 and confirm whether the issue is limited to the customer in question or it's a general question. I can add the content either way, but if there's more context around it (and examples if required) that'd be very helpful. :)

wiersgallak:
@justinegeffen I'd be very generic about the answer, but think its worth including - maybe we can adjust the question a bit: 

**How much data is sent to ElasticSearch?**

Every time a message is published, a channel is created or a user changes, either because their properties change (e.g.: change of the first name) or because they join/leave a channel, that post/channel/user is sent to ES.

As for queries, if the search feature through ES is enabled, every search will generate a query, and if the autocompletion feature is enabled, every user or channel autocompletion when writing a message or user search will generate a query.

justinegeffen:
Thanks, @wiersgallak. I've updated the doc with the new entry.

PR #3132, opened 2019-11-05
Change recommendation around jobserver and schedulers.

grundleborg:
The code changed some time in the past (I can't figure out exactly when)
such that job servers are fully-fledged members of the cluster. This is
necessary and unavoidable, therefore they can also become the cluster
leader. Because of this, they need to have the possibility of running
the schedulers, since without that, there could be no server running the
schedulers if the job server is elected cluster leader, meaning no jobs
get scheduled. Given the only use of cluster leader status in the
codebase now is job scheduling, this change should have no side effects
on any other functionality.

@wiersgallak since this is a significant change to our recommendations for enterprise cluster setup, adding you to the PR to ensure this gets noted in relevant places and communicated effectively to existing customers and our SA and CS teams.

**Note: there is no corresponding code change to this PR, it is simply a configuration recommendation change. It is compatible with all currently supported versions of Mattermost.**

### DIFF
--- a/source/deployment/cluster.rst
+++ b/source/deployment/cluster.rst
@@ -291,7 +291,7 @@ Mattermost runs periodic tasks via the `job server <https://docs.mattermost.com/
  - Compliance exports
  - Elasticsearch indexing
 
-:ref:`Make sure you have set ``JobSettings.RunScheduler`` to ``true`` in config.json for all app and job servers in the cluster. The cluster leader will then be responsible for scheduling recurring jobs.
+Make sure you have set ``JobSettings.RunScheduler`` to ``true`` in config.json for all app and job servers in the cluster. The cluster leader will then be responsible for scheduling recurring jobs.
 
 Plugins and High Availability
 ^^^^^^^^^^^^^^^^
--- a/source/deployment/cluster.rst
+++ b/source/deployment/cluster.rst
@@ -291,7 +291,7 @@ Mattermost runs periodic tasks via the `job server <https://docs.mattermost.com/
  - Compliance exports
  - Elasticsearch indexing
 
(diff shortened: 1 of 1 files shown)


source/deployment/cluster.rst
```diff
@@ -291,7 +291,7 @@ Mattermost runs periodic tasks via the `job server <https://docs.mattermost.com/
  - Compliance exports
  - Elasticsearch indexing
 
-:ref:`Run all job servers <command-line-tools-mattermost-jobserver>` with ``--noschedule flag``, then set ``JobSettings.RunScheduler`` to ``true`` in config.json for all app servers in the cluster. The cluster leader will then be responsible for scheduling recurring jobs.
+:ref:`Make sure you have set ``JobSettings.RunScheduler`` to ``true`` in config.json for all app and job servers in the cluster. The cluster leader will then be responsible for scheduling recurring jobs.
```
amyblais:
```suggestion
Make sure you have set ``JobSettings.RunScheduler`` to ``true`` in config.json for all app and job servers in the cluster. The cluster leader will then be responsible for scheduling recurring jobs.
```
Either remove the reference or add a link to the config.json doc.

mkraft:
> I don’t have a lot of context to offer here, but it seems reasonable. Is there a way to “do the right thing” without requiring human configuration that could be wrong?

Yup. I came to ask the same thing. It seems like this config adds no value to HA setups and only provides a way to screw up.

grundleborg:
So `RunSchedulers=true` *is* the default value in config.json, meaning it would be correct by default. Calling it out here is more for the benefit of those who previously changed it.

I'd also be up for deprecating/removing the config setting say in 5.20 once this recommendation has been in production for a while. I want to play it safe just in case any unforseen issues come up due to the change of recommendation and not require a code change to go back to the old behaviour.

grundleborg:
@amyblais change accepted.

PR #3146, opened 2019-11-08
Add image text below images in intro section

jasonblais:
I don't know why the second image disappears after my edits..any thoughts?


jasonblais:
Also, I still wonder if the image should have a border,

amyblais:
I asked UX and they said that without borders the layout appears more clean.

PR #3160, opened 2019-11-15
Is this a correct update?

johnthompson365:
I wasn't 100% clear on what was being described, so trying to write something more explicit.


wiersgallak:
@justinegeffen Can you help review this PR for clarity and read-ability?  I think the content is there, but worried the way it is worded is not as clear as it could be.

PR #3181, opened 2019-11-21
Clarify read and search replicas are in E10 and higher

jasonblais:


source/administration/config-settings.rst
```diff
@@ -3753,8 +3755,9 @@ Changes to this setting require a server restart before taking effect.
 | This feature's ``config.json`` setting is ``"DataSourceReplicas": []`` with string array input consisting of database connection strings.   |
 +---------------------------------------------------------------------------------------------------------------------------------------------+
 
-Search Replicas (Enterprise Edition)
```
jasonblais: @wiersgallak wondering if this *should* be in EE and we add the restriction in code, or if we just remove the EE tag from our docs.

0/5, not sure how many non-EE deployments use search replicas. That might help us determine impact of adding a restriction in the product

wiersgallak: I can chat with the Platform team and see if I can get more clarification and information on this next week.  Is there any harm leaving it consistent with the read replicas in docs for now until we have a clearer path forward?

jasonblais: We can, let me know if you prefer that and I can make the change

jasonblais: @wiersgallak I've updated the PR so it's consistent for both read and search replicas

PR #3196, opened 2019-11-28
MM Operator DB restore

justinegeffen:
Section on DB restore added. WIP.


source/install/install-kubernetes-mattermost.rst
```diff
@@ -107,3 +107,85 @@ Use your domain registration service to create a canonical name or IP address re
...
+    restoreSecret: myawscreds
+
+8. To initiate deployment, apply the file, specifying the correct path, using:
+
+.. code-block:: sh
+
+      $ kubectl create ns mattermost
+      $ kubectl apply -n mattermost -f /path/to/mattermost-installation.yaml
+
+The deployment process can be monitored in the Kubernetes user interface. Once complete, access your Mattermost
```
justinegeffen: Assuming the deployment can be monitored in Kubernetes, what are the remediation steps required if the restoration fails (i.e., restoration state = failed). Does the entire deployment fail? What error/status message is returned and what are the possible reasons for failure?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -61,7 +78,98 @@ Save the file as ``mattermost-installation.yaml``:
...
+Navigate to the ``ingressName`` URL in your browser and use Mattermost.
+
+Restoring an Existing Mattermost MySQL Database
+-----------------------------------------------
+
+You can leverage the Mattermost Operator to utilize an existing Mattermost MySQL database with a new Mattermost installation, in its own namespace.
+The steps you follow to create and upload your backup depends on the provider you're using and your use case. It's
+recommended that you consult the relevant documentation or, if your deployment is managed in a different way, consult your Administrator.
+
+It is important to note that this process requires the creation of a new Mattermost
```
justinegeffen: This needs to be much more explicit about this being a "new" installation as such. Is there a situation where this can be applied to an existing cluster installation file at all? Or would the customer always have to create a new namespace/install yaml? Not sure how much detail we want to go into about a use case where a customer could spin this up in a new namespace and then take down the other instance (if that would be relevant/accurate).

source/install/install-kubernetes-mattermost.rst
```diff
@@ -61,7 +78,98 @@ Save the file as ``mattermost-installation.yaml``:
...
+  $ kubectl -n mattermost get ingress
+
+Copy the resulting hostname or IP address from the ``ADDRESS`` column, open your browser, and connect to Mattermost.
+
+Use your domain registration service to create a canonical name or IP address record for the ``ingressName`` in your manifest,
+pointing to the address you just copied. For example, on AWS you would do this within a hosted zone in Route53.
+
+Navigate to the ``ingressName`` URL in your browser and use Mattermost.
+
+Restoring an Existing Mattermost MySQL Database
```
justinegeffen: I wonder whether this wouldn't be better off in the Install Cluster section of the guide because it covers installation/deployment and has a bit more complexity to it?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -61,7 +78,98 @@ Save the file as ``mattermost-installation.yaml``:
...
+Open a text editor and create a text file containing your credentials which will be used to access the uploaded database.
+
+Save the file as ``secret.yaml``. The example below is for AWS/S3.
+
+.. code-block:: yaml
+
+   apiVersion: v1
+   kind: Secret
+   metadata:
+    name: test-restore
```
justinegeffen: This is a "freestanding" file so the name doesn't seem to be used in any of the following examples. Correct?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -61,7 +78,98 @@ Save the file as ``mattermost-installation.yaml``:
...
+   type: Opaque
+   stringData:
+    AWS_ACCESS_KEY_ID: XXXXXXXXXXXX
+    AWS_SECRET_ACCESS_KEY: XXXXXXXXXXXX/XXXXXXXXXXXX
+    AWS_REGION: us-east-1
+    S3_PROVIDER: AWS
+
+**Parameters**
+- ``name``. The name of this manifest which is referenced in the installation manifest.
```
justinegeffen: In the restore.yaml file there is reference to "example-mattermostrestoredb". Is this the file to which the reference is being made? If so, this is the incorrect name?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -61,7 +78,98 @@ Save the file as ``mattermost-installation.yaml``:
...
+
+
+6. Create a Mattermost cluster installation manifest.
```
justinegeffen: Should we link to the install documentation which will also shortly include the custom settings and mention that this is a separate installation of Mattermost and should be sized appropriately?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -61,7 +78,98 @@ Save the file as ``mattermost-installation.yaml``:
...
+  kind: ClusterInstallation
+  metadata:
+    name: mm-example-full
+  spec:
+    size: 5000users
+    ingressName: example.mattermost-example.com
+    ingressAnnotations:
+      kubernetes.io/ingress.class: nginx
+    version: 5.14.0
+    mattermostLicenseSecret: ""
```
justinegeffen: Should this be the secret created in the initial section detailing installation/configuration? We don't mention this in the steps above so might be a good idea to call it out in the content above?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -61,7 +78,98 @@ Save the file as ``mattermost-installation.yaml``:
...
+
+.. code-block:: yaml
+
+  apiVersion: mattermost.com/v1alpha1
+  kind: ClusterInstallation
+  metadata:
+    name: mm-example-full
+  spec:
+    size: 5000users
+    ingressName: example.mattermost-example.com
```
justinegeffen: Would this be the same ingress name as their existing deployment or a new one?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -61,7 +78,98 @@ Save the file as ``mattermost-installation.yaml``:
...
+- ``name``. The name of this manifest which is referenced in the installation manifest.
+
+
+6. Create a Mattermost cluster installation manifest.
+
+Open a text editor and create a text file with the following details. Save the file as ``mattermost-installation.yaml``:
+
+.. code-block:: yaml
+
+  apiVersion: mattermost.com/v1alpha1
```
justinegeffen: How will customers know when the API version is updated and is there an easy way for them to change this?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -80,30 +188,40 @@ The Mattermost installation manifest contains fields which must be edited in lin
...
+    name: example-mattermostrestoredb
+  spec:
+    initBucketURL: s3://my-sample/my-backup.gz
+    mattermostClusterName: example-clusterinstallation
+    mattermostDBName: mattermostdb
+    mattermostDBPassword: supersecure
+    mattermostDBUser: mmuser
+    restoreSecret: myawscreds
+
+**Parameters**
```
justinegeffen: If we do choose to include this, should we remove the existing text in the example or keep it in?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -80,30 +188,40 @@ The Mattermost installation manifest contains fields which must be edited in lin
...
-
-Copy the resulting hostname or IP address from the ``ADDRESS`` column, open your browser, and connect to Mattermost.
+      $ kubectl create ns mattermost
+      $ kubectl apply -n mattermost -f /path/to/mattermost-installation.yaml
 
-Use your domain registration service to create a canonical name or IP address record for the ``ingressName`` in your manifest,
-pointing to the address you just copied. For example, on AWS you would do this within a hosted zone in Route53.
+The deployment process can be monitored in the Kubernetes user interface. If errors or issues are experienced,
+review the Mattermost log files or MySQL logs for guidance including error messages. If remediation is not successful, contact
+Mattermost customer support for assistance.
```
justinegeffen: Do we indicate that a support ticket should be opened?

source/install/use-kubernetes-mattermost.rst
```diff
@@ -82,13 +82,15 @@ Open the ``mattermost-installation.yaml`` file and add ``blueGreen`` under ``spe
...
 
-To access the new ingresses, create CNAME or IP address records in your DNS registration service for the ``ingressName`` in your manifest, pointing to the address you just copied. For example, on AWS you would do this within a hosted zone in Route53. Use the required ``ingressName`` URL in your browser to directly access blue or green at any time.
+To access the new ingresses, create CNAME or IP address records in your DNS registration service for the ``ingressName`` in your
+manifest, pointing to the address you just copied. For example, on AWS you would do this within a hosted zone in Route53.
+Use the required ``ingressName`` URL in your browser to directly access blue or green at any time.
 
 To update the version of blue or green, change the version in the manifest to
 match the current version or the version you’d like to deploy. This
 change (regardless of which is the ``productionDeployment``) initiates a database migration.
 The schema is backwards and forwards compatible across minor versions  (from 5.9 onwards) and will not disrupt the production deployment.
```
justinegeffen:
```suggestion
The schema is backwards and forwards compatible across minor versions (from 5.9 onwards) and will not disrupt the production deployment.
```

source/install/use-kubernetes-mattermost.rst
```diff
@@ -82,13 +82,15 @@ Open the ``mattermost-installation.yaml`` file and add ``blueGreen`` under ``spe
...
+manifest, pointing to the address you just copied. For example, on AWS you would do this within a hosted zone in Route53.
+Use the required ``ingressName`` URL in your browser to directly access blue or green at any time.
 
 To update the version of blue or green, change the version in the manifest to
 match the current version or the version you’d like to deploy. This
 change (regardless of which is the ``productionDeployment``) initiates a database migration.
-The schema is backwards and forwards compatible across minor versions  (from 5.9 onwards) and will not disrupt the production deployment.
-However, it will auto-upgrade the database. 
+The schema is backwards and forwards compatible across minor versions (from 5.9 onwards) and will not disrupt the production deployment.
+However, it will auto-upgrade the database.
```
justinegeffen: It seems like this could be a major change or potentially cause issues. Is there any data loss as a side-effect?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -80,30 +188,40 @@ The Mattermost installation manifest contains fields which must be edited in lin
...
+
+8. To initiate deployment, apply the file and specify the path where the newly-created files have been saved:
 
 .. code-block:: sh
 
-  $ kubectl -n mattermost get ingress
-
-Copy the resulting hostname or IP address from the ``ADDRESS`` column, open your browser, and connect to Mattermost.
+      $ kubectl create ns mattermost
+      $ kubectl apply -n mattermost -f /path/to/mattermost-installation.yaml
```
justinegeffen: Is this the right command? Would the yaml file not be "example-mattermostdbrestore" as stated above?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -80,30 +188,158 @@ The Mattermost installation manifest contains fields which must be edited in lin
...
-The deployment process can be monitored in the Kubernetes user interface.
+The deployment process can be monitored in the Kubernetes user interface. If errors or issues are experienced,
+review the Mattermost log files or MySQL logs for guidance including error messages. If remediation is not successful, contact
+Mattermost customer support for assistance.
 
-**4. Configure DNS and Use Mattermost**
+Once complete, access your Mattermost instance and confirm that the database has been restored.
 
-When the deployment is complete, obtain the hostname or IP address of your Mattermost deployment using the following command:
+Using an External Database with the Mattermost Operator
```
justinegeffen: This section is a copy of the restore DB section and review is needed to confirm content.

justinegeffen: @jwilander, I thought this section would simply be a replication of the restore DB section but it needs a review as I don't think my understanding is accurate.

source/install/install-kubernetes-custom.rst
```diff
@@ -0,0 +1,62 @@
+.. _install-kubernetes-custom:
```
justinegeffen: This section is in progress and was included in this PR for discussion. Currently adding a section on replication, as well as storage requirements using the content from https://github.com/mattermost/mattermost-operator/blob/master/pkg/apis/mattermost/v1alpha1/clusterinstallation_sizes.go. 

It's slow-going so if there's a brain dump somewhere that outlines the basics of replication and storage (in conjunction with the info provided) that'd be great.

justinegeffen:
Quick reminder for review on this when you have a moment, @jwilander.

source/install/install-kubernetes-custom.rst
```diff
@@ -0,0 +1,73 @@
...
+
+.. csv-table::
+   :header: "User Count", "Node Count", "Memory per Node", "vCPU per Node"
+   "5,000", "6", "8 GB", "4"
+   "10,000", "8", "16 GB", "4"
+   "25,000", "14", "16 GB", "4"
+
+The resource requirements, as well as the ingress name are added to the relevant field in the Mattermost manifest file, which is then deployed in your cluster.
+
+Customizing Resource Requirements
```
justinegeffen: This section likely needs some elaboration, especially around replication and the calculations performed. All the information is in the repo (https://github.com/mattermost/mattermost-operator/blob/master/pkg/apis/mattermost/v1alpha1/clusterinstallation_sizes.go) so we can either link to the repo, or provide the various calculations here, in which case some additional context may be needed.

justinegeffen:
@jwilander, @wiersgallak: I've made some updates and added more comments to the doc. Thanks!

source/install/install-kubernetes-mattermost.rst
jwilander: Yes, I'd agree with that

source/install/install-kubernetes-mattermost.rst
jwilander: The gist of it is that the restore is meant for the creation time of the installation, not for an existing installation. As long as that's clear I think we're good.

source/install/install-kubernetes-mattermost.rst
jwilander: The name is just how it shows up in kubernetes. It doesn't really affect anything other than acting as a label

source/install/install-kubernetes-mattermost.rst
jwilander: 👍

source/install/install-kubernetes-mattermost.rst
jwilander: This is a common/best practice pattern in k8s. When the api version changes, they can start using it by just applying a new manifest that use the new version

source/install/install-kubernetes-mattermost.rst
jwilander: Same name but the underlying resource will be different

source/install/install-kubernetes-mattermost.rst
jwilander: Yes, same secret

source/install/install-kubernetes-mattermost.rst
```diff
@@ -80,30 +188,158 @@ The Mattermost installation manifest contains fields which must be edited in lin
...
+    initBucketURL: s3://my-sample/my-backup.gz
+    mattermostClusterName: example-clusterinstallation
+    mattermostDBName: mattermostdb
+    mattermostDBPassword: supersecure
+    mattermostDBUser: mmuser
+    restoreSecret: myawscreds
+
+**Parameters**
+- ``mattermostClusterName``. The ClusterInstallation file name.
+- ``RestoreSecret``. The location of the backup file.
```
jwilander:
```suggestion
- ``restoreSecret``. The location of the backup file.
```

source/install/install-kubernetes-mattermost.rst
jwilander: 0/5, might be good to remove it

source/install/install-kubernetes-mattermost.rst
jwilander: The command is correct but yes the filename would be `restore.yaml` based on what the docs currently say

source/install/install-kubernetes-mattermost.rst
jwilander: This whole case doesn't need to exist. If the database is external than the operator doesn't need to manage it at all

source/install/use-kubernetes-mattermost.rst
jwilander: Nope, there shouldn't any problems. The upgrade paths are tested well.

source/install/install-kubernetes-mattermost.rst
justinegeffen: Got it, thanks!

source/install/install-kubernetes-mattermost.rst
justinegeffen: Changed it around to see whether it works better.

source/install/install-kubernetes-mattermost.rst
justinegeffen: Got it.

source/install/install-kubernetes-mattermost.rst
justinegeffen: Could they update the existing manifest with the new API version? Or must a new one be created?

source/install/install-kubernetes-mattermost.rst
justinegeffen: I've added that as "This is the same name as the existing deployment, but the underlying resource is the restored database".

source/install/install-kubernetes-mattermost.rst
justinegeffen: It is duplicate content and I totally understand why it doesn't need to be in the docs. My next question is how do we address the customer questions around using an existing db with MM operator (which is why I added this). It seemed like these would be the correct steps minus the references to restore files and replacing them with something like "newDB" but it didn't turn out as expected.

source/install/install-kubernetes-mattermost.rst
justinegeffen: Great, thanks!

source/install/install-kubernetes-mattermost.rst
justinegeffen: I've added more information in this regard but the custom installation section is incomplete pending review so have not referred to that.

source/install/install-kubernetes-custom.rst
justinegeffen: Discussed this with @wiersgallak and considering removing it - however it may be useful to customers restoring their db and needing more custom configs (and understanding replication).

source/install/install-kubernetes-mattermost.rst
justinegeffen: That makes sense but I don't think the docs are clear on that. 

So, to clarify, would this instance be the de facto one going forward and any current instances/deployments would be taken down once this is up and running? Or could they run in tandem - not sure why that would happen unless it's to ensure all is running correctly but in that case wouldn't blue-green work better?

source/install/install-kubernetes-mattermost.rst
justinegeffen: To follow on from that, should we include a link to the repo and say something like "Check the repo for details of API version updates"?

source/install/install-kubernetes-custom.rst
wiersgallak: If this is holding up the publish, recommend we start another PR (or open an issue to track) with this content to be worked on after this one is approved.

justinegeffen: Agreed. I'll move it out.

source/install/install-kubernetes-custom.rst
wiersgallak: To start, I'd recommend linking to the repo - and reduce the duplication of information.  We can always improve this document with customer feedback.

justinegeffen: I'm going to start a new PR for this so it doesn't hold up this PR.

source/install/install-kubernetes-mattermost.rst
```diff
@@ -35,9 +35,26 @@ Save the file as ``mattermost-license-secret.yaml``. Apply the file, specifying
...
+
+.. csv-table::
+    :header: "Field", "Description", "Must Edit"
 
-Save the file as ``mattermost-installation.yaml``:
+    "metadata.name", "The name of your Mattermost as it will be shown in Kubernetes. The shorter the better.", "Yes"
+    "spec.size", "The size of your installation. This can be '100users', '1000users, '5000users', '10000users', or '25000users'.", "Yes"
+    "spec.ingressName", "The DNS for your Mattermost installation.", "Yes"
+    "spec.version", "The Mattermost version.", "No"
+    "spec.mattermostLicenseSecret", "The name of the Kubernetes secret containing your license (e.g. mattermost-license-secret). Required for Enterprise deployments.", "Yes"
```
gabrieljackson: I think I would technically consider the license to be something you don't have to edit considering you could just be running without a license.

```suggestion
    "spec.mattermostLicenseSecret", "The name of the Kubernetes secret containing your license (e.g. mattermost-license-secret). Required for Enterprise deployments.", "No"
```

source/install/install-kubernetes-mattermost.rst
gabrieljackson: @jwilander, I believe backup and restore was also for possible rollback of the DB in case of a large issue.

For instance, should the docs also highlight that this could be used on an existing mattermost installation to rollback the DB without creating a new installation entirely?

source/install/install-kubernetes-mattermost.rst
```diff
@@ -107,3 +105,123 @@ Use your domain registration service to create a canonical name or IP address re
...
+It is important to note that this process requires the creation of a new Mattermost
+installation - editing the existing ``.yaml`` files is not recommended and can result in data loss.
+
+The process described below needs to be completed prior to proceeding with the Mattermost deployment.
+
+1. Create a backup of your database (e.g. using *mysqldump*).
+2. Deploy a new server (e.g. an AWS instance).
+3. Install a backup program and back up the database on the new server/instance.
+4. Upload the backed up database to your cloud storage provider (e.g. Amazon S3).
+5. Create a ``secret.yaml``` file:
```
gabrieljackson:
```suggestion
5. Create a ``secret.yaml`` file:
```

source/install/install-kubernetes-mattermost.rst
```diff
@@ -107,3 +105,123 @@ Use your domain registration service to create a canonical name or IP address re
...
+    "spec.version", "The Mattermost version.", "No"
+    "spec.mattermostLicenseSecret", "The name of the Kubernetes secret containing your license (e.g. mattermost-license). Required for Enterprise deployments.", "Yes"
+    "spec.database.storageSize", "The storage size for your database. Your Kubernetes cluster must have volumes this size or larger.", "No"
+    "spec.minio.storageSize", "The storage size for your file storage. Your Kubernetes cluster must have volumes this size or larger.", "No"
+    "spec.elasticSearch", "The section for Elasticsearch settings. Remove this section if you will not be using Elasticsearch.", "No"
+    "spec.elasticSearch.host", "The hostname for your Elasticsearch instance. Remove this section if you will not be using Elasticsearch.", "No"
+    "spec.elasticSearch.username", "The username for your Elasticsearch instance. Remove this section if you will not be using Elasticsearch.", "No"
+    "spec.elasticSearch.password", "The password for your Elasticsearch instance. Remove this section if you will not be using Elasticsearch.", "No"
```
gabrieljackson: This looks to be the same table as above. Do we want to duplicate it? 0/5 on this, but would need to change the license column to match above if that is changed.

source/install/install-kubernetes-mattermost.rst
```diff
@@ -107,3 +105,123 @@ Use your domain registration service to create a canonical name or IP address re
...
+- ``mattermostDBPassword``. The password used to access the database.
+- ``mattermostDBUser``. The username required to access the database.
+- ``initBucketURL``. The URL of the storage instance/server where the backed up DB is stored.
+
+8. To initiate deployment, apply the file and specify the path where the newly-created files have been saved:
+
+.. code-block:: sh
+
+      $ kubectl create ns mattermost
+      $ kubectl apply -n mattermost -f /path/to/restore.yaml
```
gabrieljackson: This only covers the commands for the restore manifest. I think we need the same thing for the Mattermost cluster installation and the restore secret.

```
kubectl create ns mattermost
kubectl apply -n mattermost -f /path/to/secret.yaml
kubectl apply -n mattermost -f /path/to/mattermost-installation.yaml
kubectl apply -n mattermost -f /path/to/restore.yaml
```

source/install/install-kubernetes-mattermost.rst
```diff
@@ -107,3 +105,123 @@ Use your domain registration service to create a canonical name or IP address re
...
+
+The deployment process can be monitored in the Kubernetes user interface. If errors or issues are experienced,
+review the Mattermost log files or MySQL logs for guidance including error messages. If remediation is not successful, contact
```
gabrieljackson:
```suggestion
review the Mattermost, Operator, and MySQL logs for guidance including error messages. If remediation is not successful, contact
```

justinegeffen:
I agree with you, @gabrieljackson. I think there is missing context around db restore. Do you have any use cases in mind that would fit this section?

source/install/install-kubernetes-mattermost.rst
justinegeffen: Do we want to change the license secret name or remove reference? Happy to remove the table too, also 0/5 on that.

source/install/install-kubernetes-mattermost.rst
justinegeffen: If that's the use case, which makes sense, would it be better if we moved this to the section with Canary Builds and Blue-green as another management option vs. a discrete process? Then it may have more context.

source/install/install-kubernetes-mattermost.rst
justinegeffen: Thanks, @gabrieljackson! Apologies for the delayed reply. I've added this to the restore section.

justinegeffen:
Thanks for the feedback, @gabrieljackson! I have a few comments. 

As as aside, there are some outstanding bits of content, particularly around sizing. As discussed it would be great if the help wanted ticket that was mentioned earlier this week could include a documentation update so we could add it here. :)

justinegeffen:
Quick reminder on the comments, @gabrieljackson when you have a moment. :)

justinegeffen:
@jwilander, @gabrieljackson if you have a gap to review the comments that would be awesome - I'd like to have this over the line this week. I think it's probably good to go, just a few comments that need to be confirmed/addressed. Thank you! :)

gabrieljackson:
Reviewing now. Thanks for the patience.

source/install/install-kubernetes-mattermost.rst
gabrieljackson: I am also 0/5 on it, but my slight preference would be to remove the table just to not have two things that could possibly be out of sync.

source/install/install-kubernetes-mattermost.rst
gabrieljackson: I believe the location of this information is fine, but I would suggest that we change the wording a bit. Generally, I see the backup and restore process having two very common use-cases:
1. What is outlined here. You had Mattermost running and want to start using it in k8s with our operator. To do so, you create a new installation and restore the data like is described in the PR already.
2. You have backups going for your operator-run mattermost server. One day something happens and you want to revert the database to X days ago, so you use this backup and restore process to do just that on the existing installation.

That seems correct, right @jwilander? I could have sworn we wanted both options.

@justinegeffen, I am also completely fine with that coming in a follow-up PR. The info presented here isn't wrong by any account.

justinegeffen: That's very helpful feedback, thank you! I've edited the introduction to include both aspects.

PR #3214, opened 2019-12-05
Update open-source-components.rst

amyblais:


source/administration/open-source-components.rst
```diff
@@ -45,6 +46,7 @@ Mobile
 Redux
 ------------------------------
 
+ - Mattermost Enterprise Edition v5.18.0 - `View Open Source Components <https://github.com/mattermost/mattermost-redux/blob/release-5.18/NOTICE.txt>`_.
```
jasonblais: @amyblais I forget, why do we reference `Enterprise Edition` in this case?

amyblais: @jasonblais I'm not sure but I can remove it.

jasonblais: @amyblais 0/5. We could also say `Mattermost Redux v5.18.0`, etc. But I'm not sure if there's a contractual reason for having Enterprise Edition, since the binary we ship includes webapp, server and redux. Melissa Warner might know

PR #3222, opened 2019-12-06
Update note about Mattermost supporting mobile wrapping

jasonblais:
I had originally written the below, but then condensed it to just say 'we don't support it' for clarity. Didn't want to confuse a custom build is required even with AppConfig.

```
The use of hosted applications by Mattermost :doc:`can be deployed with Enterprise Mobility Management solutions via AppConfig <mobile-appconfig>` but do not support wrapping since each provider would require their own SDK and a custom build. Moreover, majority of the providers do not support WebSocket connections without using a mobile app provider like BlueCedar or AppDome.
```


amyblais:
I'm not sure how to verify if the doc link is working.

jasonblais:
> I'm not sure how to verify if the doc link is working.

@amyblais Good question - @justinegeffen wondering if you know how to confirm doc links with RST format work prior to merging?

justinegeffen:
That's a good question! I don't know how to validate it in GitHub but I ran a local build and it it's displaying correctly and the link works. 
Screen Shot 2019-12-06 at 20 01 02.

PR #3227, opened 2019-12-09
Add step to set SiteURL.

src-r-r:
Added a step to include setting `SiteURL` in `config.json`

> 8. Also set ``"SiteURL"`` to the full base URL of the site (e.g. ``"https://mattermost.example.com"``).

The mattermost instance won't start if the SiteURL isn't set. I figured it would be easier to read ahead of time instead of reading the logs to figure out what went wrong.


cpanato:
@src-r-r Thanks for this PR and the contribution. Can you please sign the CLA? thanks again

PR #3240, opened 2019-12-12
Update data-retention.rst

justinegeffen:
Fixed procedure order and UI terms.

### DIFF
--- a/source/administration/data-retention.rst
+++ b/source/administration/data-retention.rst
@@ -18,11 +18,11 @@ Configuring a Data Retention Policy
 To set a custom data retention policy:
 
 1. Go to **System Console > Compliance > Data Retention Policy** (or **System Console > Advanced > Data Retention Policy** in versions prior to 5.12).
-2. Select a **Message Retention** option. Messages, including file attachments, older than the duration you set will be deleted nightly. The minimum message retention time is one day.
-3. Select a **File Retention** option. File uploads older than the duration you set will be deleted nightly from your file storage system, either from your local disk or your Amazon S3 service as specified in **System Console > Environment > File Storage** (or **System Console > Files > Storage** in versions prior to 5.12). The minimum file retention time is one day.
+2. Select a **Message Retention** option. When a time is specified, messages, including file attachments, older than the duration you set will be deleted at the specified time. The minimum retention period is one day.
+3. Select a **File Retention** option. When a time is specified uploaded files which are older than the duration you set will be deleted from your file storage system (either from your local disk or your Amazon S3 service as specified in **System Console > Environment > File Storage** (or **System Console > Files > Storage** in versions prior to 5.12)) at the specified time. The minimum retention period is one day.
 4. Set the start time of the daily scheduled data retention job. Choose a time when fewer people are using your system. Must be a 24-hour time stamp in the form HH:MM.
 
-Save the settings and restart your server. Messages and files older than the duration you set will be deleted daily at the specified server time.
+Save the settings and restart your server. Messages and files older than the duration you set will be deleted at the specified server time, if applicable.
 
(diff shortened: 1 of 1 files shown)


source/administration/data-retention.rst
```diff
@@ -5,28 +5,28 @@ Available in `Enterprise Edition E20 <https://about.mattermost.com/pricing/>`__.
...
-3. Similarly, set your desired retention policy for file uploads. File uploads older than the duration you set will be deleted nightly from your file storage system, either from your local disk or your Amazon S3 service as specified in **System Console > Files > Storage** in prior versions or **System Console > Environment > File Storage** in versions after 5.12. The minimum file retention time is one day.
+1. Go to **System Console > Compliance > Data Retention Policy** (or **System Console > Advanced > Data Retention Policy** in versions prior to 5.12).
+2. Select a **Message Retention** option. Messages, including file attachments, older than the duration you set will be deleted nightly. The minimum message retention time is one day.
+3. Select a **File Retention** option. File uploads older than the duration you set will be deleted nightly from your file storage system, either from your local disk or your Amazon S3 service as specified in **System Console > Environment > File Storage** (or **System Console > Files > Storage** in versions prior to 5.12). The minimum file retention time is one day.
```
justinegeffen: Question: If the deletion time is set to 10am, would the messages/files still be deleted "nightly"? How would this work for the "indefinitely" setting? 

Suggested wording: "When a set time is specified, messages, including file attachments, older than the duration you set will be deleted at the time specified. The minimum retention period is one day."

justinegeffen: Same comments as for the previous step:
Question: If the deletion time is set to 10am, would the messages/files still be deleted "nightly"? How would this work for the "indefinitely" setting? 

Suggested wording: "When a set time is specified, messages, including file attachments, older than the duration you set will be deleted at the time specified. The minimum retention period is one day."

source/administration/data-retention.rst
```diff
@@ -5,28 +5,28 @@ Available in `Enterprise Edition E20 <https://about.mattermost.com/pricing/>`__.
...
 4. Set the start time of the daily scheduled data retention job. Choose a time when fewer people are using your system. Must be a 24-hour time stamp in the form HH:MM.
```
justinegeffen: The UI calls this "Data Deletion Time" and the text says: "Set the start time of the daily scheduled data retention job. Choose a time when fewer people are using your system. Must be a 24-hour time stamp in the form HH:MM.**"

I'm not sure which aspect should be changed, but leaning toward changing the field name to "Retention Job Start Time" or similar?

source/administration/data-retention.rst
```diff
@@ -5,28 +5,28 @@ Available in `Enterprise Edition E20 <https://about.mattermost.com/pricing/>`__.
...
 
-Save the settings and restart your server. You're now all set! Data older than the duration you set are now deleted daily at the specified server time.
+Save the settings and restart your server. Messages and files older than the duration you set will be deleted daily at the specified server time.
```
justinegeffen: Again, deleted daily vs. a specified time, coupled with the "data deletion time" is confusing.

source/administration/data-retention.rst
wiersgallak: @justinegeffen Agree with your recommended suggestion.

wiersgallak: Agreed on suggestion.

source/administration/data-retention.rst
wiersgallak: I agree that this name is confusing. Could you create a ticket for proposed UI changes to the name?

source/administration/data-retention.rst
wiersgallak: Makes sense to me - nightly is not accurate in the 10 am or the indefinitely scenarios.

source/administration/data-retention.rst
justinegeffen: Yep, I'll do that. Thanks!

PR #3253, opened 2019-12-17
Improve breaking change notice in 5.18 changelog regarding module path 

hanzei:
#### Ticket Link
https://community.mattermost.com/core/pl/76uzwugmt38758a54ozf7tp7wa


source/administration/changelog.md
```diff
@@ -13,8 +13,10 @@ Mattermost v5.18.0 contains low to high level security fixes. [Upgrading](http:/
 ### Compatibility
  
 ### Important Upgrade Notes
- - Marking a post unread from the mobile app requires v1.26 or later. If using v5.18, but mobile is on v1.25 or earlier, marking a post unread from webapp/desktop will only be reflected on mobile the next time the app launches or is brought to the foreground.
- - The ``mattermost-server`` import path of all the Go files in the repository as well as the module lines were changed to comply with the module version specification.
+- Marking a post unread from the mobile app requires v1.26 or later. If using v5.18, but mobile is on v1.25 or earlier, marking a post unread from webapp/desktop will only be reflected on mobile the next time the app launches or is brought to the foreground.
+
+### Breaking Changes
+ - The go module path of ``mattermost-server`` was changed to comply with the module version specification. Go projects using ``mattermost-server`` as a library have to change the import path to ``github.com/mattermost/mattermost-server/v5``. You may use ``find . -name "*.go" -type f | xargs sed -i -e 's/github.com\/mattermost\/mattermost-server\//github.com\/mattermost\/mattermost-server\/v5\//g'`` updates the import path.
```
amyblais:
```suggestion
 - The go module path of ``mattermost-server`` was changed to comply with the module version specification. Go projects using ``mattermost-server`` as a library have to change the import path to ``github.com/mattermost/mattermost-server/v5``. You may use ``find . -name "*.go" -type f | xargs sed -i -e 's/github.com\/mattermost\/mattermost-server\//github.com\/mattermost\/mattermost-server\/v5\//g'`` to update the import path.
```

amyblais:
Same for https://docs.mattermost.com/administration/important-upgrade-notes.html.

source/administration/changelog.md
```diff
@@ -13,8 +13,10 @@ Mattermost v5.18.0 contains low to high level security fixes. [Upgrading](http:/
 ### Compatibility
  
 ### Important Upgrade Notes
- - Marking a post unread from the mobile app requires v1.26 or later. If using v5.18, but mobile is on v1.25 or earlier, marking a post unread from webapp/desktop will only be reflected on mobile the next time the app launches or is brought to the foreground.
- - The ``mattermost-server`` import path of all the Go files in the repository as well as the module lines were changed to comply with the module version specification.
+- Marking a post unread from the mobile app requires v1.26 or later. If using v5.18, but mobile is on v1.25 or earlier, marking a post unread from webapp/desktop will only be reflected on mobile the next time the app launches or is brought to the foreground.
+
+### Breaking Changes
+ - The go module path of ``mattermost-server`` was changed to comply with the module version specification. Go projects using ``mattermost-server`` as a library have to change the import path to ``github.com/mattermost/mattermost-server/v5``. You may use ``find . -name "*.go" -type f | xargs sed -i -e 's/github.com\/mattermost\/mattermost-server\//github.com\/mattermost\/mattermost-server\/v5\//g'`` to update the import path.
```
lieut-data: Should we relax the language slightly here to emphasize that this is not a breaking change for any existing integrations with Mattermost, but that this applies only to developers and only when updating their dependencies? I’m worried the current language will send the wrong message to system administrators who may not be familiar with Go.

On the `find` helper — while I appreciate that this may be helpful, I’ve been burned in the past between different platforms and their behaviour with respect to `find`, `xargs` and `sed`. Might we consider just showing a sample diff to illustrate the change, and leave the effort of doing so to developers? I note they will also have to update their `go.mod`, vendored folders (if necessary), etc., but that probably all shouldn’t have to be documented here. Is there a “standard Go module” document we could link to instead?

hanzei: Good idea. I've tweaked the language to make this clear :+1: 

I really want to give developers a tool to do the migration as painless as possible. And doing this by hand is painful. Could you name an example  bash issues you run into (I'm ignoring the fact that windows exists)?

Interestingly the Go blog did came up the almost the same command:
```
find . -type f \
    -name '*.go' \
    -exec sed -i -e 's,github.com/my/project,github.com/my/project/v2,g' {} \;
```
https://blog.golang.org/v2-go-modules

But I didn't found a standard page that we can link to.

source/administration/changelog.md
```diff
@@ -13,8 +13,10 @@ Mattermost v5.18.0 contains low to high level security fixes. [Upgrading](http:/
...
- - The ``mattermost-server`` import path of all the Go files in the repository as well as the module lines were changed to comply with the module version specification.
+- Marking a post unread from the mobile app requires v1.26 or later. If using v5.18, but mobile is on v1.25 or earlier, marking a post unread from webapp/desktop will only be reflected on mobile the next time the app launches or is brought to the foreground.
+
+### Breaking Changes
+ - The go module path of ``mattermost-server`` was changed to comply with the module version specification. Developers using using ``mattermost-server`` as a library have to change the import path to ``github.com/mattermost/mattermost-server/v5`` when they want to upgrade this dependency to `v5.18`. They may use ``find . -name "*.go" -type f | xargs sed -i -e 's/github.com\/mattermost\/mattermost-server\//github.com\/mattermost\/mattermost-server\/v5\//g'`` to update the import path.
```
jasonblais: How does this affect plugins @hanzei

source/administration/changelog.md
lieut-data: I agree that we should make this painless, but I'm not sure the release notes are the right place to do this. Could we link to that Go blog article instead? Given the need to update their `go.mod`, vendored folders, etc., a simple command might work, but could still leave them hanging.

I don't have an example on hand re: breakage, other than this burns me all the time when we write `Makefile` additions, but folks on different platforms run into things.

hanzei: I've updated the text to link to the blog post. PTAL.

source/administration/changelog.md
hanzei: In the same way as any other program using MM as a library. If they want to update there dependency to 5.18, they have to change the import path.

Does this answer your question?

hanzei:
Will update `source/administration/important-upgrade-notes.rst` once the PR is approved.

source/administration/changelog.md
```diff
@@ -16,7 +16,7 @@ Mattermost v5.18.0 contains low to high level security fixes. [Upgrading](http:/
 - Marking a post unread from the mobile app requires v1.26 or later. If using v5.18, but mobile is on v1.25 or earlier, marking a post unread from webapp/desktop will only be reflected on mobile the next time the app launches or is brought to the foreground.
 
 ### Breaking Changes
- - The go module path of ``mattermost-server`` was changed to comply with the module version specification. Developers using using ``mattermost-server`` as a library have to change the import path to ``github.com/mattermost/mattermost-server/v5`` when they want to upgrade this dependency to `v5.18`. They may use ``find . -name "*.go" -type f | xargs sed -i -e 's/github.com\/mattermost\/mattermost-server\//github.com\/mattermost\/mattermost-server\/v5\//g'`` to update the import path.
+ - The Go module path of ``mattermost-server`` was changed to comply with the Go module version specification. Developers using Go modules to manage there dependencies and have ``mattermost-server`` as a dependency, have to change the module path to ``github.com/mattermost/mattermost-server/v5`` when they want to upgrade this dependency to `v5.18`. See https://blog.golang.org/v2-go-modules for further information.
```
lieut-data: `s/there/their/`, and a condensed approach:

```suggestion
 - The Go module path of ``mattermost-server`` was changed to comply with the Go module version specification. Developers using Go modules with ``mattermost-server`` as a dependency must change the module and import paths to ``github.com/mattermost/mattermost-server/v5`` when upgrade this dependency to `v5.18`. See https://blog.golang.org/v2-go-modules for further information.
```

PR #3256, opened 2019-12-17
Update requirements.rst

amyblais:
I'm not sure if it's helpful to get analytics for this since Mozilla simply doesn't support other versions itself.

https://en.wikipedia.org/wiki/Firefox_version_history
https://www.whatismybrowser.com/guides/the-latest-version/firefox

### DIFF
--- a/source/install/requirements.rst
+++ b/source/install/requirements.rst
@@ -44,7 +44,7 @@ PC Web
     :header: "Browser", "Technical Requirement"
 
     "Chrome", "v77+"
-    "Firefox", "v60+"
+    "Firefox", "v68.3", "v71.0+"
     "Safari", "v12+"
     "Edge", "v44+"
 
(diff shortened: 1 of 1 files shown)


source/install/requirements.rst
```diff
@@ -44,7 +44,7 @@ PC Web
     :header: "Browser", "Technical Requirement"
 
     "Chrome", "v77+"
-    "Firefox", "v60+"
+    "Firefox", "v68.3", "v71.0+"
```
JtheBAB: May be change to

„Latest ESR or newer“

? Then you don’t need to change this every 6 weeks

jasonblais: Agree on the policy of the latest ESR and later (I think this was the intent, but I wrote incorrectly in the process doc..my mistake).

With that, we could say "v68+", which would stay as the supported version until September/2020 when the next ESR is released.

image

@amyblais I would also recommend checking the data each time. It can be a short one each time, but it's helpful to make sure we don't have a lot of large customers using Firefox 60. If yes, we may need to reach out to them to let them know about the version requirement change.

PR #3278, opened 2019-12-21
Fix the name of settings

kaakaa:
## Summary
Fix the name of settings for viewing archived channels.

In the document, the settings for viewing archived channels is named as `ViewArchivedChannels`
https://docs.mattermost.com/administration/config-settings.html#allow-users-to-view-archived-channels-beta
> This feature’s config.json setting is "ViewArchivedChannels": false with options true and false.

However, `config.json` in my environment doesn't have `ViewArchivedChannels`. Should its name be `ExperimentalViewArchivedChannels`?

FYI.
mattermost-server doesn't have `ViewArchivedChannels`, instead has `ExperimentalViewArchivedChannels`.
https://github.com/mattermost/mattermost-server/blob/be3e008dcab517ce1e9bb3a6c95de00ba7353b8a/model/config.go#L1525

#### Ticket Link
N/A

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -1453,7 +1453,7 @@ Allow Users to View Archived Channels (Beta)
 **False**: Users are unable to view, share or search for content of channels that have been archived.
 
 +-------------------------------------------------------------------------------------------------------------------------------------------------------+
-| This feature's ``config.json`` setting is ``"ViewArchivedChannels": false`` with options ``true`` and ``false``.                                      |
+| This feature's ``config.json`` setting is ``"ExperimentalViewArchivedChannels": false`` with options ``true`` and ``false``.                                      |
 +-------------------------------------------------------------------------------------------------------------------------------------------------------+
 
 Show Email Address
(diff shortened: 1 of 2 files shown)


justinegeffen:
Thanks for picking this up! You are absolutely correct about the discrepancy. I've requested feedback on this to cross-check that any other instances of this (where applicable) are also addressed.

source/administration/config-settings.rst
```diff
@@ -1453,7 +1453,7 @@ Allow Users to View Archived Channels (Beta)
 **False**: Users are unable to view, share or search for content of channels that have been archived.
 
 +-------------------------------------------------------------------------------------------------------------------------------------------------------+
-| This feature's ``config.json`` setting is ``"ViewArchivedChannels": false`` with options ``true`` and ``false``.                                      |
+| This feature's ``config.json`` setting is ``"ExperimentalViewArchivedChannels": false`` with options ``true`` and ``false``.                                      |
```
amyblais: The table is not rendering anymore. You can fix the table formatting by changing the "Line Wrap Mode" to "No Wrap" and then editing the table so that it is aligned.

kaakaa: Ah...sorry for my fix without any validations. I've fixed the table formatting by [this commit](https://github.com/mattermost/docs/pull/3278/commits/0a909a4d9f244ae44ab2df46d3f7afa0afd0d3bf) and confirmed that the table is rendering by building html.  Thanks!

kaakaa:
I found other instance in docs, and I've fixed this by [this](https://github.com/mattermost/docs/pull/3278/commits/e77e6bf3000e72db963d2d869b5ca6ca45bf52cb).

PR #3291, opened 2020-01-08
Update architecture.rst

comharris:
fixed typo

-_-

Also wondering about `real-time` vs `real time` (both appear here)


justinegeffen:
Good point about real-time vs. real time. I've done some Googling and either one is acceptable so it's better to have consistency and in that respect our docs seem to lean more toward real-time. 

I suggest using real-time. @lieut-data might have more insight into this though?

lieut-data:
@justinegeffen, I'd probably lean towards `real-time` if you want to make that change, if only because that term seems to be used most consistently in various Mattermost documentation.

comharris:
Done. Thanks @lieut-data and @justinegeffen

PR #3294, opened 2020-01-08
Update install-mmte-helm-gitlab-helm.rst

justinegeffen:
Added two new comments to the deployment step as per GitLab team suggestion.


jasonblais:
cc Katie in case you wanted to take a peak as well

PR #2645, opened 2019-04-19
Def of Mattermost staff & Mattermost employee

it33:
Drafting explicit definitions of "Mattermost staff", vs. "Mattermost employee" as they're been sprinkled throughout our docs but not yet made explicit. 

There's discussions of "Mattermoster" in different threads and it's not clear if things have gone dead or if people are thinking about follow-ups, in my mind, at the minimum the definitions of Mattermost staff and Mattermost employee are consistent with what we've been using in the past. 

Please flag if anything seems different here than before?  

Other threads

 
- https://gitlab.com/mattermost/proposals/issues/28
- https://github.com/mattermost/docs/pull/2617

PR #3296, opened 2020-01-09
Update training.rst

justinegeffen:
Revised some punctuation and fixed "Github" to "GitHub".

### DIFF
--- a/source/process/training.rst
+++ b/source/process/training.rst
@@ -26,14 +26,14 @@ Important things to know
 Hiring
 ---------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


justinegeffen:
I reviewed this document when reviewing this PR: https://github.com/mattermost/docs/pull/2645. I think the definition of Mattermost staff/Mattermost employee should be refined and added - along with definitions of "core committer" and "contributor". 

There should be a description to define contributors and FTEs but "staff" and "employee" may not be the best terms especially if we want to dive deeper into the various contribution "roles" we have. 

Perhaps this is best addressed by HR in a separate issue?

Issue #3303, opened 2020-01-13
SMTP/Email configuration still says to use /bin/bash in docker


svelle:
Since the switch to alpine linux we don't have bash installed in the docker images. 

[This page](https://docs.mattermost.com/install/smtp-email-setup.html) however still refers to using it.
We should change it to use /bin/sh instead **or** install bash in the docker images.

svelle:
Also it says to use telnet but we don't install telnet on the app image.

PackElend:
is `openssl `part of alpine linux?
As `SMTP `shall be encrypted, `openssl `is the right tool to use.

PR #3305, opened 2020-01-13
Update smtp-email-setup.rst

svelle:
Re: https://github.com/mattermost/docs/issues/3303

### DIFF
--- a/source/install/smtp-email-setup.rst
+++ b/source/install/smtp-email-setup.rst
@@ -163,25 +163,8 @@ Checking your SMTP server is reachable
 -  You must run the following commands from the same machine or virtual
    instance where ``mattermost/bin/mattermost`` is located. So if you're
    running Mattermost from Docker you need to
-   ``docker exec -ti mattermost-dev /bin/bash``
--  Telnet to the email server with ``telnet mail.example.com 25``. If
-   the command works you should see something like
-
-   ::
-
-       Trying 24.121.12.143...
-       Connected to mail.example.com.
-       220 mail.example.com NO UCE ESMTP
-
--  Then type something like ``HELO <your mail server domain>``. If the
-   command works you should see something like
-
-   ::
-
-       250-mail.example.com NO UCE
-       250-STARTTLS
-       250-PIPELINING
-       250 8BITMIME
+   ``docker exec -ti mattermost-dev /bin/sh``
+-  Ping the smtp server to see if a connection can be established.
 
 .. note::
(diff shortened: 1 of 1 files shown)


source/install/smtp-email-setup.rst
```diff
@@ -163,25 +163,8 @@ Checking your SMTP server is reachable
 -  You must run the following commands from the same machine or virtual
```
justinegeffen: It looks like the "following" commands have been removed from this section. Are there still commands that can/must be run? If not, perhaps something like this could work: 

- Ensure you are running commands from the same machine or virtual instance where mattermost/bin/mattermost is located. 
- Attempt to telnet to the email service to make sure the server is reachable.
- Ping the smtp server to see if a connection can be established.

justinegeffen:
Hi @svelle. Following up on this to confirm whether you've had a chance to look at the feedback. :)

svelle:
Hi @justinegeffen still struggling with github notifications so sorry if I missed it, I'll check it out now and respond to you within the next few minutes ;)

justinegeffen:
No problem at all - I've also had some issues hence the follow up. :)

source/install/smtp-email-setup.rst
svelle: So the problem here, and why I made the change, is that we don't install telnet inside the docker containers so we can't advise using that to verify the connection. 

I think referring to ping is probably the best thing we can do at this point apart from telling people to install telnet themselves and test it that way.

svelle:
@justinegeffen I completely redid the section and removed the direct reference to docker commands. 
Instead I added a note about telnet not being installed in the docker images and added some advice on how to work from there.

Let me know what you think :)

source/install/smtp-email-setup.rst
```diff
@@ -182,6 +180,10 @@ Checking your SMTP server is reachable
        250-STARTTLS
        250-PIPELINING
        250 8BITMIME
+       
+
+.. note:: 
+  We're not installing telnet by default on the official docker images, so you either need to use ``ping`` on those or install telnet yourself either directly in the machine or by modifying the dockerfile.
```
justinegeffen:
```suggestion
  As we're not installing telnet by default on the official docker images you either need to use ``ping`` on those or install telnet yourself either directly modifying the dockerfile.
```

Not sure what machine we're referring to so I removed it - happy to put it back if we can clarify.

PR #3306, opened 2020-01-13
Update config-settings.rst

justinegeffen:
Added value of 1500 to AD/LDAP Maximum Page Size config setting.

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -2225,7 +2225,10 @@ Set how often Mattermost accounts synchronize attributes with AD/LDAP, in minute
 
 Maximum Page Size
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-The maximum number of users the Mattermost server will request from the AD/LDAP server at one time. Use this setting if your AD/LDAP server limits the number of users that can be requested at once. 0 is unlimited.
+The maximum number of users the Mattermost server will request from the AD/LDAP server at one time. Use this setting if your AD/LDAP server limits the number of users that can be requested at once. 
+
+- A value of 0 is unlimited and does not paginate the results. 
+- A value of 1500 is recommended to align with the default AD/LDAP ``MaxPageSize`` setting. 
 
 +----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | This feature's ``config.json`` setting is ``"MaxPageSize": 0`` with numerical input.                                                                                 |


justinegeffen:
@jasonblais raised a good point about this in that the default in config.json is 0 but we recommend 1500. Will there be a code change to support that value?

mkraft:
> @jasonblais raised a good point about this in that the default in config.json is 0 but we recommend 1500. Will there be a code change to support that value?

Good call-out. I have created a story for it https://mattermost.atlassian.net/browse/MM-21765

justinegeffen:
> > @jasonblais raised a good point about this in that the default in config.json is 0 but we recommend 1500. Will there be a code change to support that value?
> 
> Good call-out. I have created a story for it https://mattermost.atlassian.net/browse/MM-21765

Should we mark this as WIP for now?

mkraft:
> > > @jasonblais raised a good point about this in that the default in config.json is 0 but we recommend 1500. Will there be a code change to support that value?
> > 
> > 
> > Good call-out. I have created a story for it https://mattermost.atlassian.net/browse/MM-21765
> 
> Should we mark this as WIP for now?

The way I read what you have written in this PR there's no mention of the default so MM-21765 wouldn't affect this text.

justinegeffen:
The documentation mentions this:
"----------------------------------------------------+
| This feature's ``config.json`` setting is ``"MaxPageSize": 0`` with numerical input.                                                                                 |
"
so I was wondering whether that would be problematic?

mkraft:
> The documentation mentions this:
> "----------------------------------------------------+
> | This feature's `config.json` setting is `"MaxPageSize": 0` with numerical input. |
> "
> so I was wondering whether that would be problematic?

That will need to be updated with MM-21765.

justinegeffen:
> > The documentation mentions this:
> > "----------------------------------------------------+
> > | This feature's `config.json` setting is `"MaxPageSize": 0` with numerical input. |
> > "
> > so I was wondering whether that would be problematic?
> 
> That will need to be updated with MM-21765.

So can we leave it in there for now, or should we hold off on pushing this change until MM-21765 is done, and then update the content to say "1500" instead?

mkraft:
> > > The documentation mentions this:
> > > "----------------------------------------------------+
> > > | This feature's `config.json` setting is `"MaxPageSize": 0` with numerical input. |
> > > "
> > > so I was wondering whether that would be problematic?
> > 
> > 
> > That will need to be updated with MM-21765.
> 
> So can we leave it in there for now, or should we hold off on pushing this change until MM-21765 is done, and then update the content to say "1500" instead?

We could push these changes now and do a separate documentation change once the default value is changed in code.

justinegeffen:
That works for me, thanks. I've started watching the Jira ticket and added a Documentation label so it's on my radar. Thanks for all the help. :)

PR #3311, opened 2020-01-16
5.18.2 docs

amyblais:


source/administration/changelog.md
```diff
@@ -8,6 +8,8 @@ Also see [changelog in progress](http://bit.ly/2nK3cVf) for the next release.
 
 Mattermost v5.18.0 contains low to high level security fixes. [Upgrading](http://docs.mattermost.com/administration/upgrade.html) is recommended. Details will be posted on our [security updates page](https://about.mattermost.com/security-updates/) 30 days after release as per the [Mattermost Responsible Disclosure Policy](https://www.mattermost.org/responsible-disclosure-policy/).
 
+- **v5.18.2, released 2020-01-16**
+  - Fixed an issue with a panic in ``web.shouldSendEventToGuest``. [MM-21481](https://mattermost.atlassian.net/browse/MM-21481)
```
jasonblais: What's the user (or admin) facing issue we're fixing?

jasonblais: ^ie. what info will an admin need to decide if they need to upgrade to 5.18.2?

amyblais: @streamer45 @agnivade Can you help describe the issue that was fixed from a user's (or admin's) point of view? For https://mattermost.atlassian.net/browse/MM-21481

streamer45: As far as I know the issue was triggered when a user update event was to be sent to a guest user in a cluster setup. So, from an admin point of view, if I had guest accounts enabled and a cluster setup I would definitely upgrade to avoid this.

@agnivade might have more info on this since he conducted the investigation and fixed the issue but he's currently OoO.

jasonblais: @amyblais @streamer45 Thank you! Definitely an improvement.

Would there be an opportunity to describe the issue in simpler terms? In particular, when I read `Fixed an issue with a panic in ``web.shouldSendEventToGuest`` triggered in cases where a user update event was sent`, I don't really understand the impact. Does the app crash? Does the server crash? Does it prevent a guest from receiving notifications? Something else?

streamer45: @jasonblais 

The server crashed when the specific case occurred (`user_updated` event sent through a cluster to a guest user) and the event would not reach the guest user.

source/administration/changelog.md
```diff
@@ -70,6 +70,8 @@ Mattermost v5.19.0 contains low to high level security fixes. [Upgrading](http:/
 
 Mattermost v5.18.0 contains low to high level security fixes. [Upgrading](http://docs.mattermost.com/administration/upgrade.html) is recommended. Details will be posted on our [security updates page](https://about.mattermost.com/security-updates/) 30 days after release as per the [Mattermost Responsible Disclosure Policy](https://www.mattermost.org/responsible-disclosure-policy/).
 
+- **v5.18.2, released 2020-01-16**
+  - Fixed an issue where the server crashed when a ``user_updated`` event was sent through a cluster to a guest user and the event did not reach the guest user. [MM-21481](https://mattermost.atlassian.net/browse/MM-21481)
```
jasonblais: @amyblais @streamer45 Thank you for the updates, I think we're almost there. Two questions that come to mind for an IT admin deciding to upgrade:

1-Does this always happen when the ``user_updated`` event is sent to a guest user, or only some of the time - if some of the time, is there an example?
2-What's the most common ``user_updated`` event, for those who don't know what it is. E.g. I'm thinking of `Fixed an issue ... when a ``user_updated`` event, such as an update to account settings, was sent through ...`

streamer45: 1. To properly answer this I would have to debug the issue on a cluster since I wasn't involved in the fixing process for this issue. By just looking at the code the crash would always happen as long as there's a guest user connected when another user is being updated and they are not connected to the same server for that cluster. So it's not something that will happen 100% of the times but could happen fairly frequently if a lot of guest users are connected.

2. Changing almost anything from the account settings I would say (username, nickname, firstname, lastname, profile picture etc.)

jasonblais: @streamer45 Thank you! How would this sound like:

```
Fixed an issue where the server crashed when a guest user updated their Account Settings in a high availability cluster environment and the corresponding ``user_updated`` event did not reach the guest user. [MM-21481](https://mattermost.atlassian.net/browse/MM-21481)
```

With this we start by explaining in non-technical terms on what happens, when, in which environment - helping the admin understand the issue - and a short description on the technical why.

streamer45: @jasonblais Looks good but it's not important that the user that gets updated is a guest. Any user that gets updated will generate that event. It's the recipient of the event that needs to be a guest for the problem to occur.

jasonblais: Got it! I removed `guest` from the first part, is the below correct?

```
Fixed an issue where server crashed when a user updated their Account Settings in a high availability cluster environment, and the corresponding ``user_updated`` event did not reach a guest user. [MM-21481](https://mattermost.atlassian.net/browse/MM-21481)
```

streamer45: LGTM :+1:

source/administration/changelog.md
agnivade: @streamer45 has explained it perfectly. I don' have anything else to add.

source/administration/changelog.md
agnivade: > By just looking at the code the crash would always happen as long as there's a guest user connected when another user is being updated and they are not connected to the same server for that cluster. 

Yes, that's right.

PR #3321, opened 2020-01-18
Slight tweak to 5.19 changelog to specify desktop app

jasonblais:


amyblais:
I would not add it because it's not desktop app specific issue. E.g users on RN v1.27 are seeing this issue if they're on the latest server versions.

PR #3324, opened 2020-01-21

amyblais:
Adding a breaking change note for pre-packaged plugins.


amyblais:
Adding to changelog as well,

PR #3327, opened 2020-01-23
Adding Customer Thermometer to docs

justinegeffen:
Adding a rating functionality to the documentation.

### DIFF
--- a/source/templates/customer_thermometer.html
+++ b/source/templates/customer_thermometer.html
@@ -22,6 +22,7 @@
 		margin-left: 5px;
 	}
 </style>
+
 <script type="text/javascript">
 	function txt(t) {
 		document.getElementById('rating').innerHTML = t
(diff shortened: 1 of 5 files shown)


justinegeffen:
Putting together the content from: https://github.com/mattermost/docs/blob/pr-customer-thermometer/source/templates/customer_thermometer.html and although it's showing in the source code, it's not displaying in the output. 

@icelander, do you have any idea why this might be? I am able to display it on the index page though. See below: 
Screen Shot 2020-01-23 at 15 42 47

Screenshot from local build source code for reference: Screen Shot 2020-01-23 at 15 40 07

jasonblais:
cc @asaadmahmood on above as well in case you have thoughts.

Thanks @justinegeffen for this! One additional question I have is what are your thoughts on using emojis instead of coloured dots to represent the rating? The emojis might more clearly convey emotion from the visitor. 0/5

justinegeffen:
I'm 5/5 for using emojis - currently setting that up and getting new URLs for the feedback responses.

justinegeffen:
Closing this PR as we're going with a custom UI as per: MM-23098

PR #3335, opened 2020-01-27
Update sso-saml-okta.rst

justinegeffen:
Changes to configuration steps for mattermost/mattermost-webapp#4427 and as per https://mattermost.atlassian.net/browse/MM-21843.

### DIFF
--- a/source/deployment/sso-saml-okta.rst
+++ b/source/deployment/sso-saml-okta.rst
@@ -57,9 +57,7 @@ Set Up a Connection App for Mattermost SSO
 
 	.. image:: ../../source/images/okta_8_view_instructions.PNG
 
-11. Take note of **Identity Provider Single Sign-On URL** (also known as **SAML SSO URL**), and the Identity Provider Issuer, as both may be needed to configure SAML for Mattermost.
-
-Furthermore, you **must download the X.509 Certificate file** and save it. You will need to upload it to Mattermost at a later step.
+11. Take note of **Identity Provider Single Sign-On URL** (also known as **SAML SSO URL**), and the Identity Provider Issuer, as both may be needed to configure SAML for Mattermost. Additionally, ensure you **download the X.509 Certificate file** and save it. You may need to upload it to Mattermost at a later step.
 
 	.. image:: ../../source/images/okta_9_view_instructions.PNG
 
--- a/source/deployment/sso-saml-okta.rst
+++ b/source/deployment/sso-saml-okta.rst
@@ -66,13 +66,15 @@ Furthermore, you **must download the X.509 Certificate file** and save it. You w
 Configure SAML Sign-in for Mattermost
 --------------------------------------
 
-1. Start Mattermost server and sign into Mattermost as a System Administrator. Go to **System Console > Authentication > SAML**, and enter the following fields:
+Start Mattermost server and sign into Mattermost as a System Administrator. Go to **System Console > Authentication > SAML**, and paste the copied Identity Provider Metadata URL in the **Identity Provider Metadata URL** field, then select **Get SAML Metadata from IdP**.
+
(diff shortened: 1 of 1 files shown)


source/deployment/sso-saml-okta.rst
```diff
@@ -16,59 +16,62 @@ Set Up a Connection App for Mattermost SSO
...
 
 1. Start Mattermost server and sign into Mattermost as a System Administrator. Go to **System Console > Authentication > SAML**, and enter the following fields:
- - **SAML SSO URL**: ``Identity Provider Single Sign-On URL`` from Okta, specified earlier.
- - **Identity Provider Issuer URL**: ``Identity Provider Issuer`` from Okta, specified earlier.
- - **Identity Provider Public Certificate**: X.509 Public Certificate file you downloaded from Okta earlier.
+ - **SAML SSO URL:** ``Identity Provider Single Sign-On URL`` from Okta, specified earlier.
+ - **Identity Provider Issuer URL:** ``Identity Provider Issuer`` from Okta, specified earlier.
+ - **Identity Provider Public Certificate:** X.509 Public Certificate file you downloaded from Okta earlier.
+ 
+ Alternatively, you can paste the copied Identity Provider Metadata URL in the **Identity Provider Metadata URL** and **Identity Provider Issuer URL** fields, then select **Get SAML Metadata from IdP**.
```
catalintomai: would remove this.

justinegeffen: Thanks, @catalintomai. I included that as it was in the original notes and matched up with the previous steps (as well as the UI). Can you confirm that using the metadata URL negates the need for the Identity Provider Issuer URL field to be populated?

catalintomai: We past the copied Identity Provider Metadata URL in the **Identity Provider Metadata URL** only. When we do that and we then click on the **Get SAML Metadata from IdP**. button, the SAML SSO URL and the Identity Provider Issuer URL fields are populated automatically (without user intervention). The Identity Provider Public Certificate is also downloaded from the server and set locally, without further user intervention. Bottom line, the fields still need to be populated (manually or otherwise), the newly introduced metadata Url allows us to fetch their content from the IdP server.

justinegeffen: Updated the docs with your feedback, thank you!

source/deployment/sso-saml-okta.rst
```diff
@@ -16,60 +16,65 @@ Set Up a Connection App for Mattermost SSO
...
 	.. image:: ../../source/images/okta_7_support_configuration.PNG
 
-9. Click **Finish**. On the next screen, click the **Sign On** tab and click **View Setup Instructions**.
+10. Click **Finish**. On the next screen, click the **Sign On** tab and click **View Setup Instructions**.
+  10.1 Select the **Identity Provider metadata** link and copy the link from the browser URL box. This will be used during the SAML configuration steps in the next section. 
 
 	.. image:: ../../source/images/okta_8_view_instructions.PNG
 
-10. Take note of ``Identity Provider Single Sign-On URL`` (also known as ``SAML SSO URL``), and the Identity Provider Issuer, as both will be needed to configure SAML for Mattermost.
+11. Take note of **Identity Provider Single Sign-On URL** (also known as **SAML SSO URL**), and the Identity Provider Issuer, as both will be needed to configure SAML for Mattermost.
```
catalintomai: unless they take the alternate metadataUrl route - maybe replace "will be" with "might be"?

justinegeffen: Thanks Catalin, will do that.

catalintomai: Forgot to mention - same observation for "Identity Provider Public Certificate"

justinegeffen: Thanks, done!

Issue #3336, opened 2020-01-27
Wrong iOS version in mobile app requirements


faase:
It's really just a typo in the [mobile app requirements](https://docs.mattermost.com/install/requirements.html#mobile-apps).

wrong: `iPhone 5s devices and later with iOS 12+`
correct: `iPhone 5s devices and later with iOS 11+`

thanks to @enahum for reporting it.

justinegeffen:
Thanks, @faase!

faase:
ok my bad! looks like i jumped the gun. just read this

> @enahum This was changed via https://github.com/mattermost/docs/pull/3087 based on our [software requirement guidelines](https://docs.mattermost.com/process/software-requirements.html#mobile-apps) that we support latest and next-to-latest versions.

_Originally posted by @amyblais in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MzQ4MzI2NjE3/pull_request_reviews/more_threads_

so it was correct in the mobile app requirements, but not in the change logs.

faase:
closing the issue

enahum:
@faase, @amyblais and @esethna so this means that we should then bump the supported OS to iOS 12+ and Android 9+? And then why does the changelog is different than the doc? I’m really confused here.

enahum:
Re-opening until we decide where to correct the disparity

amyblais:
@enahum I will fix the changelog for iOS.

Android version is correct since the guideline is any versions supported by Google.

enahum:
@amyblais is not only the changelog BUT also the app in itself.

amyblais:
@enahum What is the work involved in changing it in the app itself?

enahum:
@amyblais is just changing a setting, but do we bump it just like that?

actually there is this ticket in the backlog https://mattermost.atlassian.net/browse/MM-20643

amyblais:
I've merged the PR for now to change it back to iOS 11 since it's not changed in the app itself.

Will open a new discussion with PMs on how our software guidelines can be improved and planned ahead properly.

PR #3087, opened 2019-10-18
Update requirements.rst

amyblais:
- Change Chrome to 77+ https://en.wikipedia.org/wiki/Google_Chrome_version_history
- ~~Change Safari to 13+ https://en.wikipedia.org/wiki/Safari_version_history~~
- Change iOS to 12+ https://en.wikipedia.org/wiki/IOS_version_history
- ~~Combined OS versions because there are more than 1 or 2~~

PR #3337, opened 2020-01-27
Update requirements.rst

justinegeffen:
Typo corrected as per https://github.com/mattermost/docs/issues/3336

### DIFF
--- a/source/install/requirements.rst
+++ b/source/install/requirements.rst
@@ -56,7 +56,7 @@ Mobile Apps
 .. csv-table::
     :header: "Operating System", "Technical Requirement"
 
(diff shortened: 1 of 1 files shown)


faase:
as mentioned [here](https://github.com/mattermost/docs/issues/3336#issuecomment-579145011)

ok my bad! looks like i jumped the gun. just read this

@enahum This was changed via #3087 based on our software requirement guidelines that we support latest and next-to-latest versions.

Originally posted by @amyblais in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MzQ4MzI2NjE3/pull_request_reviews/more_threads

so it was correct in the mobile app requirements, but not in the change logs.

PR #3344, opened 2020-01-29
Add image proxy details to Outbound Proxy

bbodenmiller:

### DIFF
--- a/source/install/outbound-proxy.rst
+++ b/source/install/outbound-proxy.rst
@@ -3,7 +3,7 @@
 Using an Outbound Proxy
 =======================
 
-In some scenarios, you may wish to use Mattermost behind a proxy. This can be used to do things such as monitoring outbound traffic from Mattermost or controlling which websites can appear in link previews and other embedded content.
+In some scenarios, you may wish to use Mattermost behind a proxy. This can be used to do things such as monitoring outbound traffic from Mattermost or controlling which websites can appear in link previews and other embedded content. If you only want to use a proxy for images, the `image proxy <../administration/image-proxy.rst>`_ is also an option.
 
 Configuration
 -------------


hmhealey:
I don't think this is quite correct. The image proxy serves a different purpose from the server's proxy since the image proxy is to prevent clients from connecting to third party servers while the regular proxy is to prevent the server from making those connections. Thoughts on that, @bbodenmiller?

bbodenmiller:
You're welcome to improve this but I think it's still helpful. Most users
are likely on this page because their system requires usage of a proxy.
Image proxy in that case is another option that should be called out.

jasonblais:
@hmhealey thoughts on the above response from @bbodenmiller?

hmhealey:
Sorry, I didn't realize this slipped off my radar. My problem with this is that this document is primarily about using a proxy for the server, not the client, so it doesn't make sense to me to have the link in the middle of the document. Maybe we could add a section to the end of the document that's kind of "hey, if you want to set up an image proxy for clients, see here".

I just don't want it to sound like the server will use the image proxy for its own requests because we've had some confusion around that in the past where admins expected that the server would load its own images through the proxy.

jasonblais:
@bbodenmiller curious on your thoughts above?

bbodenmiller:
@hmhealey the image proxy is only used by clients? That's not at all clear by the doc:

> Using image proxies means that users make only secure requests

> When enabled, the image proxy needs to be publicly accessible to both the Mattermost client and server.

hmhealey:
The image proxy is mainly used by the clients, but I'm not sure if the server uses it atill. Even if it does use the image proxy for some requests, it still needs to request other data (such as OpenGraph metadata) from external servers directly since those requests can't be served by the image proxy.

PR #3358, opened 2020-02-05
Update sso-saml-technical.rst

justinegeffen:
Added SAML metadata URL overview for 5.20 release blog post.


thefactremains:
@catalintomai @sbishel Guys, this might be overkill, but let us know if this might be useful as a separate page.

justinegeffen:
Some context: This was written for inclusion in the 5.20 release announcement but I subsequently realized that it wasn't needed. If it's helpful to include it then that's great, otherwise we can just disregard it.

justinegeffen:
Thanks, @sbishel. :)

PR #3363, opened 2020-02-06
Update description of attachment fallback

hmhealey:
This is just clarifying the description of one of the message attachment fields

### DIFF
--- a/source/developer/message-attachments.rst
+++ b/source/developer/message-attachments.rst
@@ -12,7 +12,7 @@ Attachment Options
 
 When sending an attachment, you can use any of the following to format how you want the posted message to look.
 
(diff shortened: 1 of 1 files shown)


source/developer/message-attachments.rst
```diff
@@ -12,7 +12,7 @@ Attachment Options
 
 When sending an attachment, you can use any of the following to format how you want the posted message to look.
 
-``fallback``: A required plain-text summary of the post. This is used in notifications, and in clients that don’t support formatted text (eg IRC).
+``fallback``: A required plain-text summary of the attachment. This is used in notifications, and in clients that don’t support formatted text (eg IRC).
```
justinegeffen:
```suggestion
``fallback``: A required plain-text summary of the attachment. This is used in notifications, and in clients that don’t support formatted text (e.g. IRC).
```

PR #3375, opened 2020-02-11
Update LHS; Moved Troubleshooting

justinegeffen:
Moved "Troubleshooting" to its own LHS nav item for visibility.

### DIFF
--- a/source/install/troubleshooting.rst
+++ b/source/install/troubleshooting.rst
@@ -1,16 +1,11 @@
 ..  _troubleshooting:
 
-Troubleshooting
-===============
+This document summarizes common troubleshooting issues and techniques.
 
-This document summarizes common troubleshooting issues and techniques. 
-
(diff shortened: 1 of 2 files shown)


justinegeffen:
The Troubleshooting Guide does have sections similar to that but at some point it would be good to review it and make it flow better. I'll create an issue for it.

source/install/troubleshooting.rst
```diff
@@ -1,16 +1,11 @@
 ..  _troubleshooting:
 
-Troubleshooting
```
amyblais: Why do we remove the title?

justinegeffen: I moved the document within the administrator.rst file (from Installation to its own section called "Troubleshooting" . Doing that duplicated "Troubleshooting" as a header and then the title of the page was a sub-header, with the document contents being a third level of headers (among other formatting issues). Changing this and some of the heading formatting fixed the issue. 

Hope that makes sense and if there is a less convoluted way to have done that let me know! :)

PR #3380, opened 2020-02-12
Update: Changed "Administrator" to "Admin"

justinegeffen:
Fixed formatting of UI references, changed "Administrator" to "Admin" for consistency.


justinegeffen:
General comment:
In the steps we refer to "Click on your name in the Navigation Panel" but we also refer to "Click on the three-dot Main Menu". Both open the same menu. Should we change all "Navigation Panel" references to "Main Menu"?

Issue #3130, opened 2019-11-05
Request for Documentation: What happens when you change the configured "data retention" period?


matterdoc:
Mattermost user `george` from https://community-daily.mattermost.com has requested the following be documented:

```
Data retention runs once a day at the time specified in the config file. Changing the retention period does *not* automatically schedule any additional run of the data retention job. It just updates the configured period, which is honoured on the next scheduled data retention job run.
```

See the original post [here](https://community-daily.mattermost.com/_redirect/pl/6cxcyquyipr3jcs9bjjr1anjnw).

_This issue was generated from [Mattermost](https://mattermost.com) using the [Doc Up](https://github.com/jwilander/mattermost-plugin-docup) plugin._

PR #3386, opened 2020-02-12
Update data-retention: New FAQ

justinegeffen:
Update for: https://github.com/mattermost/docs/issues/3130.

### DIFF
--- a/source/administration/data-retention.rst
+++ b/source/administration/data-retention.rst
@@ -35,7 +35,7 @@ Frequently Asked Questions (FAQs)
 What happens when the data retention period is changed?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Data retention runs once a day at the time specified in the ``config.json`` file. Changing the retention period does not automatically schedule any additional run of the data retention job. It just updates the configured period, which is honored when the next scheduled data retention job is run.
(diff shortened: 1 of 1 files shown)


source/administration/data-retention.rst
```diff
@@ -32,6 +32,11 @@ You can also run the deletion job manually at any time by clicking **Run Deletio
 Frequently Asked Questions (FAQs)
 ---------------------------------
 
+What happens when the data retention period is changed?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Data retention runs once a day at the time specified in the ``config.json`` file. Changing the retention period does not automatically schedule any additional run of the data retention job. It just updates the configured period, which is honored when the next scheduled data retention job is run.
```
jasonblais:
```suggestion
Data retention runs once a day at the time specified in the ``config.json`` file. Changing the retention period does not automatically schedule any additional run of the data retention job - it only updates how long data is kept in Mattermost.
```

jasonblais: Propose simplifying the language

justinegeffen:
> Thanks Justine! Propose moving this to above `Does the System Administrator get any notification when the data retention period is changed?` question, rather than as the first question of the FAQ.
> 
> Otherwise just one minor comment

That's a much better place for it, I agree. Thanks!

PR #3393, opened 2020-02-14
Formatting fixes

bbodenmiller:

### DIFF
--- a/source/install/outbound-proxy.rst
+++ b/source/install/outbound-proxy.rst
@@ -24,7 +24,7 @@ Note that when proxying HTTPS resources, you will need to configure your Matterm
 
 The ``NO_PROXY`` environment variable can be set to prevent certain requests from going through the proxy, such as to an SSO provider (e.g. GitLab or SAML) or intranet sites that should be accessible in link previews. It can be configured as a set of comma-separated IP addresses (e.g. ``1.2.3.4``), IP address ranges specified in CIDR notation (e.g. ``1.2.3.4/8``), or domain names. An IP address or domain name can also include a port number.
 
-When specifying a domain name, that domain name also matches its sub domains as well. A domain name with a leading ``.`` matches only sub domains. For example, ``example.com`` matches both ``example.com`` as well as ``sub.example.com`` while ``.example.com`` only matches the latter.
+When specifying a domain name, ensure that domain name also matches its subdomains. A domain name with a leading ``.`` matches only subdomains. For example, ``example.com`` matches both ``example.com`` as well as ``sub.example.com`` while ``.example.com`` only matches the latter.
 
 Sample Configuration
 --------------------
--- a/source/install/outbound-proxy.rst
+++ b/source/install/outbound-proxy.rst
@@ -22,7 +22,7 @@ Note that when proxying HTTPS resources, you will need to configure your Matterm
 ``NO_PROXY``
 ~~~~~~~~~~~~
 
-The ``NO_PROXY`` environment variable can be set to prevent certain requests from going through the proxy, such as to an SSO provider (such as GitLab or SAML) or to intranet sites that should be accessible in link previews. It can be configured as a set of comma-separted IP addresses (e.g. 1.2.3.4), IP address ranges specified in CIDR notation (e.g. 1.2.3.4/8), or domain names. An IP address or domain name can also include a port number.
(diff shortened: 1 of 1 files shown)


source/install/outbound-proxy.rst
```diff
@@ -22,7 +22,7 @@ Note that when proxying HTTPS resources, you will need to configure your Matterm
 ``NO_PROXY``
 ~~~~~~~~~~~~
 
-The ``NO_PROXY`` environment variable can be set to prevent certain requests from going through the proxy, such as to an SSO provider (such as GitLab or SAML) or to intranet sites that should be accessible in link previews. It can be configured as a set of comma-separted IP addresses (e.g. 1.2.3.4), IP address ranges specified in CIDR notation (e.g. 1.2.3.4/8), or domain names. An IP address or domain name can also include a port number.
+The ``NO_PROXY`` environment variable can be set to prevent certain requests from going through the proxy, such as to an SSO provider (such as GitLab or SAML) or to intranet sites that should be accessible in link previews. It can be configured as a set of comma-separted IP addresses (e.g. ``1.2.3.4``), IP address ranges specified in CIDR notation (e.g. ``1.2.3.4/8``), or domain names. An IP address or domain name can also include a port number.
```
justinegeffen:
```suggestion
The ``NO_PROXY`` environment variable can be set to prevent certain requests from going through the proxy, such as to an SSO provider (e.g. GitLab or SAML) or intranet sites that should be accessible in link previews. It can be configured as a set of comma-separated IP addresses (e.g. ``1.2.3.4``), IP address ranges specified in CIDR notation (e.g. ``1.2.3.4/8``), or domain names. An IP address or domain name can also include a port number.
```
Should we have an example of an IP address or domain name with a port number included?

source/install/outbound-proxy.rst
```diff
@@ -22,7 +22,7 @@ Note that when proxying HTTPS resources, you will need to configure your Matterm
...
 
 When specifying a domain name, that domain name also matches its sub domains as well. A domain name with a leading ``.`` matches only sub domains. For example, ``example.com`` matches both ``example.com`` as well as ``sub.example.com`` while ``.example.com`` only matches the latter.
```
justinegeffen:
```suggestion
When specifying a domain name, ensure that domain name also matches its subdomains. A domain name with a leading ``.`` matches only subdomains. For example, ``example.com`` matches both ``example.com`` as well as ``sub.example.com`` while ``.example.com`` only matches the latter.
```

source/install/outbound-proxy.rst
bbodenmiller: Idk. I think either way people will get the point.

bbodenmiller:
@justinegeffen addressed but it still says changes are requested so idk what's going on here.

justinegeffen:
> @justinegeffen addressed but it still says changes are requested so idk what's going on here.

I think I sorted that out, thanks for letting me know!

source/install/outbound-proxy.rst
```diff
@@ -22,17 +22,17 @@ Note that when proxying HTTPS resources, you will need to configure your Matterm
...
+The ``NO_PROXY`` environment variable can be set to prevent certain requests from going through the proxy, such as to an SSO provider (e.g. GitLab or SAML) or intranet sites that should be accessible in link previews. It can be configured as a set of comma-separated IP addresses (e.g. ``1.2.3.4``), IP address ranges specified in CIDR notation (e.g. ``1.2.3.4/8``), or domain names. An IP address or domain name can also include a port number.
 
-When specifying a domain name, that domain name also matches its sub domains as well. A domain name with a leading ``.`` matches only sub domains. For example, ``example.com`` matches both ``example.com`` as well as ``sub.example.com`` while ``.example.com`` only matches the latter.
+When specifying a domain name, ensure that domain name also matches its subdomains. A domain name with a leading ``.`` matches only subdomains. For example, ``example.com`` matches both ``example.com`` as well as ``sub.example.com`` while ``.example.com`` only matches the latter.
```
hmhealey: This isn't quite correct since we're trying to show the difference between using `example.com` and `.example.com`. How about this?
```suggestion
When a domain name is specified, the domain and all of its subdomains are matched, however a domain name with a leading ``.`` only matches the subdomains. For example, ``example.com`` matches both ``example.com`` and ``sub.example.com`` while ``.example.com`` only matches the latter.
```

PR #3394, opened 2020-02-14
No image proxy for preview of plaintext URLs

bbodenmiller:
#### Summary

Document https://github.com/mattermost/mattermost-server/issues/11857

### DIFF
--- a/source/administration/image-proxy.rst
+++ b/source/administration/image-proxy.rst
@@ -13,7 +13,7 @@ also helps preserve posts by protecting them from dead images.
 
 When enabled, the image proxy needs to be publicly accessible to both the Mattermost client and server.
 
-Currently the image proxy on the server side only works for Markdown embedded images and not for `image preview of plaintext URLs <https://github.com/mattermost/mattermost-server/issues/11857>`_.
+The Mattermost clients will use the image proxy to load all external images. The Mattermost server will use the image proxy when possible, but it will not use it when requesting content that may not be an image, such as for `image previews of plaintext URLs <https://github.com/mattermost/mattermost-server/issues/11857>`_.
 
 You may alternatively use `atmos/camo <https://github.com/atmos/camo>`_ http proxy to route images through SSL:
 
--- a/source/administration/image-proxy.rst
+++ b/source/administration/image-proxy.rst
@@ -3,7 +3,7 @@
 Image Proxy
 ================================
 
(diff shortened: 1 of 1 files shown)


source/administration/image-proxy.rst
```diff
@@ -13,6 +13,8 @@ also helps preserve posts by protecting them from dead images.
 
 When enabled, the image proxy needs to be publicly accessible to both the Mattermost client and server.
 
+Currently the image proxy only works for Markdown embedded images and not `image preview of plaintext URLs <https://github.com/mattermost/mattermost-server/issues/11857>`_.
```
amyblais:
```suggestion
Currently the image proxy only works for Markdown embedded images and not for `image preview of plaintext URLs <https://github.com/mattermost/mattermost-server/issues/11857>`_.
```

source/administration/image-proxy.rst
```diff
@@ -3,7 +3,7 @@
 Image Proxy
 ================================
 
-Using image proxies means that users make only secure requests. By allowing image requests to go straight to third-party
+Using image proxies means that users make only secure requests for Markdown embedded images. By allowing image requests to go straight to third-party
```
hmhealey: We could change "users" to "clients" to make this a bit clearer, but the changes to this line are incorrect. Clients use the image proxy for all image requests which is what this document is talking about, but the issue linked below refers specifically to server requests.

bbodenmiller: Okay how would you update then to accurately document linked bug? It sounds like while the client may get the image from the Mattermost server the sever is not using the image proxy but rather grabbing and serving image directly.

hmhealey: The way it currently works in more detail is that:
1. When a Markdown image or plain text link is posted, the server hits the URL directly to see what type of content it is
2. The server tells the clients what type of content is in the link
3. The client requests the image from the server and the server redirects the request through the proxy if necessary.

The first request is the problem there since it's before the server knows if it can use the image proxy. Maybe we can clarify that the server needs to make that first request before it confirms the data?

It's been a while since I've last looked at that bug, and I'm not quite sure if it's a bug. It might not be ideal, but when we've thought about improving it in the past (such as making non-image requests through the image proxy), it's come down to basically replacing the image proxy with a full proxy.

bbodenmiller: Updated accordingly. Please feel free to edit as you see fit.

PR #3417, opened 2020-02-26
Fix Cluster Encryption Version

DSchalla:
#### Summary
Update the version for the cluster encryption transport

### DIFF
--- a/source/install/transport-encryption/config.rst
+++ b/source/install/transport-encryption/config.rst
@@ -8,7 +8,7 @@ of the Mattermost cluster, all transports rely on TLS encryption.
 
 **Note:** The transport between the Application servers is not used by default and
 requires additional setup steps. Enhancing the core product to include automatic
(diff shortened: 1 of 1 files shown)


source/install/transport-encryption/config.rst
```diff
@@ -8,7 +8,7 @@ of the Mattermost cluster, all transports rely on TLS encryption.
 
 **Note:** The transport between the Application servers is not used by default and
 requires additional setup steps. Enhancing the core product to include automatic
-encryption between cluster nodes is in progress and planned for the 5.20 release.
+encryption between cluster nodes is in progress and planned for a later release.
```
jasonblais: Is there a Jira ticket we can link to, so users can follow progress if they're interested?

DSchalla: I think it's currently marked as internal, so I fear not. Good suggestion though!

DSchalla:
Sync'd with George on Mattermost

PR #3429, opened 2020-03-04
Update command-line-tools.rst

### LINKED ISSUE #3236: Add note to config in database docs about CLI usage
Just got off a call with a customer that had database connection errors when running the CLI commands after they ingested the config into the database as described here: https://github.com/mattermost/docs/blob/master/source/administration/config-in-database.rst

They are supplying the Mattermost service with the envvars using a systemd envvar file but the CLI of course does not have access to this. We should add a note to the troubleshooting steps or anywhere else in this document to make this clear. In their case the server was trying to connect to the dockerhost because it was reading the config.json when running CLI commands as there was no `--config` parameter appended to the command nor was the `MM_CONFIG` variable set.

svelle:
Fixes #3236

### DIFF
--- a/source/administration/command-line-tools.rst
+++ b/source/administration/command-line-tools.rst
@@ -56,7 +56,7 @@ To run the CLI commands, you must be in the Mattermost root directory. On a defa
   Ensure you run the Mattermost binary as the ``mattermost`` user. Running it as ``root`` user (for example) may cause complications with permissions as the binary initiates plugins and accesses various files when running CLI commands. Running the server as ``root`` may result in ownership of the plugins and files to be overwritten as well as other potential permissions errors.
   
 .. note::
-When running Mattermost with the configuration stored in the database you might need to pass the database connection string when running CLI commands which can do this as follows: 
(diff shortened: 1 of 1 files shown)


source/administration/command-line-tools.rst
```diff
@@ -54,6 +54,11 @@ To run the CLI commands, you must be in the Mattermost root directory. On a defa
 
 .. note::
   Ensure you run the Mattermost binary as the ``mattermost`` user. Running it as ``root`` user (for example) may cause complications with permissions as the binary initiates plugins and accesses various files when running CLI commands. Running the server as ``root`` may result in ownership of the plugins and files to be overwritten as well as other potential permissions errors.
+  
+.. note::
+When running Mattermost with the configuration stored in the database you might need to pass the database connection string when running CLI commands which can do this as follows:
```
justinegeffen:
```suggestion
When running Mattermost and using the CLI with the configuration stored in the database you might need to pass the database connection string as follows: 
```

svelle:
@justinegeffen sounds good, committed ;)

justinegeffen:
Awesome, thanks @svelle! Adding PM reviewer for visibility.

PR #3432, opened 2020-03-04
s/Requires=/BindsTo=/g to let Mattermost hold the state the db holds

flexo3001:
With `Requires=` Mattermost will stop if the db service stops but it will not start again because it does not hold the active and inactive state from service it requires. With BindTo Mattermost will be stopped and started and therefore holds the state the db holds. db needs update -> db stops -> mattermost stops -> db updates -> db starts -> mattermost starts.

### DIFF
--- a/source/install/install-debian-mattermost.rst
+++ b/source/install/install-debian-mattermost.rst
@@ -71,7 +71,7 @@ Assume that the IP address of this server is 10.10.10.2.
     Description=Mattermost
     After=network.target
     After=postgresql.service
-    Requires=postgresql.service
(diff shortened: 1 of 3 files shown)


justinegeffen:
Adding @hmhealey for visibility and feedback.

hmhealey:
Re-assigning the review since I don't know much about configuring a service. I think @icelander would have a better idea of what to look for here?

justinegeffen:
Following up on this, @icelander, let me know if I can help with anything or you have any questions!

PR #3453, opened 2020-03-16
Clarifying TE & E10 scope

it33:
We need to update the SKU definitions, working with Ken in marketing, for now, adding some adjustments moving to the right direction.

### DIFF
--- a/source/overview/product.rst
+++ b/source/overview/product.rst
@@ -37,7 +37,7 @@ A free-to-use, open source, self-hosted alternative to proprietary SaaS messagin
 - Multi-factor authentication
 - Highly customizable `third party bots, integrations <https://about.mattermost.com/community-applications/#publicApps>`__, and `command line tools <https://docs.mattermost.com/administration/command-line-tools.html>`__
 - Extensive integration support via `webhooks, APIs, drivers <https://docs.mattermost.com/guides/integration.html>`__ and `third party extensions <https://about.mattermost.com/default-app-directory/>`__
-- Easily scalable from dozens to hundreds of users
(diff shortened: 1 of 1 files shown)


jasonblais:
@lfbrock FYI

source/overview/product.rst
```diff
@@ -37,7 +37,7 @@ A free-to-use, open source, self-hosted alternative to proprietary SaaS messagin
 - Multi-factor authentication
 - Highly customizable `third party bots, integrations <https://about.mattermost.com/community-applications/#publicApps>`__, and `command line tools <https://docs.mattermost.com/administration/command-line-tools.html>`__
 - Extensive integration support via `webhooks, APIs, drivers <https://docs.mattermost.com/guides/integration.html>`__ and `third party extensions <https://about.mattermost.com/default-app-directory/>`__
-- Easily scalable from dozens to hundreds of users
+- Easily scalable to dozens of users per team, and hundreds of teams
```
lfbrock, line 40: Can we cut "hundreds of teams"? I believe "Team Edition" is meant for small teams that can be managed by social norms rather than controls, I'm not sure hundreds of teams fits that description.

justinegeffen, line 40:
```suggestion
- Easily scalable to dozens of users per team. 
```

justinegeffen: The suggested change is also more consistent with the E10 description of: "Scale to handle to hundreds of users per team" without mention of number of teams.

it33: Love it, much better thanks!

PR #3455, opened 2020-03-16
Fix broken links

amyblais:


amyblais:
Added a few other fixes. Now this is good for review.

PR #3469, opened 2020-03-20
Adding NPS rating for docs pages

asaadmahmood:
#### Summary
Adding NPS rating for docs pages

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23098

PR #3495, opened 2020-04-06
Revert "Adding NPS rating for docs pages"

justinegeffen:
Reverts mattermost/docs#3469


justinegeffen:
Reverting due to some sizing and performance changes that need to be made.

PR #3500, opened 2020-04-08
update to add TLS vhost configuration

Pafzedog:
add mod_header and RequestHeader directives to get public links with HTTPS when Mattermost is not set with TLS itself

#### Summary
Add Vhost apache configuration with TLS and add directives to get public link in HTTPS 

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/14194

### DIFF
--- a/source/install/config-proxy-apache2.rst
+++ b/source/install/config-proxy-apache2.rst
@@ -11,10 +11,10 @@ The Apache2 proxy configuration is done through the ``/etc/apache2/sites-availab
 
 Copy the `default` configuration file found in the same directory.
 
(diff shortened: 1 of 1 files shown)


Pafzedog:
> I do see there is a separate page for configuring apache with tls - https://docs.mattermost.com/install/config-apache2.html#configuring-apache2-with-ssl-and-http-2-unofficial. Perhaps it's better added there to avoid duplication.

ok, I'll add modification on this page.

agnivade:
Thanks. Btw, you could have used the same PR for this. But no worries :+1:

PR #3509, opened 2020-04-11
Update feature parity table in product FAQ

jasonblais:


jasonblais:
@esethna Updated to full support :+1:

Issue #3472, opened 2020-03-23
Add documentation for changes to bulk import


svelle:
More of a note to myself.

We need to add documentation for the changes introduced to the bulk import with this PR:
https://github.com/mattermost/mattermost-server/pull/14034

Basically the post object now has a new attribute called props which can contain message attachments.

justinegeffen:
Awesome, thanks @svelle! This is a very helpful change. :)

justinegeffen:
Hi @svelle. Wanted to follow up and find out whether you've had a chance to update the docs for this? If not, let me know what the change is and I'll be happy to sort it out. :)

svelle:
> Hi @svelle. Wanted to follow up and find out whether you've had a chance to update the docs for this? If not, let me know what the change is and I'll be happy to sort it out. :) 

Hey @justinegeffen ,
I'm out for the long weekend because of Easter holidays. I'll get on to it this week (ideally before the release of the new server version ;))

justinegeffen:
Thanks @svelle! I've opened [this PR](https://github.com/mattermost/mattermost-server/pull/14034) in the meantime and will sync up with you when you're back. :)

PR #3520, opened 2020-04-14
Update bulk-export-data.rst with post prop information

svelle:
This PR updates the table for the post object with info on the new props attribute introduced in v5.22.

Related Issue: #3472 

PR for bulk exporter changes: https://github.com/mattermost/mattermost-server/pull/14034


svelle:
did some further changes and also added changes to another file which describes the bulk data format.

long term we might want to think about merging those two pages so we don't have to maintain two separate pages about the same topic with almost the same content :)

PR #3536, opened 2020-04-16
Fixing broken doc

amyblais:
The gif is too small.


justinegeffen:
Thanks, @amyblais! I'm wondering if removing the "35" scale from the two images in the previous section could be added to this PR? Although they may be that size on purpose for clarity, it is a bit jarring.

amyblais:
@justinegeffen Done!

justinegeffen:
Interesting - I did a quick Google and found this [info](https://docutils.sourceforge.io/docs/ref/rst/directives.html#image); sounds like the scale value doesn't actually impact the size of the image and the image itself is large. 

I can look into this and submit a separate PR for it. Otherwise LGTM. :)

PR #3544, opened 2020-04-20
v5.19.2 docs

amyblais:


source/administration/changelog.md
```diff
@@ -317,9 +317,9 @@ Multiple setting options were added to `config.json`. Below is a list of the add
 
 Mattermost v5.19.0 contains low to high level security fixes. [Upgrading](http://docs.mattermost.com/administration/upgrade.html) is recommended. Details will be posted on our [security updates page](https://about.mattermost.com/security-updates/) 30 days after release as per the [Mattermost Responsible Disclosure Policy](https://www.mattermost.org/responsible-disclosure-policy/).
 
-- **v5.19.2, release day TBD**
+- **v5.19.2, released 2020-04-21**
   - Direct access to the ``Props`` field in the ``model.Post`` structure has been deprecated. The available ``GetProps()`` and ``SetProps()`` methods should now be used. Also, direct copy of the ``model.Post`` structure must be avoided in favor of the provided ``Clone()`` method. [MM-21378](https://mattermost.atlassian.net/browse/MM-21378)
```
jasonblais, line 321: What does this mean? What's the impact to decide as an admin if I should update?

amyblais: @streamer45 Is there any additional text we can add here?

streamer45: The previous behaviour was not *safe* and could led to unexpected crashes like the one reported by a community member [here](https://github.com/mattermost/mattermost-server/issues/13797).  
As an admin, you should really consider updating. The message is mainly aimed at developers which need to use the new provided semantics to interact with post properties from now on to avoid reoccurrences of the same issue.

jasonblais: @streamer45 @amyblais do we know if there are certain situations where the server may crash, or is it just randomly?

streamer45: @jasonblais 

It's really unpredictable. Anything internally that modifies post properties could cause the crash.
In the community reported case it was push notifications that triggered it.

PR #3552, opened 2020-04-23
Added GA tracking code to rating widget.

justinegeffen:
Google Analytics is not displaying any events information from the NPS widget. 

I have consulted some help sites and based on that, have added the site tag to the code. Here is one site that I looked at: https://support.google.com/analytics/thread/3259118?hl=en. I have also installed the GA debug tool and tested the current widget on docs.mattermost.com and no errors are listed for the widget. 

I've tested the widget on a local build and it works as expected. 

@ztrayner, if you know of another way to validate this prior to pushing live that'd be awesome. And if there's anything in the code that could affect your metrics please let me know. 

Any other feedback welcome.


ztrayner:
I don't have any concerns about it affecting existing metrics.  As far as I can tell, this should work correctly when pushed live.  👍

PR #3563, opened 2020-04-27
Update changelog.md

amyblais:
Adding a note for https://github.com/mattermost/mattermost-webapp/pull/5358

### DIFF
--- a/source/administration/changelog.md
+++ b/source/administration/changelog.md
@@ -6,6 +6,8 @@ Also see [changelog in progress](http://bit.ly/2nK3cVf) for the next release.
 
 ## Release v5.22 - [Feature Release](https://docs.mattermost.com/process/release-faq.html#release-overview)
 
+- **v5.22.2, release day TBD**
+  - Fixing an issue where ``Create posts`` permission is missing from the User Interface. There are no known issues that require this change but if there were issues, this change would provide customers with a means to diagnose and solve the problem. Workarounds include using the "Reset to Defaults" button for a system scheme and deleting and re-creating the button for team scheme.
 - **v5.22.1, released 2020-04-23**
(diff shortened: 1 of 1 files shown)


amyblais:
@jasonblais There are no other reports. It doesn't fix the customer issue but Martin found this as a related issue while troubleshooting the issue.

jasonblais:
If we have no other reports and it didn't fix the customer issue, I'd lean towards holding off until at least one of those occur. Shipping a dot release is extra worry and burden to the admin (which is fine if there is an issue, but not best for the customer if they don't need to worry about it)

source/administration/changelog.md
```diff
@@ -6,6 +6,8 @@ Also see [changelog in progress](http://bit.ly/2nK3cVf) for the next release.
 
 ## Release v5.22 - [Feature Release](https://docs.mattermost.com/process/release-faq.html#release-overview)
 
+- **v5.22.2, release day TBD**
+  - Fixing an issue where ``Create posts`` permission is missing from the User Interface. There are no known issues that require this change but if there were issues, this change would provide customers with a means to diagnose and solve the problem. Workarounds include using the "Reset to Defaults" button for a system scheme and deleting and re-creating the button for team scheme.
```
mkraft, line 10:
```suggestion
  - Exposing the ``Create Posts`` permission in the user interface in case the permission were missing. Previous workarounds included using the "Reset to Defaults" button for a system scheme or deleting and re-creating a team scheme.
```

mkraft:
> If we have no other reports and it didn't fix the customer issue, I'd lean towards holding off until at least one of those occur. Shipping a dot release is extra worry and burden to the admin (which is fine if there is an issue, but not best for the customer if they don't need to worry about it)

No objections from me.

PR #3579, opened 2020-04-30
Update backup.rst

funkytwig:
Added 'Please note that to make a 'clean' backup you need to stop mattermost during the duration of the backup otherwise the database and files may become out of sync.'  I did not realise that when I read the guide and think it needs mentioning.

### DIFF
--- a/source/administration/backup.rst
+++ b/source/administration/backup.rst
@@ -25,7 +25,7 @@ To backup your Mattermost server:
      - If you use local storage using a non-default directory specified in the ``Directory`` setting in ``config.json``, back up files in that location.
      - If you store your files in S3, you can typically keep the files where they are located without backup.
      
-Please note that to make a 'clean' backup you need to stop mattermost during the duration of the backup otherwise the database and files may become out of sync.
+Please note that to make a 'clean' backup you need to stop Mattermost during the duration of the backup otherwise the database and files may become out of sync.
 
 To restore a Mattermost instance from backup, restore your database, ``config.json`` file and optionally locally stored user files into the locations from which they were backed up. 
 
@@ -99,7 +99,6 @@ Once IT is contacted about an SSO outage issue, they can temporarily change a us
 If the administrator is unable to log into the System Console because of the SSO outage, they can switch their authentication method to email-password to gain access using the `command line tool <https://docs.mattermost.com/administration/command-line-tools.html>`__.
 
 It is highly important after the outage to switch everyone back to SSO from email-password to maintain consistency and security.
-
 
 
 
--- a/source/administration/backup.rst
+++ b/source/administration/backup.rst
@@ -25,6 +25,8 @@ To backup your Mattermost server:
      - If you use local storage using a non-default directory specified in the ``Directory`` setting in ``config.json``, back up files in that location.
      - If you store your files in S3, you can typically keep the files where they are located without backup.
      
+Please note that to make a 'clean' backup you need to stop mattermost during the duration of the backup otherwise the database and files may become out of sync.
+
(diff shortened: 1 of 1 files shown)


funkytwig:
OK, I've filled in the  'contribution license agreement'.  It's my first pull request ever, be gnetle :-).

source/administration/backup.rst
```diff
@@ -25,6 +25,8 @@ To backup your Mattermost server:
      - If you use local storage using a non-default directory specified in the ``Directory`` setting in ``config.json``, back up files in that location.
      - If you store your files in S3, you can typically keep the files where they are located without backup.
      
+Please note that to make a 'clean' backup you need to stop mattermost during the duration of the backup otherwise the database and files may become out of sync.
```
justinegeffen, line 28:
```suggestion
Please note that to make a 'clean' backup you need to stop Mattermost during the duration of the backup otherwise the database and files may become out of sync.
```

justinegeffen:
Hi @funkytwig. Thanks for the update, it's really helpful. I've made one suggestion to your change, and have included a Product Manager for visibility. 

One question I have is, are there any remediation steps if Mattermost isn't stopped and the files become out of sync?

funkytwig:
> 
> 
> Hi @funkytwig. Thanks for the update, it's really helpful. I've made one suggestion to your change, and have included a Product Manager for visibility.
> 
> One question I have is, are there any remediation steps if Mattermost isn't stopped and the files become out of sync?

Very good question.  From what I can see (from searching web) can't see anything about it.  What would be good to add would be something about what 'stopping' MM actually means. i.e. the easiest way of doing this.  I have a similar thing with ownCloud and put it into maintenance mode when I do backups.   Can't find anything similar for MM.  There is a docker pause but this may stop MM halfway through updating file/DB (unless it does this as a transaction type operation.

funkytwig:
Ive actually been working on a backup script, work in progress https://github.com/funkytwig/mattermost-docker/blob/master/backup.sh.  Think a full backup/restore infrastructure would be good.

justinegeffen:
Thanks, @funkytwig. I really like what you have here and the feedback around maintenance mode. I am looping in @grundleborg as it sounds like there is scope for a larger piece of work and he may be able to provide some additional guidance. 

In terms of editorial feedback, I am happy with this. :)

funkytwig:
Thanks, @justinegeffen.  So in terms of this PR we are good to go as its really just an added sentence to the doc? In terms of maintenance mode, where do we discuss this?  For reference in ownCloud rather than getting the type of generic connect error (you get when you stop the app/database, e.g. containers) it has an error in the interface.  It says the system has gone into maintenance mode. The icing on the cake would be to allow text saying why.  I.e. 'mattermost has been put into maintenance mode, to allow us to make a clean backup, it should be back very soon'. The bit before the comma is the fixed text, the comma and the rest is user-defined (could be an argument).

funkytwig:
In terms of backup/restore scripts, I will carry on working on these.  I'm going to make them generic but have hooks for mattermost (and other things).  I.e allows commands to be run after and before the backup.  At some later date we can decide if its a (maybe affiliated) side project or possibly folded into the docker-compose project at https://github.com/mattermost/mattermost-docker. Or both.

justinegeffen:
Thanks, @funkytwig. Apologies for the delayed reply. For this PR I think we're good to go. I would first like to make sure that you have some answers to your questions though, so I'm following up on that for you.

justinegeffen:
Thanks, @grundleborg! I'll merge this PR; @funkytwig, thanks again. Looking forward to working with you going forward. :) Let me know if I can help with anything.

Issue #3483, opened 2020-03-30
Request for Documentation: Add (optional) pgpass steps to postgres installation in docs


matterdoc:
Mattermost user `sven.huester` from https://community-daily.mattermost.com has requested the following be documented:

```
So .. postgres recommends the standard way to pass passwords is through a `.pgpass` file. It is a special file that HAS to be 0600 in permissions otherwise it doesnt work. One can also pass the location of the file using the PGPASS env variable.
This leaves the security to the user running the application. If you run it as a service, then you need sudo permissions to start a service file. Therefore, this file can be only root read-write and disabled for everybody else. After all, if you are root, there's nothing you can't do.
https://www.postgresql.org/docs/current/libpq-pgpass.html
```

See the original post [here](https://community-daily.mattermost.com/_redirect/pl/f98sb8g873fj7esw63kfiuhtkr).

_This issue was generated from [Mattermost](https://mattermost.com) using the [Doc Up](https://github.com/jwilander/mattermost-plugin-docup) plugin._

justinegeffen:
Thanks, @svelle. Do you have an idea of where this content should be added in the docs? I've taken a look through the original post but if you have more context that'd be great. :)

svelle:
Hey @justinegeffen that would be here: https://docs.mattermost.com/install/install-ubuntu-1604.html#installing-postgresql-database-server

and on the same section of the other OS's

justinegeffen:
Thanks, @svelle. Would you be open to updating that doc, and then I could replicate it across the other docs if it's one-size-fits-all?

svelle:
sure once I find the time :)

justinegeffen:
Thanks, @svelle! :) :)

fakela:
@justinegeffen has this doc been updated.

svelle:
@fakela that would be fantastic!
Unfortunately I haven't had the time to do so yet.

justinegeffen:
Thanks so much, @fakela, that would be awesome. Please let me or @svelle know how we can help. :)

PR #3603, opened 2020-05-10
Add (optional) pgpass steps to postgres installation in docs #3483

fakela:
#### Summary

This part of docs was updated to add pgpass steps to Postgres installation in docs

#### Ticket Link

If this pull request addresses a Help Wanted ticket, please link the relevant GitHub issue, e.g.

  Fixes https://github.com/mattermost/docs/issues/3483

This part of the docs is supposed to be in the same section of the other OS's but I want to confirm that this is added correctly before moving on

### DIFF
--- a/source/install/install-ubuntu-1604-postgresql.rst
+++ b/source/install/install-ubuntu-1604-postgresql.rst
@@ -104,7 +104,7 @@ Assume that the IP address of this server is 10.10.10.1
 13. (Optional) If you don't want to enter your password at each login, you can use a ``.pgpass`` file . It is a special file which has to be assigned ``0600`` in permissions otherwise it will not work. One can also pass the location of the file using the ``PGPASSFILE`` env variable.
 
 This leaves the security to the user running the application. If you run it as a service, then you need *sudo* permissions to start a service file. As a result only *root* user has read-write access. 
-
-For further info on the pgpass file please refer to `the official PostgreSQL documentation<https://www.postgresql.org/docs/9.4/libpq-pgpass.html>`_.
+ 
+For further info on the pgpass file please refer to `the official PostgreSQL documentation   <https://www.postgresql.org/docs/9.4/libpq-pgpass.html>`_
 
 With the database installed and the initial setup complete, you can now install the Mattermost server.
--- a/source/install/install-ubuntu-1604-postgresql.rst
+++ b/source/install/install-ubuntu-1604-postgresql.rst
@@ -100,5 +100,9 @@ Assume that the IP address of this server is 10.10.10.1
       You might have to install the PostgreSQL client software to use the command.
 
   The PostgreSQL interactive terminal starts. To exit the PostgreSQL interactive terminal, type ``\q`` and press **Enter**.
+  
+13. (Optional) Postgres recommends the standard way to pass passwords is through a ``.pgpass`` file. It is a special file that HAS to be 0600 in permissions otherwise it doesnt work. One can also pass the location of the file using the ``PGPASS`` env variable.
+
+  This leaves the security to the user running the application. If you run it as a service, then you need sudo permissions to   start a service file. Therefore, this file can be only root read-write and disabled for everybody else. After all, if you     are root, there's nothing you can't do
 
 With the database installed and the initial setup complete, you can now install the Mattermost server.
--- a/source/install/install-ubuntu-1604-postgresql.rst
+++ b/source/install/install-ubuntu-1604-postgresql.rst
@@ -101,7 +101,7 @@ Assume that the IP address of this server is 10.10.10.1
 
   The PostgreSQL interactive terminal starts. To exit the PostgreSQL interactive terminal, type ``\q`` and press **Enter**.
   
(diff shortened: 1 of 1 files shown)


source/install/install-ubuntu-1604-postgresql.rst
```diff
@@ -100,5 +100,9 @@ Assume that the IP address of this server is 10.10.10.1
       You might have to install the PostgreSQL client software to use the command.
 
   The PostgreSQL interactive terminal starts. To exit the PostgreSQL interactive terminal, type ``\q`` and press **Enter**.
+  
+13. (Optional) Postgres recommends the standard way to pass passwords is through a ``.pgpass`` file. It is a special file that HAS to be 0600 in permissions otherwise it doesnt work. One can also pass the location of the file using the ``PGPASS`` env variable.
```
justinegeffen, line 104:
```suggestion
13. (Optional) Postgres recommends the standard way to pass passwords is through a ``.pgpass`` file. It is a special file which has to be ``0600`` in permissions otherwise it will not work. One can also pass the location of the file using the ``PGPASS`` env variable.
```

source/install/install-ubuntu-1604-postgresql.rst
```diff
@@ -100,5 +100,9 @@ Assume that the IP address of this server is 10.10.10.1
...
+
+  This leaves the security to the user running the application. If you run it as a service, then you need sudo permissions to   start a service file. Therefore, this file can be only root read-write and disabled for everybody else. After all, if you     are root, there's nothing you can't do
```
justinegeffen, line 106:
```suggestion
This leaves the security to the user running the application. If you run it as a service, then you need *sudo* permissions to start a service file. As a result only *root* user has read-write access. 
```

justinegeffen:
Thanks, @fakela. This is another great PR, really appreciate it. I have made some minot suggestions related to formatting and wording. Other than those, this is a really helpful change!

source/install/install-ubuntu-1604-postgresql.rst
streamer45, line 104: As far as I understand using `.pgpass` is just a conveniency to remove the need to type the password at every login. Taking a quick look at the [official documentation](https://www.postgresql.org/docs/9.5/libpq-pgpass.html) it doesn't appear to be explicitly recommended nor a standard way to authenticate. Also the environment variable to pass would be `PGPASSFILE`.

fakela:
@justinegeffen @streamer45 Thanks a lot for the comments and reviews.
I made a change to the docs adding the suggestions from @streamer45 and also looking at the official docs

Here it is

13. (Optional) The file ``.pgpass``  is a convenient way to remove the need to type a password at every login. The file  ``.pgpass``which is located in the user’s home directory or the file cited by ``PGPASSFILE`` contains passwords to be used if the connection requires a password (and no password has been provided). ``.pgpass`` is a special file that has to be 0600 otherwise it will not work.

fakela:
Meanwhile, this is such a warm community always eager to respond and drop feedbacks. Thanks, @justinegeffen @streamer45. Looking forward to making more contributions

source/install/install-ubuntu-1604-postgresql.rst
justinegeffen: I've edited the content a bit to work with @streamer45's recommendations. @fakela, let me know what you think. If you're happy with the changes, you can commit them. Otherwise, more than happy to talk them over. :)

fakela: Looks alot simplified, I will commit them.
 Also are we adding this content on the same section of the other OS's. If yes I would open a seperate PR for them.

source/install/install-ubuntu-1604-postgresql.rst
```diff
@@ -100,5 +100,9 @@ Assume that the IP address of this server is 10.10.10.1
       You might have to install the PostgreSQL client software to use the command.
 
   The PostgreSQL interactive terminal starts. To exit the PostgreSQL interactive terminal, type ``\q`` and press **Enter**.
+  
+13. (Optional) If you don't want to enter your password at each login, you can use a ``.pgpass`` file . It is a special file which has to be assigned ``0600`` in permissions otherwise it will not work. One can also pass the location of the file using the ``PGPASSFILE`` env variable.
+
+This leaves the security to the user running the application. If you run it as a service, then you need *sudo* permissions to start a service file. As a result only *root* user has read-write access.
```
svelle, line 106: Maybe add a link to the official postgres docs like so for more info on the pgpass file as it's not mattermost specific and some details might change in the future. Other than that it looks good to me.

```suggestion
This leaves the security to the user running the application. If you run it as a service, then you need *sudo* permissions to start a service file. As a result only *root* user has read-write access. 

For further info on the pgpass file please refer to `the official PostgreSQL documentation<https://www.postgresql.org/docs/9.4/libpq-pgpass.html>`_.
```

fakela: @svelle this is a nice suggestion, will commit the changes

svelle: Perfect, thanks.
Now if @justinegeffen doesn't have any issues with wording this should be good to merge :)

PR #3605, opened 2020-05-11
Update custom-emoji.rst

justinegeffen:
Updates to doc including new sections and reorganization to improve rating.

PR #3608, opened 2020-05-11
Add files via upload

justinegeffen:
Added delete Custom Emoji


amyblais:
If this is for https://github.com/mattermost/docs/pull/3605, you can add it directly to ``justinegeffen-patch-3`` so it will show up in the same PR.

justinegeffen:
> If this is for #3605, you can add it directly to `justinegeffen-patch-3` so it will show up in the same PR.

Good point - didn't think of that! :)

PR #3628, opened 2020-05-19

amyblais:


source/administration/important-upgrade-notes.rst
```diff
@@ -7,6 +7,22 @@ Important Upgrade Notes
 +----------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | If you’re upgrading from a version earlier than... | Then...                                                                                                                                                          |
 +====================================================+==================================================================================================================================================================+
+| v5.24.0                                            | Some Prometheus metrics being removed to be replaced by better ones                                                                                              |
```
amyblais, line 10: @agnivade I will need some help with the Prometheus upgrade note

agnivade: Here is what I'd say:

```
The `mattermost_http_request_duration_seconds` histogram metric has been removed. This information was already captured by `mattermost_api_time`, which also contains the api handler name, HTTP method, and the response code.

As an example, if you are using this

`rate(mattermost_http_request_duration_seconds_sum{server=~"$var"}[5m]) /   rate(mattermost_http_request_duration_seconds_count{server=~"$var"}[5m])` to measure average call duration, it needs to be replaced with 
`sum(rate(mattermost_api_time_sum{server=~"$var"}[5m])) by (instance) /   sum(rate(mattermost_api_time_count{server=~"$var"}[5m])) by (instance)`.
```

source/administration/important-upgrade-notes.rst
```diff
@@ -7,6 +7,30 @@ Important Upgrade Notes
...
+|                                                    | MySQL: ``ALTER TABLE Reactions DROP PRIMARY KEY, ADD PRIMARY KEY (PostId, UserId, EmojiName);``                                                                  |
+|                                                    |                                                                                                                                                                  |
+|                                                    | Postgres: ``ALTER TABLE reactions DROP CONSTRAINT reactions_pkey, ADD PRIMARY KEY (PostId, UserId, EmojiName);``                                                 |
+|                                                    +------------------------------------------------------------------------------------------------------------------------------------------------------------------+                                                  
+|                                                    | The at-mentioning an LDAP group (E20) feature is not yet supported on mobile apps. In mobile apps, users who attempt to use the feature will see an error.       |
```
amyblais, line 32: @srkgupta @thefactremains I'm adding release note here about group mentions (E20) feature not supported on mobile apps. Will need more details here on what customers may experience on the mobile apps since the feature is missing.

srkgupta: Hi @amyblais 
The following statement needs to be changed as its not true and users will not see any errors:
`In mobile apps, users who attempt to use the feature will see an error.`

Instead it should be reworded to something like this:
`In mobile apps, users will not be able to see the group mentions in the autocomplete dropdown. Users will still receive notifications if they are part of any group mentioned. However the group mention keyword would not be highlighted. `

agnivade:
Oh btw @amyblais - I just realized that the Prometheus metric which got removed is only available on EE editions. TE edition is not affected by this. Should that be mentioned ?

amyblais:
@agnivade Added a note.

PR #3633, opened 2020-05-22
Update changelog.md

amyblais:
A customer couldn't tell if Postgres deprecation note was because 9.4 goes out of support or because some enhancement for 10 comes in August. Clarified if users should be upgrading now.


jasonblais:
@wiersgallak friendly reminder :)

Issue #3589, opened 2020-05-01
Request for documentation: radio button for interactive dialog


kaakaa:
Radio button have been supported for interactive dialog by https://github.com/mattermost/mattermost-server/issues/10465, but [`Interactive Dialogs` page](https://docs.mattermost.com/developer/interactive-dialogs.html#dialog-submission) doesn't have any mentions about `radio` type.

justinegeffen:
Hi @kaakaa, thanks for picking this up! Would you be open to updating the page with your feedback?

justinegeffen:
Following up on this, @kaakaa, to find out if you'd be open to updating the docs for this. I'm happy to answer any questions you have. :)

kaakaa:
I'm sorry for my late replay. @justinegeffen 
I'm interested in updating docs, however, I can't promise when I'm going to start working on it because I don't have much time right now.

justinegeffen:
That would be extremely helpful, thank you so much. :) I totally understand time constraints so if you are able to find the time that would be awesome. Please let me know if I can assist in any way. :)

PR #3635, opened 2020-05-24
Including documentation for radio element in interactive dialog

kaakaa:
#### Summary
Add description about `radio` type in interactive dialog.

#### Ticket Link
https://github.com/mattermost/docs/issues/3589


kaakaa:
As for [the explanation of `options`](https://github.com/mattermost/docs/pull/3635/files#diff-2cd2701b686698512b36aebea072fd15R282), I am using [the explanation of the existing Select Element](https://github.com/mattermost/docs/pull/3635/files#diff-2cd2701b686698512b36aebea072fd15R209), but isn't the type of `options` an `Array` instead of a `String`?

>     "options", "String", "(Optional) An array of options for the radio element."

If it looks good to you, I can fix it, but what do you think?

source/developer/interactive-dialogs.rst
```diff
@@ -240,6 +242,48 @@ The full list of supported fields are included below:
...
+        "value": "engineering"
+      },
+      {
+        "text": "Sales",
+        "value": "sales"
+      },
+      {
+        "text": "Administration",
+        "value": "administration"
+      }
```
hanzei, line 270: Given that the image above as a `default` element, it should also be added to this payload.

source/developer/interactive-dialogs.rst
```diff
@@ -240,6 +242,48 @@ The full list of supported fields are included below:
...
+    ]
```
hanzei, line 271: Non blocking: I would also love to see the example above being extended with a `help_text` to ensure that every supported field is shown.

source/developer/interactive-dialogs.rst
```diff
@@ -66,6 +66,8 @@ Each dialog supports elements for users to enter information.
 - **text**: Single-line plain text field. Use this for inputs such as names, email addresses or phone numbers.
 - **textarea**: Multi-line plain text field. Use this field when the answer is expected to be longer than 150 characters. 
 - **select**: Message menu. Use this for pre-selected choices. Can either be static menus or dynamic menus generated from users and public channels of the system. For more information on message menus, see :doc:`the documentation <interactive-messages>`.
+- **bool**: Checkbox option. Use this for binary selection.
```
hanzei, line 69: Are you also interested in documenting this type?

source/developer/interactive-dialogs.rst
```diff
@@ -240,6 +242,48 @@ The full list of supported fields are included below:
...
+
+The full list of supported fields are included below:
+
+.. csv-table::
+    :header: "Field", "Type", "Description"
+
+    "display_name", "String", "Display name of the field shown to the user in the dialog. Maximum 24 characters."
+    "name", "String", "Name of the field element used by the integration. Maximum 300 characters. You should use unique “name” fields in the same dialog."
+    "type", "String", "Set this value to ``radio`` for a radio element."
+    "options", "String", "(Optional) An array of options for the radio element."
```
hanzei, line 282: Correct, that is supposed to be an `Array`

source/developer/interactive-dialogs.rst
kaakaa: No, documentation about `bool` type had already been written.
https://docs.mattermost.com/developer/interactive-dialogs.html#checkbox-element

When writing about `radio type`, I noticed that `bool` is not here, so I just added it.

kaakaa:
Thanks @hanzei! I've updated as your comment 👍

source/developer/interactive-dialogs.rst
hanzei: Missed that, thanks for the explanation :+1:

source/developer/interactive-dialogs.rst
```diff
@@ -213,7 +215,7 @@ The list of supported fields for the ``select`` type element is included below:
 Checkbox Element
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-In Mattermost version 5.16 and later, you can use checkbox elements. It looks like a plain text field with a checkbox to be selected. Below is an example of a checkbox element that asks for meeting feedback:
+In Mattermost version 5.16 and later, you can use checkbox elements. It looks like a plain text field with a checkbox to be selected. Below is an example of a checkbox element that asks for meeting feedback.
```
justinegeffen, line 218:
```suggestion
From Mattermost 5.16 you can use checkbox elements. It looks like a plain text field with a checkbox to be selected. Below is an example of a checkbox element that asks for meeting feedback.
```

source/developer/interactive-dialogs.rst
```diff
@@ -240,6 +242,50 @@ The full list of supported fields are included below:
     "placeholder", "String", "(Optional) A string displayed to include a label besides the checkbox. Maximum 150 characters."
     
 
+Radio Element
+^^^^^^^^^^^^^^^^^^^^^^^
+
+In Mattermost version 5.16 and later, you can use radio elements. It looks like a plain text field with a radio button to be selected. Below is an example of a radio element that asks for a department.
```
justinegeffen, line 248:
```suggestion
From Mattermost 5.16 you can use radio elements. It looks like a plain text field with a radio button to be selected. Below is an example of a radio element that asks for a department.
```

justinegeffen:
Thanks, @kaakaa. This is super-helpful. :)

PR #3646, opened 2020-05-28
Remove unnecessary plugin directory rename step

gigawhitlocks:
#### Summary

This PR removes a step from the Mattermost Server upgrade instructions. Step 8 has the administrator rename the plugins directory and the client plugins directory, however this seems to be an unnecessary step since there is a `prepackaged_plugins` directory so that the prepackaged plugins are overwritten when the new version is copied in, but user plugins are left untouched.

I made a note to myself during upgrade of my server from 5.21 to 5.22 to test out the next upgrade while skipping Step 8 and my instance appears to have suffered no ill effects from the omission of this step during my upgrade from 5.22 to 5.23.

If there's any additional testing or verification that I need to do in order to further assuage any concerns about this change, please let me know and I will be happy to do so.

### DIFF
--- a/source/administration/upgrade.rst
+++ b/source/administration/upgrade.rst
@@ -146,16 +146,6 @@ Location of your local storage directory
 
    Your current settings are preserved, and new settings are added with default values.
 
-#. Reinstate the ``plugins`` directories, then restart the mattermost service.
-
-   .. code-block:: sh
-
-     cd {install-path}/mattermost
-     sudo rsync -au plugins~/ plugins
-     sudo rm -rf plugins~
-     sudo rsync -au client/plugins~/ client/plugins
-     sudo rm -rf client/plugins~
-
 After the server is upgraded, users might need to refresh their browsers to experience any new features.
 
 Upgrading Team Edition to Enterprise Edition
--- a/source/administration/upgrade.rst
+++ b/source/administration/upgrade.rst
@@ -97,13 +97,6 @@ Location of your local storage directory
 
      sudo find mattermost/ mattermost/client/ -mindepth 1 -maxdepth 1 \! \( -type d \( -path mattermost/client -o -path mattermost/client/plugins -o -path mattermost/config -o -path mattermost/logs -o -path mattermost/plugins -o -path mattermost/data \) -prune \) | sort | sudo xargs rm -r
     
-#. Rename the ``plugins`` directories so they do not interfere with the upgrade.
-
-   .. code-block:: sh
-
-     sudo mv mattermost/plugins/ mattermost/plugins~
-     sudo mv mattermost/client/plugins/ mattermost/client/plugins~
-    
 #. Change ownership of the new files before copying them.
 
    .. code-block:: sh


metanerd:
I am unqualified to review this.

metanerd:
https://github.com/mattermost/mattermost-plugin-starter-template/pull/97
Seems to address the same issue

source/administration/upgrade.rst
```diff
@@ -97,13 +97,6 @@ Location of your local storage directory
 
      sudo find mattermost/ mattermost/client/ -mindepth 1 -maxdepth 1 \! \( -type d \( -path mattermost/client -o -path mattermost/client/plugins -o -path mattermost/config -o -path mattermost/logs -o -path mattermost/plugins -o -path mattermost/data \) -prune \) | sort | sudo xargs rm -r
     
-#. Rename the ``plugins`` directories so they do not interfere with the upgrade.
```
lieut-data, line 100: The intent here is to relocate unmanaged plugins out of the way before reinstating them later on, e.g. see line 149:
```
#. Reinstate the ``plugins`` directories, then restart the mattermost service.
```

Unmanaged plugins are plugins are relatively rare: most plugins are prepackaged or managed in the filestore, with the `plugins` directory being a "working space" for both. However, it remains possible to manually put a plugin into the `plugins` directory and have the server consume /that/.

Perhaps this is something we might change for v6.0, and just deprecate support for unmanaged plugins altogether.

ali-farooq0: If an admin leaves `plugins` intact, the plugins directory gets clobbered right?

lieut-data: Ah, that's a good observation @ali-farooq0 -- in fact, I think to @gigawhitlocks's point, unmanaged plugins in this directory should remain untouched through the upgrade. So maybe we just remove the additional step at the end re: `Reinstate the plugins directories` and everything will be fine?

gigawhitlocks: Unmanaged plugins remain untouched and you don't need to move or reinstate them. That's what I observed in my testing/usage.

gigawhitlocks: The plugin directories are omitted in the `find` command and then `mv` is later done with no clobber so I was able to do an upgrade and just leave my unmanaged plugins in place.

gigawhitlocks: > deprecate support for unmanaged plugins altogether.

I hope this doesn't mean deprecate support for custom unpublished plugins :P

lieut-data: > I hope this doesn't mean deprecate support for custom unpublished plugins :P

No, definitely not -- just the "installation" strategy for same. @gigawhitlocks, if you want to remove the other half of this instruction from the `Reinstate the ``plugins`` directories` line, I think we're a go.

ali-farooq0: Excellent, done and done.

lieut-data:
Looping in @aaronrothschild for context before final merge.

justinegeffen:
@aaronrothschild, I think this is good to go pending your sign off. :)

PR #3663, opened 2020-06-03
Update feature quality level descriptions

lfbrock:
#### Summary

Proposing some definition updated after I received feedback that the current definitions were confusing. For example, the wording around beta features made it seem like they were not recommended for use in production. However, this does not match our actual usage of the term, as we actively encourage customers we partner with to use beta features so we can get feedback. 

Reviewers -- I think you have all used the beta label the most. Can you review these definitions and see if they match with how you have been using the term, or if there are other parts we should update?

### DIFF
--- a/source/overview/faq.rst
+++ b/source/overview/faq.rst
@@ -419,7 +419,7 @@ Beta Level Quality
 Experimental Level Quality
   - Not recommended for use in production
   - Unknown level of stability
-  - Minimal feature set that is not yet complete, and functionality may change overtime
+  - Minimal feature set that is not yet complete, and functionality may change over time
   - Minimal documentation is available
   - Disabled by default and must be turned on through a configuration setting
 
--- a/source/overview/faq.rst
+++ b/source/overview/faq.rst
@@ -413,7 +413,7 @@ Production Level Quality
 Beta Level Quality
   - `Premier Support <https://about.mattermost.com/support/>`__ is recommended for use in production environments
   - Core functionality is stable, but iteration based on feedback is ongoing
-  - Documentation may be incomplete
+  - Documentation may not be completed yet
   - Disabled by default and must be turned on through a configuration setting
   - Tested thoroughly on a small set of platforms, and made available for testing and feedback on additional platforms
 Experimental Level Quality
--- a/source/overview/faq.rst
+++ b/source/overview/faq.rst
@@ -421,7 +421,7 @@ Experimental Level Quality
   - Unknown level of stability
   - Minimal feature set that is not yet complete, and functionality may change over time
   - Minimal documentation is available
-  - Disabled by default and must be turned on through a configuration setting
+  - Disabled by default and must be enabled by modifying the configuration setting
 
 High Trust Questions
 ----------------------
--- a/source/overview/faq.rst
+++ b/source/overview/faq.rst
@@ -414,7 +414,7 @@ Beta Level Quality
   - `Premier Support <https://about.mattermost.com/support/>`__ is recommended for use in production environments
   - Core functionality is stable, but iteration based on feedback is ongoing
   - Documentation may not be completed yet
-  - Disabled by default and must be turned on through a configuration setting
+  - Disabled by default and must be enabled by modifying the configuration setting
   - Tested thoroughly on a small set of platforms, and made available for testing and feedback on additional platforms
 Experimental Level Quality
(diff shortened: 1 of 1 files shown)


source/overview/faq.rst
```diff
@@ -411,17 +411,17 @@ Production Level Quality
...
   - Tested on several platforms
 Beta Level Quality
-  - Is a work in progress
-  - Not recommended in production without `Premier Support <https://about.mattermost.com/support/>`__ engagement with Mattermost, Inc.
-  - Minimal documentation
-  - Disabled by default and must be turned on with a run-time feature flag
-  - Tested thoroughly on a small set of platforms, with tickets and work items drafted, and made available for testing and feedback on additional platforms
+  - `Premier Support <https://about.mattermost.com/support/>`__ is recommended for use in production environments
+  - Core functionality is stable, but iteration based on feedback is ongoing
+  - Documentation may be incomplete
```
aaronrothschild, line 416:
```suggestion
  - Documentation may not be completed yet
```
Indicates it is going to get done, but will be.

source/overview/faq.rst
```diff
@@ -411,17 +411,17 @@ Production Level Quality
...
+  - Disabled by default and must be turned on through a configuration setting
+  - Tested thoroughly on a small set of platforms, and made available for testing and feedback on additional platforms
 Experimental Level Quality
   - Not recommended for use in production
   - Unknown level of stability
-  - Minimal feature set that is not yet complete
-  - Little or no documentation available
-  - Disabled by default and must be turned on with a run-time feature flag
+  - Minimal feature set that is not yet complete, and functionality may change overtime
```
aaronrothschild, line 422:
```suggestion
  - Minimal feature set that is not yet complete, and functionality may change over time
```

source/overview/faq.rst
```diff
@@ -411,17 +411,17 @@ Production Level Quality
...
+  - Minimal documentation is available
+  - Disabled by default and must be turned on through a configuration setting
```
aaronrothschild, line 424:
```suggestion
  - Disabled by default and must be turned on using a configuration setting
```

justinegeffen: Thanks, @lfbrock and apologies for the delay. In that context "with" does read better. Another option could be: 
"Disabled by default and must be enabled by modifying the setting in ``config.json``". If appropriate that could be applied to all instances of this sentence in the doc.

source/overview/faq.rst
```diff
@@ -411,17 +411,17 @@ Production Level Quality
...
+  - Tested thoroughly on a small set of platforms, and made available for testing and feedback on additional platforms
 Experimental Level Quality
   - Not recommended for use in production
   - Unknown level of stability
-  - Minimal feature set that is not yet complete
-  - Little or no documentation available
-  - Disabled by default and must be turned on with a run-time feature flag
+  - Minimal feature set that is not yet complete, and functionality may change over time
+  - Minimal documentation is available
+  - Disabled by default and must be turned on through a configuration setting
```
justinegeffen, line 424:
```suggestion
  - Disabled by default and must be enabled by modifying the setting in ``config.json``
```

source/overview/faq.rst
```diff
@@ -411,17 +411,17 @@ Production Level Quality
...
 Beta Level Quality
-  - Is a work in progress
-  - Not recommended in production without `Premier Support <https://about.mattermost.com/support/>`__ engagement with Mattermost, Inc.
-  - Minimal documentation
-  - Disabled by default and must be turned on with a run-time feature flag
-  - Tested thoroughly on a small set of platforms, with tickets and work items drafted, and made available for testing and feedback on additional platforms
+  - `Premier Support <https://about.mattermost.com/support/>`__ is recommended for use in production environments
+  - Core functionality is stable, but iteration based on feedback is ongoing
+  - Documentation may not be completed yet
+  - Disabled by default and must be turned on through a configuration setting
```
justinegeffen, line 417:
```suggestion
  - Disabled by default and must be enabled by modifying the setting in ``config.json``
```

justinegeffen:
Following up on this to confirm thoughts around the suggested changes. Thanks!

lfbrock:
@justinegeffen - I like the change, let's just call it "modifying the configuration setting" rather than config.json though as I think they are moving towards config in the DB rather than config.json

PR #3690, opened 2020-06-15
Updated Skype for Business information

ahmaddanialmohd:
Included the link to the the plugin that is developed by Mattermost itself instead of korgz.

### DIFF
--- a/source/deployment/video-and-audio-calling.rst
+++ b/source/deployment/video-and-audio-calling.rst
@@ -37,10 +37,10 @@ BigBlueButton
 Skype for Business
 ~~~~~~~~~~~~~~~~~~~~
 
-- Start and join voice calls, video calls and use screensharing with your team members, developed by kosgrz.
+- Start and join voice calls, video calls and use screensharing with your team members, developed by Mattermost (Beta).
 - Clicking a video icon in a Mattermost channel invites team members to join a Skype for Business call, hosted using the credentials of the user who initiated the call.
 - Supports a vendor-hosted cloud solution.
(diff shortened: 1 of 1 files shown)


justinegeffen:
Looks good to me, editorially. Adding @jasonblais as I know he was working on this recently and he may have some feedback. Thanks!

source/deployment/video-and-audio-calling.rst
```diff
@@ -37,10 +37,10 @@ BigBlueButton
 Skype for Business
 ~~~~~~~~~~~~~~~~~~~~
 
-- Start and join voice calls, video calls and use screensharing with your team members, developed by kosgrz.
+- Start and join voice calls, video calls and use screensharing with your team members, developed by Mattermost (Beta).
```
jasonblais, line 40:
```suggestion
- Start and join voice calls, video calls and use screensharing with your team members, developed by kosgrz, and maintained by Mattermost.
```

jasonblais: Also, if you wanted to highlight that the plugin is beta for customer trust, that makes sense. Instead of adding it to the first line, we can add it to the title, `Sykpe for Business (Beta)`.

ahmaddanialmohd: Hi, @justinegeffen and @jasonblais 

Thank you for the feedback. Yes, I believe that retaining kosgrz as the original developer is a good idea.

And yes, the (Beta) on the title also works for me. All good.

PR #3695, opened 2020-06-16
remove production word from the docker setup page

svelle:
Apparently we do not support this deployment so we shouldn't use the word production anywhere on the page to not mislead people into thinking this is a recommended setup.


justinegeffen:
Great catch, thanks @svelle!

Issue #3705, opened 2020-06-19
Request for Documentation: Document the use of managed resources in the desktop app


matterdoc:
Mattermost user `guillermo.vaya` from https://community-release.mattermost.com has requested the following be documented:

```

[\[mattermost/desktop\]](https://github.com/mattermost/desktop) Pull request [#1311 [MM-25748]Allow access to managed resources](https://github.com/mattermost/desktop/pull/1311) was merged by @guillermo.vaya.

```

See the original post [here](https://community-release.mattermost.com/_redirect/pl/t7u7qe4o83fg3czjm5pgng41fc).

_This issue was generated from [Mattermost](https://mattermost.com) using the [Doc Up](https://github.com/jwilander/mattermost-plugin-docup) plugin._

[MM-25748]: https://mattermost.atlassian.net/browse/MM-25748

PR #3720, opened 2020-06-29
New document for managed resources

faase:
#### Summary
Adds a page for managed resources to the admin section of the docs.

#### Ticket Link
Related to this [ticket ](https://github.com/mattermost/docs/issues/3705)

### DIFF
--- a/source/install/desktop-managed-resources.rst
+++ b/source/install/desktop-managed-resources.rst
@@ -24,9 +24,9 @@ Here are two example server URLs each with valid and invalid managed resource UR
 
   - A valid conference service: ``https://mattermost.my.org/conference``
 
-  - An invalid video service: ``http://mattermost.my.org/video``
+  - An invalid video service using a different protocol: ``http://mattermost.my.org/video``
 
-  - An invalid conference service: ``https://conference.my.org``
+  - An invalid conference service having a different origin: ``https://conference.my.org``
 
 - Mattermost server: ``https://my.org/mattermost``
 
@@ -34,7 +34,7 @@ Here are two example server URLs each with valid and invalid managed resource UR
 
   - A valid conference service: ``https://my.org/conference``
 
-  - An invalid video service: ``http://my.org/video``
+  - An invalid video service using a different protocol: ``http://my.org/video``
   
-  - An invalid conference service: ``https://conference.my.org``
+  - An invalid conference service having a different origin: ``https://conference.my.org``
 
--- a/source/install/desktop-managed-resources.rst
+++ b/source/install/desktop-managed-resources.rst
@@ -1,0 +1,38 @@
+
+Desktop - Managed Resources
+===========================
+
+This guide provides steps to use managed resources with the Mattermost Desktop App. A managed resource can be any service available on the same hostname using the same protocol as the Mattermost server.
+
+.. note::
+    Using this feature requires a `custom build of the Mattermost Desktop App <https://docs.mattermost.com/deployment/desktop-app-deployment.html>`_.
+
+When adding the folder of a managed resource to `src/common/config/buildConfig.js <https://github.com/mattermost/desktop/blob/master/src/common/config/buildConfig.js>`_ it is available from the Mattermost Desktop App as a pop-up window instead of opening in a web-browser. In the below example we add the managed resource `/video`.
+
+.. code-block:: json
+    [...]
+    managedResources: ['trusted', 'video'],
+    [...]
+
+Here are two example server URLs each with valid and invalid managed resource URLs:
+
+- Mattermost server: ``https://mattermost.my.org``
+
+  - A valid video service: ``https://mattermost.my.org/video``
+
+  - A valid conference service: ``https://mattermost.my.org/conference``
+
+  - An invalid video service: ``http://mattermost.my.org/video``
+
+  - An invalid conference service: ``https://conference.my.org``
+
+- Mattermost server: ``https://my.org/mattermost``
+
+  - A valid video service: ``https://my.org/video``
+
+  - A valid conference service: ``https://my.org/conference``
+
+  - An invalid video service: ``http://my.org/video``
+  
+  - An invalid conference service: ``https://conference.my.org``
+
--- a/source/install/desktop-managed-resources.rst
+++ b/source/install/desktop-managed-resources.rst
@@ -7,7 +7,7 @@ The Mattermost Desktop App supports managed resources. A managed resource can be
 .. Note::
     Using this feature requires a `custom build of the Mattermost Desktop App <https://docs.mattermost.com/deployment/desktop-app-deployment.html>`_.
 
-When adding the path of a managed resource to `src/common/config/buildConfig.js <https://github.com/mattermost/desktop/blob/master/src/common/config/buildConfig.js>`_ it opens as a pop-up window in the Mattermost Desktop App. 
+Add the path of a managed resource to your configuration file. When selected, it opens as a pop-up window in the Mattermost Desktop App. 
 
 In the below example we add the managed resource ``/video``.
 
@@ -37,4 +37,3 @@ Here are two example server URLs each with valid and invalid managed resource UR
   - An invalid video service using a different protocol: ``http://my.org/video``
   
   - An invalid conference service having a different origin: ``https://conference.my.org``
-
--- a/source/install/desktop-managed-resources.rst
+++ b/source/install/desktop-managed-resources.rst
@@ -2,12 +2,14 @@
 Desktop - Managed Resources
 ===========================
 
-This guide provides steps to use managed resources with the Mattermost Desktop App. A managed resource can be any service available on the same hostname using the same protocol as the Mattermost server.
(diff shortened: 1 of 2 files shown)


source/install/desktop-managed-resources.rst
```diff
@@ -0,0 +1,38 @@
+
+Desktop - Managed Resources
+===========================
+
+This guide provides steps to use managed resources with the Mattermost Desktop App. A managed resource can be any service available on the same hostname using the same protocol as the Mattermost server.
```
justinegeffen, line 5: Will steps be added to the guide in the next iteration?

In the interim, one option could be: 

```suggestion
The Mattermost Desktop App supports managed resources. A managed resource can be any service available on the same hostname using the same protocol as the Mattermost server.
```

source/install/desktop-managed-resources.rst
```diff
@@ -0,0 +1,38 @@
...
+
+.. note::
+    Using this feature requires a `custom build of the Mattermost Desktop App <https://docs.mattermost.com/deployment/desktop-app-deployment.html>`_.
+
+When adding the folder of a managed resource to `src/common/config/buildConfig.js <https://github.com/mattermost/desktop/blob/master/src/common/config/buildConfig.js>`_ it is available from the Mattermost Desktop App as a pop-up window instead of opening in a web-browser. In the below example we add the managed resource `/video`.
```
justinegeffen, line 10:
```suggestion
When adding the folder of a managed resource to `src/common/config/buildConfig.js <https://github.com/mattermost/desktop/blob/master/src/common/config/buildConfig.js>`_ it opens as a pop-up window in the Mattermost Desktop App. In the below example we add the managed resource ``/video``.
```

justinegeffen: Is the example linked to the pop-up window, or is the pop-up window incidental to the example code? If they're unrelated one option could be to end this paragraph at the period and then start a new paragraph about the video resource.

source/install/desktop-managed-resources.rst
```diff
@@ -0,0 +1,38 @@
...
+
+.. code-block:: json
+    [...]
+    managedResources: ['trusted', 'video'],
+    [...]
+
+Here are two example server URLs each with valid and invalid managed resource URLs:
```
justinegeffen, line 17: Suggestion: Include an introduction or sentence around the resource URLs (it's not immediately clear how they're related to the example).

justinegeffen: Another option might be details around what makes the URL valid/invalid (looks like it's because it needs to be HTTPS).

faase: I've changed `folder` to `path`, would that make more sense now?

source/install/desktop-managed-resources.rst
```diff
@@ -0,0 +1,40 @@
+
+Desktop - Managed Resources
+===========================
+
+The Mattermost Desktop App supports managed resources. A managed resource can be any service available on the same hostname using the same protocol as the Mattermost server.
+
+.. Note::
+    Using this feature requires a `custom build of the Mattermost Desktop App <https://docs.mattermost.com/deployment/desktop-app-deployment.html>`_.
+
+When adding the path of a managed resource to `src/common/config/buildConfig.js <https://github.com/mattermost/desktop/blob/master/src/common/config/buildConfig.js>`_ it opens as a pop-up window in the Mattermost Desktop App.
```
justinegeffen, line 10: Would the user add the path to something in this repo? Or would it be in their own config file? If the latter, one suggestion is to replace the URL with ``buildConfig`` in the ``config.json`` file (replace with correct config setting  of course) and maybe a link to the config doc?

Willyfrog: that would be on their own clone of the repo, it shouldn't appear in a PR
the reason for it being in buildConfig is to prevent users from modifying as this is for admins to provide safe resources

source/install/desktop-managed-resources.rst
```diff
@@ -0,0 +1,40 @@
...
+
+Here are two example server URLs each with valid and invalid managed resource URLs:
+
+- Mattermost server: ``https://mattermost.my.org``
+
+  - A valid video service: ``https://mattermost.my.org/video``
+
+  - A valid conference service: ``https://mattermost.my.org/conference``
+
+  - An invalid video service: ``http://mattermost.my.org/video``
```
Willyfrog, line 27:
```suggestion
  - An invalid video service: ``http://mattermost.my.org/video`` (this is invalid as it is using a different protocol)
```

source/install/desktop-managed-resources.rst
```diff
@@ -0,0 +1,40 @@
...
+
+  - An invalid conference service: ``https://conference.my.org``
```
Willyfrog, line 29:
```suggestion
  - An invalid conference service: ``https://conference.my.org`` (this is invalid as it has a different origin)
```

source/install/desktop-managed-resources.rst
```diff
@@ -0,0 +1,40 @@
...
+
+- Mattermost server: ``https://my.org/mattermost``
+
+  - A valid video service: ``https://my.org/video``
+
+  - A valid conference service: ``https://my.org/conference``
+
+  - An invalid video service: ``http://my.org/video``
```
Willyfrog, line 37:
```suggestion
  - An invalid video service: ``http://my.org/video`` (this is invalid as it is using a different protocol)
```

source/install/desktop-managed-resources.rst
```diff
@@ -0,0 +1,40 @@
...
+  
+  - An invalid conference service: ``https://conference.my.org``
```
Willyfrog, line 39:
```suggestion
  - An invalid conference service: ``https://conference.my.org`` (this is invalid as it has a different origin)
```

source/install/desktop-managed-resources.rst
justinegeffen: Thanks, @Willyfrog. That makes sense! So they'd need to be a Mattermost System Admin in order to do this? Or general admin for their environment?

Willyfrog: A general admin in their org, someone who knows how's the network of their org setup to make the appropriate changes

justinegeffen: In that case, one option is to include an example of what they'd change and where. 

One other question is around the pop-up window. This seems like an important aspect of the content. Would it be an option to expand on it in the opening paragraph with an outline of the benefits or explanation of why one would want that?

source/install/desktop-managed-resources.rst
```diff
@@ -0,0 +1,40 @@
...
+===========================
+
+The Mattermost Desktop App supports managed resources. A managed resource can be any service available on the same hostname using the same protocol as the Mattermost server.
+
+.. Note::
+    Using this feature requires a `custom build of the Mattermost Desktop App <https://docs.mattermost.com/deployment/desktop-app-deployment.html>`_.
+
+When adding the path of a managed resource to `src/common/config/buildConfig.js <https://github.com/mattermost/desktop/blob/master/src/common/config/buildConfig.js>`_ it opens as a pop-up window in the Mattermost Desktop App. 
+
+In the below example we add the managed resource ``/video``.
```
justinegeffen, line 12: Suggestion:
As per the above comment, this would work really well if it was part of a larger example of where one would find this section as it would add some context and perhaps some clarity around the goal and outcome.

justinegeffen: "This" refers to the code example below.

source/install/desktop-managed-resources.rst
justinegeffen: I can't see where that it - is it part of the example?

justinegeffen: Is that in this example, or for the line "When adding the path of a managed resource to"?

source/install/desktop-managed-resources.rst
justinegeffen:
```suggestion
Add the path of a managed resource to your configuration file. When selected, it opens as a pop-up window in the Mattermost Desktop App. 
```

source/install/desktop-managed-resources.rst
justinegeffen: @Willyfrog, would you be open to reviewing the code below and confirming that it is accurate? If you would be able to confirm the path the user would follow to locate the area where they're adding the content that would also be great. :)

source/install/desktop-managed-resources.rst
justinegeffen: @Willyfrog, should we keep the examples below or should we trim them down to an example of one valid and one invalid?

source/install/desktop-managed-resources.rst
Willyfrog: Yes,  it is accurate 😊

source/install/desktop-managed-resources.rst
Willyfrog: I think those are fine, but if it feels like too much info they can be removed. I'm 0/5 on that.

Issue #3674, opened 2020-06-08
Add troubleshooting steps to MFA page


svelle:
https://docs.mattermost.com/deployment/auth.html?highlight=mfa

Has no troubleshooting steps

svelle:
One of the most common issues with MFA is when the system's time is out of sync with the real clock. 

If your server time is adrift too much the MFA code verification will stop working.

you can check this on systemd powered systems using `timedatectl`

faase:
@svelle I will add this to the [corresponding troubleshooting section](https://docs.mattermost.com/guides/administrator.html#troubleshooting). 

Are there any other issues i should cover?

PR #3722, opened 2020-06-29
Added MFA time sync solution to troubleshooting

faase:
#### Summary
Added a solution for an issue arising when the Mattermost server has an out of sync time.

#### Ticket Link
https://github.com/mattermost/docs/issues/3674

### DIFF
--- a/source/install/troubleshooting.rst
+++ b/source/install/troubleshooting.rst
@@ -132,7 +132,7 @@ The password reset process is completed on the IdP provider side, and not via th
 Multi-Factor Authentication (MFA)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If your MFA stopped working make sure your server clock not out of sync. On Unix systems you can show the hardware time of your system with ``# hwclock -r``.
+If your MFA stopped working make sure your server clock not out of sync. On Unix systems you can show the hardware time of your system with ``# hwclock -r``. If you are using systemd you can also use ``# timedatectl``.
 
 SAML issues
 -------------------
--- a/source/install/troubleshooting.rst
+++ b/source/install/troubleshooting.rst
@@ -132,7 +132,7 @@ The password reset process is completed on the IdP provider side, and not via th
 Multi-Factor Authentication (MFA)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If your MFA stopped working make sure your server clock not out of sync. On Unix systems you can show the hardware time of your system with ``# hwclock -r``. If you are using systemd you can also use ``# timedatectl``.
+If MFA has stopped working, confirm that your server clock is not out of sync. On Unix systems you can view the hardware time of your system using ``# hwclock -r``. If you're using systemd you can also use ``# timedatectl``.
 
 SAML issues
 -------------------
--- a/source/install/troubleshooting.rst
+++ b/source/install/troubleshooting.rst
@@ -129,6 +129,11 @@ If you experience issues with your password for your LDAP log in, contact your o
 
 The password reset process is completed on the IdP provider side, and not via the Mattermost System Console.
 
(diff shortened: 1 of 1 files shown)


source/install/troubleshooting.rst
```diff
@@ -129,6 +129,11 @@ If you experience issues with your password for your LDAP log in, contact your o
 
 The password reset process is completed on the IdP provider side, and not via the Mattermost System Console.
 
+Multi-Factor Authentication (MFA)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If your MFA stopped working make sure your server clock not out of sync. On Unix systems you can show the hardware time of your system with ``# hwclock -r``.
```
justinegeffen, line 135:
```suggestion
If MFA has stopped working, confirm that your server clock is not out of sync. On Unix systems you can show the hardware time of your system with ``# hwclock -r``.
```

justinegeffen: Would MFA stop working on all devices, and would there be other steps if the server clock is in sync and MFA still isn't working?

svelle: I'd additionally check with timedatectl since it might show more information than hwclock and is (afaik) the preferred method on systemd enabled systems.

faase: AFAIK `timedatectl ` is for system time where `hwclock` is for the hardware clock that is used as seed for the Unix system time. 

Do you know if the users had issues with their hardware clock or did their system time run out of sync due to high server up-time and no time synchronization? I assumed the former, but in case of the latter we can advise to use `date` (which should be POSIX compatible?) WDYT @svelle ?

svelle: `hwclock` isn't really useful here, since mattermost doesn't check with the hardware clock at all.

image

`date` would be more appropriate than `hwclock` but I'd still prefer timedatectl since it also shows RTC as well as info on time synchronization if enabled or not.

faase: @svelle Thanks for clarifying! This sounds like there is more then just checking local time. Do you know what information about the time is needed exactly in order to properly debug the issue?

source/install/troubleshooting.rst
```diff
@@ -129,6 +129,11 @@ If you experience issues with your password for your LDAP log in, contact your o
 
 The password reset process is completed on the IdP provider side, and not via the Mattermost System Console.
 
+Multi-Factor Authentication (MFA)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If your MFA stopped working make sure your server clock not out of sync. On Unix systems you can show the hardware time of your system with ``# hwclock -r``. If you are using systemd you can also use ``# timedatectl``.
```
justinegeffen, line 135:
```suggestion
If MFA has stopped working, confirm that your server clock is not out of sync. On Unix systems you can view the hardware time of your system using ``# hwclock -r``. If you're using systemd you can also use ``# timedatectl``.
```

justinegeffen:
hey @svelle, following up on this, what would be the next steps to finalize this PR? Should we add remediation steps or is this information sufficient?

jasonblais:
@svelle @justinegeffen friendly reminder for reviews :)

justinegeffen:
Thanks, @jasonblais! I was going to chat to @svelle about it in our sync later but what do you think about this being a forum post? It seems like there are various options around resolving this and it might be helpful to the community to have a more dynamic solution rather than a static FAQ entry?

jasonblais:
@justinegeffen I'd treat the separation as 'official, formal resource' goes to docs, 'unofficial Q&A and resources' goes to forums

justinegeffen:
Thanks @jasonblais, that is a better way of looking at it.

I see this has been updated to include the instruction to use ``timedatectl``. @svelle would you recommend any further steps or is the actionable information that the user would see enough? And does it always resolve the issue?

PR #3730, opened 2020-07-01
Fix Kubernetes license instructions

icelander:
#### Summary

The existing instructions will not put the license secret in the Mattermost namespace, which will prevent the app server pod from starting

### DIFF
--- a/source/install/install-kubernetes-mattermost.rst
+++ b/source/install/install-kubernetes-mattermost.rst
@@ -84,7 +84,7 @@ First, create the Mattermost namespace with this command:
 
   $ kubectl create ns mattermost
 
-Then, if deploying Mattermost enterprise edition, apply the license file specifying the path to the file you created in step 1:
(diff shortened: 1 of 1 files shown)


source/install/install-kubernetes-mattermost.rst
```diff
@@ -82,11 +78,22 @@ Save the file as ``mattermost-installation.yaml``.
...
 **3. Apply the Installation Manifest File**
 
-To initiate deployment, apply the file, specifying the correct path, using:
+First, create the Mattermost namespace with this command:
 
 .. code-block:: sh
 
   $ kubectl create ns mattermost
+
+Then, if deploying Mattermost enterprise edition, apply the license file specifying the path to the file you created in step 1:
```
justinegeffen, line 87:
```suggestion
If you're deploying Mattermost Enterprise Edition, apply the license file by specifying the path to the file you created in step 1:
```

justinegeffen:
Thanks, @icelander, this is a great update. I've made one suggestion which included enterprise edition > Enterprise Edition, and committed it.

PR #3760, opened 2020-07-18
Enable Plugin Upload -> more specific where

JtheBAB:
#### Summary

As the setting is only available in the config.json, i think it is better to write it out. See also https://community-daily.mattermost.com/core/pl/qhcbmsg3njygjr396iujmr9yur

### DIFF
--- a/source/administration/plugins.rst
+++ b/source/administration/plugins.rst
@@ -272,7 +272,7 @@ Custom plugins
 ~~~~~~~~~~~~~~
 Installing a custom plugin introduces some risk. As a result, plugin uploads are disabled by default and cannot be enabled via the System Console or REST API.
 
-To enable plugin uploads, manually set **PluginSettings > EnableUploads** to ``true`` in your configuration and restart your server. You can disable plugin uploads at any time without affecting previously uploaded plugins.
+To enable plugin uploads, manually set **PluginSettings > EnableUploads** to ``true`` in your config.json file and restart your server. You can disable plugin uploads at any time without affecting previously uploaded plugins.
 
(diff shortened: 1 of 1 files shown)


source/administration/plugins.rst
```diff
@@ -272,7 +272,7 @@ Custom plugins
 ~~~~~~~~~~~~~~
 Installing a custom plugin introduces some risk. As a result, plugin uploads are disabled by default and cannot be enabled via the System Console or REST API.
 
-To enable plugin uploads, manually set **PluginSettings > EnableUploads** to ``true`` in your configuration and restart your server. You can disable plugin uploads at any time without affecting previously uploaded plugins.
+To enable plugin uploads, manually set **PluginSettings > EnableUploads** to ``true`` in your config.json file and restart your server. You can disable plugin uploads at any time without affecting previously uploaded plugins.
```
justinegeffen, line 275:
```suggestion
To enable plugin uploads, manually set **PluginSettings > EnableUploads** to ``true`` in your ``config.json`` file and restart your server. You can disable plugin uploads at any time without affecting previously uploaded plugins.
```

justinegeffen: I think these steps refer to the System Console as the config.json file doesn't include the fields **PluginSettings > Enable Uploads**. The config.json reference [here](https://docs.mattermost.com/administration/config-settings.html#enable-plugins) differs slightly. 

That said, it looks like the instructions for the System Console need to be updated as there is no **Enable Plugins** option in **Plugin Management**. 

@aaronrothschild, can you confirm that the steps should be **System Console > Plugins (Beta) > Plugin Management > Enable Plugins** to **true**?

JtheBAB: There is the field in the config.json. See my screenshot when you go to the permlink.

justinegeffen:
Reviewed with @JtheBAB and my feedback was incorrect in this context. Thanks for the update, @JtheBAB!

PR #3771, opened 2020-07-21

amyblais:
For https://github.com/mattermost/enterprise/pull/638, https://github.com/mattermost/mattermost-server/pull/14640, and PostgreSQL deprecation.


source/administration/important-upgrade-notes.rst
```diff
@@ -13,6 +13,20 @@ Important Upgrade Notes
 +----------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | If you’re upgrading from a version earlier than... | Then...                                                                                                                                                          |
 +====================================================+==================================================================================================================================================================+
+| v5.26.0                                            | Admins should re-index Elasticsearh using the ``Index now`` button so that all the changes are included in the index and all the entities start using it.        |
```
ethervoid, line 16: The should use the `Purge index` and then the `Index now` because we need to recreate the indexes. Also tell them that this would leave the system with a limited search so should be scheduled when there is little or no activity because it could take hours (In our installation it takes 1 day or so)

PR #3782, opened 2020-07-27
Remove Team Edition reference from upgrade guide

jasonblais:
Remove Team Edition reference from upgrade guide, including the "upgrade to MM v3.0" referenced in the upgrade guide.

Note that Team Edition is not referenced in any install guides, so including it may cause confusion among admins. As a data point, we see a significant number of admins first installing E0, then upgrading to a new version with TE binary for no apparent reason. The hypothesis is that the Enterprise Edition referenced in the upgrade guide is confused with the Enterprise subscription.

For detailed analysis and data, see https://docs.google.com/document/d/1iyNN1ytJ7nKetDjZCRUPG2yqEBgCKVbGjSaB64bvpf0/edit#

Previously approved by @lfbrock at https://community.mattermost.com/private-core/pl/ebtxzz7b3bfzzpszh57ojah85h

EDIT: We will do an analysis on whether this change has a positive impact to the admin user experience.

### DIFF
--- a/source/administration/upgrade.rst
+++ b/source/administration/upgrade.rst
@@ -46,17 +46,9 @@ Location of your local storage directory
 
 #. Download `the latest version of Mattermost Server <https://about.mattermost.com/download/>`__. In the following command, replace ``X.X.X`` with the version that you want to download:
 
-   *Enterprise Edition*
-
    .. code-block:: sh
 
      wget https://releases.mattermost.com/X.X.X/mattermost-X.X.X-linux-amd64.tar.gz
-
-   *Team Edition*
-
-   .. code-block:: sh
-
-    wget https://releases.mattermost.com/X.X.X/mattermost-team-X.X.X-linux-amd64.tar.gz
 
 #. Extract the Mattermost Server files.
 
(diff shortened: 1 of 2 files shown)


justinegeffen:
Thanks, @jasonblais, this works for me. Wondering if we should also edit:
- https://docs.mattermost.com/administration/legacy-upgrade.html#upgrade-enterprise-edition-to-3-0-3 (I spotted a typo that needs to be fixed there and we can remove reference to TE)
- https://docs.mattermost.com/administration/upgrade.html#upgrading-team-edition-to-enterprise-edition (minor edit to the sentence which references TE)
- https://docs.mattermost.com/administration/upgrading-to-2.0.html

jasonblais:
@justinegeffen 

1 - I considered it, but it doesn't have many pageviews and is not referenced in the main upgrade guide, so feels low impact.
2 - We should keep, as we want those in Team Edition to have clear instructions to upgrade to Enterprise Edition
3 - Enterprise Edition was released in 2.1, so upgrading to 2.0 can only be done for TE

justinegeffen:
Makes sense! Thanks. :)

PR #3795, opened 2020-07-31
Update guest-accounts.rst

wiersgallak:


wiersgallak:
Used for a quick training demo.

Issue #3798, opened 2020-08-03
Deprecate Ubuntu 16.04


svelle:
Since Ubuntu 16.04 is ending support soon we should move this guide to a deprecated spot and make it clear that mattermost should be setup with newer versions of ubuntu instead.

PR #3801, opened 2020-08-03
add deprecation warning to ubuntu 16.04 docs

svelle:
First step for #3798

### DIFF
--- a/source/install/install-ubuntu-1604-server.rst
+++ b/source/install/install-ubuntu-1604-server.rst
@@ -3,6 +3,10 @@
 Installing Ubuntu Server 16.04 LTS
 ==================================
 
+.. warning::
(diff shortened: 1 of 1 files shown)


source/install/install-ubuntu-1604-server.rst
```diff
@@ -3,6 +3,10 @@
 Installing Ubuntu Server 16.04 LTS
 ==================================
 
+.. warning::
+  Ubuntu 16.04 is approaching its End of Life in April 2021.
+  We don't recommend installing new instances of Mattermost on Ubuntu 16.04 and instead advise you to go with a newer release like Ubuntu 20.04.
```
justinegeffen, line 8:
```suggestion
  Because of this, we don't recommend installing new instances of Mattermost on Ubuntu 16.04. Rather use a newer version such as Ubuntu 20.04.
```

PR #3805, opened 2020-08-05
Add an FAQ for how to install Mattermost with the open source Team Edition licensed under MIT

jasonblais:
Currently no documentation in place, and the removal of the TE link on the download page makes it harder to discover.

This adds an FAQ entry, with the intent that it's web discoverable and easily referenceable. Also added a brief description of the Team Edition link in the version archive.

### DIFF
--- a/source/overview/faq.rst
+++ b/source/overview/faq.rst
@@ -31,6 +31,13 @@ We have done performance testing of 60,000 concurrent users and 60 million posts
 
 Mattermost provides an open source, well-documented `load test simulator <https://github.com/mattermost/mattermost-load-test>`_ to verify that your Mattermost deployment can achieve the stated scale benchmarks ahead of production deployment.
 
+How do I deploy the open source Mattermost Team Edition under an MIT license?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The open source Mattermost Team Edition is functionally identical to the commercial Mattermost Enterprise Edition in its free “team mode”, but there is no ability to unlock enterprise features. It deploys as single Linux binary with MySQL or PostgreSQL under an MIT license.
+
+To deploy the Team Edition, download the `Mattermost Team Edition binary <https://docs.mattermost.com/administration/version-archive.html#mattermost-team-edition-server-archive>`_, and follow our standard install guides. The same applies to server upgrades.
+
 Community Questions
 -------------------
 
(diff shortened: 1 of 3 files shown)


source/overview/faq.rst
```diff
@@ -31,6 +31,13 @@ We have done performance testing of 60,000 concurrent users and 60 million posts
 
 Mattermost provides an open source, well-documented `load test simulator <https://github.com/mattermost/mattermost-load-test>`_ to verify that your Mattermost deployment can achieve the stated scale benchmarks ahead of production deployment.
 
+How do I deploy the open source Mattermost Team Edition under an MIT license?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The open source Mattermost Team Edition is functionally identical to the commercial Mattermost Enterprise Edition in its free “team mode”, but there is no ability to unlock enterprise features. It deploys as single Linux binary with MySQL or PostgreSQL under an MIT license.
```
lfbrock, line 37: Can we add a sentence about what we recommend, similar to how GitLab does?

E.g. "We generally recommend installing Enterprise Edition, even if you're not sure if you will ever install a license. However, if you only want to install software with a fully open source code base, then Team Edition is the best choice for you." 

Not sure if "fully open source code base" sounds exactly right, but I think we should have something along the lines of the example so we're giving a clear recommendation on when to use each.

source/overview/product.rst
```diff
@@ -40,7 +40,7 @@ Features include:
 - New features and improvements released every two months
 - Multiple languages including U.S. English, Chinese (Simplified and Traditional), Dutch, French, German, Italian, Japanese, Korean, Polish, Brazilian Portuguese, Romanian, Russian, Turkish, Spanish, and Ukrainian
 
-To get started, `download the open source Mattermost Team Edition server <https://about.mattermost.com/download>`__ under an MIT license.
+To get started, `download the open source Mattermost Team Edition server <https://docs.mattermost.com/administration/version-archive.html#mattermost-team-edition-server-archive>`__ under an MIT license.
```
lfbrock, line 43: Can't comment on the earlier part of this doc -- but where it says "the open source edition is functionally equivalent...", I would recommend making a clear recommendation here as well on when to use TE and when to use EE.

source/overview/faq.rst
justinegeffen, line 37:
```suggestion
The open source Mattermost Team Edition is functionally identical to the commercial Mattermost Enterprise Edition in its free “team mode”, but there is no ability to unlock enterprise features. It deploys as single Linux binary with MySQL or PostgreSQL under an MIT license.

We generally recommend installing Enterprise Edition, even if you don't currently need a license. However, if you only want to install software with a fully open source code base, then Team Edition is the best choice for you.
```

justinegeffen: 2/5 on my phrasing but 3/5 on adding it.

jasonblais:
Thanks for the feedback! Made a few updates.

PR #3809, opened 2020-08-05
Update changelog.md

amyblais:
For v5.25.3 dot release


hahmadia:
@amyblais This should be cherrypicked into 5.25 and 5.26, right?

amyblais:
@hahmadia That's correct.

PR #3813, opened 2020-08-10
[ MM-27369] Update code examples

larkox:
#### Summary
Update code examples so when updating the post, we delete the attachments.
We may also want to make people more aware of the section `How do I manage properties of an interactive message?`.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-27369


justinegeffen:
Thanks, @larkox, great update. :)

PR #3839, opened 2020-08-13
Update team-channel-management.rst

amyblais:
@thefactremains You mentioned that team management didn't make it to v5.26? Then it has to be removed from the docs. Please review which sections need to be removed.


amyblais:
@thefactremains You mentioned that team management didn't make it to v5.26? Then it has to be removed from the docs. Please review which sections need to be removed.

amyblais:
@thefactremains Is there anything else that needs to be updated / removed?

thefactremains:
> @thefactremains Is there anything else that needs to be updated / removed?

No, that covers it. thank you @amyblais

PR #3846, opened 2020-08-15

justinegeffen:
Minor formatting and terminology fixes.


justinegeffen:
> Edits are fine but the table is now broken

Yeah I saw that and I can't figure out why. Let me get back to this and see what I broke.

amyblais:
You can change the Line Wrap Mode to "No Wrap" when editing the file and then you should be able to see which parts of the table formatting are not aligned.

justinegeffen:
> You can change the Line Wrap Mode to "No Wrap" when editing the file and then you should be able to see which parts of the table formatting are not aligned.

Ah - so I did that when I was editing but didn't double-check and I'd bumped out some of the formatting. Looks fixed now, thanks!

PR #3868, opened 2020-08-25
Update desktop-changelog.rst

amyblais:
For v4.5.3 dot release.

### DIFF
--- a/source/help/apps/desktop-changelog.rst
+++ b/source/help/apps/desktop-changelog.rst
@@ -4,9 +4,9 @@ Desktop Application Changelog
 Release v4.5
 ----------------------------
 
-**Download Binaries:** `Mattermost Desktop on GitHub <https://github.com/mattermost/desktop/releases/tag/v4.5.2>`_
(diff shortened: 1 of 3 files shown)


source/help/apps/desktop-changelog.rst
```diff
@@ -4,8 +4,10 @@ Desktop Application Changelog
 Release v4.5
 ----------------------------
 
-**Download Binaries:** `Mattermost Desktop on GitHub <https://github.com/mattermost/desktop/releases/tag/v4.5.2>`_
+**Download Binaries:** `Mattermost Desktop on GitHub <https://github.com/mattermost/desktop/releases/tag/v4.5.3>`_
 
+- **v4.5.3, released 2020-08-25**
+ - Fixed an issue where users were unable to log in to the desktop app when users had to select a certificate for authentication even when there was only one option to manage a certificate login. `MM-27331 <https://mattermost.atlassian.net/browse/MM-27331>`_
```
Willyfrog, line 10: This only seems to happen with physical certificates which requires a pin to make it work. Do you think we should clarify that?

Willyfrog (approved):
Looks good, added a comment but feel free to disregard if you don't think we should be too specific

PR #3917, opened 2020-09-16
Provide id for post actions

hanzei:
#### Summary
Add an `id` to each Post Action

#### Ticket Link
https://github.com/mattermost/mattermost-server/pull/13225

### DIFF
--- a/source/developer/interactive-messages.rst
+++ b/source/developer/interactive-messages.rst
@@ -40,6 +40,7 @@ The following payload gives an example that uses message buttons.
         "text": "This is the attachment text.",
         "actions": [
           {
+            "id": "message",
             "name": "Ephemeral Message",
             "integration": {
               "url": "http://127.0.0.1:7357",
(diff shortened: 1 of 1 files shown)


source/developer/interactive-messages.rst
```diff
@@ -200,6 +205,9 @@ Parameters
 
 Below is a brief description of each parameter to help you customize the interactive message button and menu in Mattermost. For more information on message attachments, `see our documentation <https://docs.mattermost.com/developer/message-attachments.html>`__.
 
+ID
+  A per post unique identifier.
```
justinegeffen, line 209:
```suggestion
  A unique identifier required for each post.
```

justinegeffen (approved):

justinegeffen:
Just one suggestion, not a blocker. Thanks a lot for putting this together, really appreciate it. :)

source/developer/interactive-messages.rst
hanzei: @justinegeffen That is actually not what I meant. The ID is unique per post. It's not required for every post action. How can I make this more clear?

justinegeffen: I think we can go with what you've got. No need to over-complicate it - the wording does make sense.

levb (approved):

PR #3924, opened 2020-09-17
Update mobile-appconfig.rst to include new AppConfig `timeout` property

sadohert:


enahum (commented):
I remember saying that we should not document this, unless it is extremely clear that you should **NOT** use it as it may be hiding underlying issues in your deployment.

justinegeffen (approved):

justinegeffen:
LGTM, thanks @sadohert!

sadohert:
@enahum - My Bad... The customer was looking for docs from me and I dove in here.  

Do we you want me to pull it back?  Woudl adding heavy caveats be okay?  "Not recommended for most deployments.  This parameter helps to address issues with infrastructure performance"  or something like that?

enahum:
@sadohert I want us to discourage people from using it.. not sure about the words to use, but 5/5 on the intention.

justinegeffen:
@sadohert, @enahum: should this PR be closed?

enahum:
I'd say we should close it

sadohert:
I understand we really don't want it to be used, but my concern would be that the option gets lost.  There's a customer that needs it now, and there could be others in the future.  If this isn't surfaced in any way it would be an unfortunate lost opportunity to help another customer that might need it.

I would like to hold onto it with strong caveats against it.

enahum:
@sadohert maybe add a WARNING saying that this is something that we discourage as it could be hiding underlaying issues blah blah blah, this is something that the current customer that is using it should really look into and not relay on this.

justinegeffen:
I agree with a warning and retaining the info. This could also work well as a forum post, if we don't specifically want to commit to/support this and if we'd like feedback from the community in terms of usage and improvement?

sadohert:
+1 on putting this into a forum post instead.

justinegeffen:
@svelle, would this be beneficial to Support in a different format/place than the docs?

cwarnermm:
I recommend that this ticket for documentation updates be converted to a ticket to create a new article in the Support KB. @svelle - If you agree, can a member of the Support team take this one on, please?

svelle:
thanks @cwarnermm good point about this not needing to live in the docs.

@sadohert seeing as this is a relatively old PR do you see any use in us adding this to the KB?

sadohert:
@svelle  - I think its worth adding to the KB.  I would really prefer there was SOME artifact that can be discovered to help someone in teh future.

PR #3925, opened 2020-09-17

amyblais:
For https://github.com/mattermost/mattermost-server/pull/15274


source/administration/important-upgrade-notes.rst
```diff
@@ -13,6 +13,8 @@ Important Upgrade Notes
 +----------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | If you’re upgrading from a version earlier than... | Then...                                                                                                                                                          |
 +====================================================+==================================================================================================================================================================+
+| v5.28.0                                            | The service will crash with a coredump instead of just dumping the stack trace to console.                                                                       |
```
amyblais, line 16: Unsure if more details, perhaps why this change was made or who is impacted.

agnivade: I realize the wording suggests that the service will just crash, which makes it sound odd. Here's my new suggestion:

Now when the service crashes, it will generate a coredump instead of just dumping the stack trace to the console. This allows us to preserve the full information of the crash to help in debugging later.

@streamer45 - do you think it's necessary to talk about `/proc/sys/kernel/core_pattern` here? I think all Linux systems have some value by default. And if sysadmins have deliberately set them to not generate core dump, they already know what they are doing.

streamer45: I wouldn't go deep into it but I think a mention with a link to https://man7.org/linux/man-pages/man5/core.5.html can't hurt. Mainly to explain that in case of a crash, the core dump output/naming can be controlled by that kernel parameter.

agnivade: Ok, then final wording:

Now when the service crashes, it will generate a coredump instead of just dumping the stack trace to the console. This allows us to preserve the full information of the crash to help in debugging later.

For more information about coredumps, please see: https://man7.org/linux/man-pages/man5/core.5.html

agnivade (approved):

PR #3928, opened 2020-09-22
Update deprecation note for Postgres v9.4

amyblais:
Will also do a Forum post and add this to the changelogs.


wiersgallak (approved):
Thanks!

PR #3929, opened 2020-09-22
Update notes on supported versions

amyblais:
v5.19 ESR support is ending. Also planning a forum post and blog post.
Also mobile app changelog is updated.


esethna:
@amyblais please help also draft another email push to customers based on the last one we wrote and add me for review. We should send to an updated email list based on the most current server version data (ie send to all admins <v5.25 minus fed)

esethna (approved):

PR #3981, opened 2020-10-12
Update bulk-export.rst

svelle:
Just adding an additional note for visibility since it really isn't *that* clear if you're unfamiliar with the product.

### DIFF
--- a/source/administration/bulk-export-data.rst
+++ b/source/administration/bulk-export-data.rst
@@ -19,7 +19,7 @@ You can export the following data types:
 - Direct Message Posts
 
 .. note:: 
-  Configuration for exporting specific areas of the server, exporting additional types of posts, permissions schemes, file attachments, webhooks, and bot messages are not yet supported. Deleted objects are also not yet supported.
+  Configuration for data types such as exporting specific areas of the server, exporting additional types of posts, permissions schemes, file attachments, webhooks, and bot messages is not yet supported. Deleted objects are also not yet supported.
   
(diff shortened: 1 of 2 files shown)


svelle:
removed initial note and instead made the already existing note in the bulk-export-data.rst file more visible by marking it as a note.

source/administration/bulk-export-data.rst
```diff
@@ -14,15 +15,17 @@ You can export the following data types:
...
 - Custom Emoji
-- Direct Message Channels 
+- Direct Message Channels
 - Direct Message Posts
 
-Configuration for exporting specific areas of the server, exporting additional types of posts, permissions schemes, file attachments, webhooks and bot messages are not yet supported. Deleted objects are also not yet supported.  
-
-For requests to add additional attributes or objects to our exporter, please add a feature request on our `feature idea forum <https://mattermost.uservoice.com/forums/306457-general>`__.  
+.. note:: 
+  Configuration for exporting specific areas of the server, exporting additional types of posts, permissions schemes, file attachments, webhooks, and bot messages are not yet supported. Deleted objects are also not yet supported.
```
justinegeffen, line 22:
```suggestion
  Configuration for data types such as exporting specific areas of the server, exporting additional types of posts, permissions schemes, file attachments, webhooks, and bot messages is not yet supported. Deleted objects are also not yet supported.
```

justinegeffen: @svelle, did I get the tense right in my suggestion?

svelle: I think so, this part wasn't originally written by me. But from my understanding this is what it's supposed to say.

justinegeffen (approved):

PR #3994, opened 2020-10-15
Update postgres instructions to add trust for local ipv4 connections

svelle:
Previously we only had the postgres trust connection change made for local socket connections. 
However on almost all of the systems postgres may also very well connect through ipv4 when  for example using localhost as the domain. This adds an additional line of change to the `pg_hba.conf` to make that connection possible. 
From a security standpoint not much should change since it's still only for local connections.

### DIFF
--- a/source/install/install-ubuntu-2004-postgresql.rst
+++ b/source/install/install-ubuntu-2004-postgresql.rst
@@ -65,7 +65,7 @@ Assume that the IP address of this server is 10.10.10.1.
 
   **If the Mattermost server and the database are on the same machine**:
 
-    a. Open ``/var/lib/pgsql/10/main/pg_hba.conf`` as root in a text editor.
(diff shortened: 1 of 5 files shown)


source/install/install-ubuntu-2004-postgresql.rst
```diff
@@ -65,15 +65,19 @@ Assume that the IP address of this server is 10.10.10.1.
 
   **If the Mattermost server and the database are on the same machine**:
 
-    a. Open ``/etc/postgresql/10/main/pg_hba.conf`` in a text editor as *root* user.
+    a. Open ``/etc/postgresql/10/main/pg_hba.conf`` as root in a text editor.
```
justinegeffen, line 68:
```suggestion
    a. Open ``/etc/postgresql/10/main/pg_hba.conf`` as *root* in a text editor.
```

justinegeffen (approved):

PR #3999, opened 2020-10-15
Update notices.rst

amyblais:
Propose removing the image since we haven't been able to figure out why it doesn't work.


esethna:
@amyblais I fixed the image ^

esethna (approved):

esethna:
@amyblais @justinegeffen `../` syntax is referring to going back one file directory, that's why the initial syntax didn't work

PR #4011, opened 2020-10-19
docs: Updated elasticsearch documentation

yash2189:
#### Summary

This PR adds a note section under `source/deployment/elasticsearch.rst`  which includes a limitation about the ElasticSearch progress bar regarding its inaccuracy for large post tables. The required changes are added in a notes section so that users can know about the same.

Please find the updated docs [here](https://github.com/yash2189/docs/blob/fix%234000/source/deployment/elasticsearch.rst)

### DIFF
--- a/source/deployment/elasticsearch.rst
+++ b/source/deployment/elasticsearch.rst
@@ -26,6 +26,10 @@ Elasticsearch v5.x, v6.x, and v7.x are supported.
 .. note::
     From Mattermost v5.26, you can filter inactive users, search by user role, and also search for terms inside links. This update introduces a breaking change which affects the "from" part of the search. To avoid this, reindex your Elasticsearch instance/cluster prior to upgrading.
 
+
+.. note::
+    If your installation has a large number of posts (typically in excess of 1 million but not strictly defined) the reindexing progress percentage may stay at 99% for a long time.
+    
 Setting Up an Elasticsearch Server
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


source/deployment/elasticsearch.rst
```diff
@@ -26,6 +26,10 @@ Elasticsearch v5.x, v6.x, and v7.x are supported.
 .. note::
     From Mattermost v5.26, you can filter inactive users, search by user role, and also search for terms inside links. This update introduces a breaking change which affects the "from" part of the search. To avoid this, reindex your Elasticsearch instance/cluster prior to upgrading.
 
+
+.. note::
+    If your installation has a large number of posts (typically in excess of 1 million but not strictly defined) the reindexing progress percentage may stay at 99% for a long time.
```
justinegeffen, line 31:
```suggestion
    If your deployment has a large number of posts (typically in excess of 1 million but not strictly defined) the reindexing progress percentage may stay at 99% for a long time.
```

justinegeffen: One suggestion for this is to place it after the set up steps, so that there's more context (based on the assumption that the reader has set Elasticsearch up and the problem may affect them). Let me know what you think. :)

yash2189: Well I agree to you , yes I was actually thinking about placing it after the setup steps but wasn't sure if that was correct.

yash2189:
@justinegeffen for signing the CLA do I just need to comment `/check CLA`?

source/deployment/elasticsearch.rst
justinegeffen: I totally understand! I think it fits well in either place but I'm more inclined for it to be after the set up steps. :)

yash2189:
I have made the changes @justinegeffen

justinegeffen (approved):

justinegeffen:
Looks good to me! I've added a PM for additional visibility. :) Thank you, @yash2189!

wiersgallak (approved):

Issue #3943, opened 2020-09-28
LATIN1 DB encoding error


DavidePrincipi:
I want to insert an € (euro sign) in a chat message text, but I get an error:

![immagine](https://user-images.githubusercontent.com/2920838/94421627-21d84880-0186-11eb-8790-17a03544e5ad.png)

In `mattermost.log`

```json
{"level":"error","ts":1601288852.6284842,"caller":"mlog/log.go:190","msg":"Unable to save the Post.","path":"/api/v4/posts","request_id":"579m3tr71pn9dq6hd5uihw46nr","ip_addr":"127.0.0.1","user_id":"mc4xjqo6rpgdbfbr5o3daumbow","method":"POST","err_where":"SqlPostStore.Save","http_code":500,"err_details":"pq: character with byte sequence 0xe2 0x82 0xac in encoding \"UTF8\" has no equivalent in encoding \"LATIN1\""}
```

What I did is just running this (weird) command in a CentOS 7 system:

    localectl set-locale LANG=en_US.iso88591

...then installed PostgreSQL 12 and run `CREATE DATABASE mattermost;` (full story here https://github.com/NethServer/dev/issues/6286).

Actual database encoding information is:

```
[root@vm5 ~]# su - postgres  -c "scl enable rh-postgresql12 -- psql --port=55434"
psql (12.1)
Type "help" for help.

postgres=# \l
                                     List of databases
    Name    |  Owner   | Encoding |    Collate     |     Ctype      |   Access privileges   
------------+----------+----------+----------------+----------------+-----------------------
 mattermost | postgres | LATIN1   | en_US.iso88591 | en_US.iso88591 | =Tc/postgres         +
            |          |          |                |                | postgres=CTc/postgres+
            |          |          |                |                | mattuser=CTc/postgres
 postgres   | postgres | LATIN1   | en_US.iso88591 | en_US.iso88591 | 
 template0  | postgres | LATIN1   | en_US.iso88591 | en_US.iso88591 | =c/postgres          +
            |          |          |                |                | postgres=CTc/postgres
 template1  | postgres | LATIN1   | en_US.iso88591 | en_US.iso88591 | =c/postgres          +
            |          |          |                |                | postgres=CTc/postgres
(4 rows)
```

What I expect is to get an error somewhere (as the system encoding is not fully functional) and to see a big warning in the docs: https://docs.mattermost.com/install/install-rhel-7.html#installing-postgresql-database

justinegeffen:
[...]

PR #4029, opened 2020-10-23
Ensure DB encoding is UTF-8

DavidePrincipi:
#### Summary

The default PostgreSQL DB encoding depends on the shell locale set at initialization time. 
As Mattermost might not work with other encodings, create the DB by setting the UTF-8 encoding explicitly.

#### Ticket Link

More details here:

- https://github.com/mattermost/docs/issues/3943
- https://github.com/NethServer/nethserver-mattermost/pull/56

### DIFF
--- a/source/install/install-rhel-7-postgresql.rst
+++ b/source/install/install-rhel-7-postgresql.rst
@@ -44,9 +44,9 @@ Installing PostgreSQL Database
 
   ``psql``
 
-10.  Create the Mattermost database.
+10.  Create the Mattermost database with UTF-8 encoding.
 
-  ``postgres=# CREATE DATABASE mattermost;``
+  ``postgres=# CREATE DATABASE mattermost WITH ENCODING 'UTF8' LC_COLLATE='en_US.UTF-8' LC_CTYPE='en_US.UTF-8' TEMPLATE=template0;``
 
 11.  Create the Mattermost user 'mmuser'.
 
--- a/source/install/install-rhel-7-postgresql.rst
+++ b/source/install/install-rhel-7-postgresql.rst
@@ -44,7 +44,7 @@ Installing PostgreSQL Database
 
   ``psql``
 
-10.  Create the Mattermost database with UTF-8 encoding.
+10.  Create the Mattermost database.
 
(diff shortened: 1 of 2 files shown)


justinegeffen:
Thank you for this, @DavidePrincipi! To confirm, should this also be updated for the RHEL 8 docs? Or is it just for RHEL 7?

DavidePrincipi:
...I guess the same applies to any PostgreSQL-based documentation, however I can't verify it. Do I have to fix them all anyway?

justinegeffen:
Thanks, @DavidePrincipi, that's great. I'll defer to @agnivade and @streamer45 in terms of verifying it. And then we'll have a clearer idea of the extent of the update. For now, though, I don't want to add unnecessary work so I'll await their feedback.

source/install/install-rhel-7-postgresql.rst
```diff
@@ -44,9 +44,9 @@ Installing PostgreSQL Database
 
   ``psql``
 
-10.  Create the Mattermost database.
+10.  Create the Mattermost database with UTF-8 encoding.
```
agnivade, line 47: We don't need to mention the UTF-8 part here. The command is all the user needs to execute.

source/install/install-rhel-8-postgresql.rst
```diff
@@ -30,9 +30,9 @@ Installing PostgreSQL Database
 
   ``psql``
 
-8.  Create the Mattermost database.
+8.  Create the Mattermost database with UTF-8 encoding.
```
agnivade, line 33: Same as above.

agnivade (changes requested):
Thank you @DavidePrincipi for the change! Looks good to me, just one small thing to take care of.

justinegeffen:
Looks good to me editorially; agree with the suggestions made by @agnivade.

agnivade (approved):

streamer45 (approved):
Grazie @DavidePrincipi :)

@justinegeffen , @agnivade If you agree I think it makes sense to open a ticket to also update other postgres specific installation guides since this could be helpful on other distributions/systems as well.

agnivade:
Makes sense to me.

justinegeffen (approved):

justinegeffen:
Thanks, @streamer45! I do agree with that. I'll merge this PR and then create a HW ticket for the further update. :)

PR #4096, opened 2020-11-09
Update telemetry.rst

amyblais:
- Added ``enum ManagedResourcePaths``; for https://github.com/mattermost/mattermost-server/pull/16221


hmhealey (changes requested):
I think this should actually be added under the True/false section below since we only track whether or not a system is using the setting

amyblais:
Ok, I added it to the True/False section under ServiceSettings as ``bool ManagedResourcePaths``.

hmhealey (approved):

PR #4121, opened 2020-11-18
Update requirements.rst

wiersgallak:
Added note about Admin being a reserved keyword in MySQL versions 8.0.0-8.0.11. Related ticket: https://mattermost.atlassian.net/browse/MM-30001

### DIFF
--- a/source/install/requirements.rst
+++ b/source/install/requirements.rst
@@ -91,7 +91,7 @@ While community support exists for Fedora, FreeBSD, and Arch Linux, Mattermost d
 Database Software
 ^^^^^^^^^^^^^^^^^
 
--  MySQL 5.6, 5.7, 8 (see note below on MySQL 8 support)
+-  MySQL 5.6, 5.7, 8.0.12 (see note below on MySQL 8 support)
 -  PostgreSQL 10.0+
 -  Amazon Aurora MySQL 5.6+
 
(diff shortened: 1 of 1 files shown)


source/install/requirements.rst
```diff
@@ -116,6 +116,8 @@ In MySQL 8.0.4, the default authentication plugin was changed from ``mysql_nativ
    
    [mysqld]
    default-authentication-plugin=mysql_native_password
+   
+In MySQL versions 8.0.0-8.0.11 ```ADMIN``` is a `reserved keyword <https://dev.mysql.com/doc/refman/8.0/en/keywords.html>`_. We recommend installing 8.0.12 or higher since it is no longer reserved in these versions.
```
agarciamontoro, line 120: I think we should not "recommend", but require it. If someone's using v8, the minimum supported version is 8.0.12. Otherwise, they will run into problems. Also, maybe people don't really care about the details of the admin word being a reserved keyword (it's just the implementation detail that makes Mattermost incompatible with 8.0.0-8.0.11)

agarciamontoro (changes requested):
Thank you, @wiersgallak!

source/install/requirements.rst
wiersgallak: @agarciamontoro I've removed the language around "recommendations" and clarified the requirement around 8.0.4 on updating the plugin portion.  I have chosen to keep the information that explains why we don't support v 8.0.0-8.0.11 so customers have more context around it.

cwarnermm (approved):
Requirement is clearly documented and includes an explanation of why the requirement exists.

source/install/requirements.rst
```diff
@@ -91,7 +91,7 @@ While community support exists for Fedora, FreeBSD, and Arch Linux, Mattermost d
 Database Software
 ^^^^^^^^^^^^^^^^^
 
--  MySQL 5.6, 5.7, 8 (see note below on MySQL 8 support)
+-  MySQL 5.6, 5.7, 8.0.12 (see note below on MySQL 8 support)
```
grundleborg, line 94: Should this say `v8.0.12+` ?

agarciamontoro (approved):
Thank you, @wiersgallak :tada:

PR #4125, opened 2020-11-20
Added static robots.txt file, corrected https reference on 404 page

lawrencejohnson:
#### Summary
This page includes the addition of a static robots.txt file provided by the SEO team and an absolute URL correction to the 404 page.

#### Ticket Link
https://mattermost.atlassian.net/browse/MAR-162

### DIFF
--- a/source/404.html
+++ b/source/404.html
@@ -22,7 +22,7 @@
 </head>
 <body>
   <header>
-    <a href="http://docs.mattermost.com/" class="header__logo">
+    <a href="https://docs.mattermost.com/" class="header__logo">
       <img width="176" src="https://about.mattermost.com/wp-content/uploads/2017/03/headerlogo.svg" alt="Mattermost Logo">
     </a>
   </header>
--- a/source/_static/robots.txt
+++ b/source/_static/robots.txt
@@ -1,0 +1,8 @@
+User-Agent: *
+Disallow: /cgi-bin/
+Disallow: /&utm_source
+Disallow: /&utm_medium
+Disallow: /&utm_campaign
+Disallow: /?utm_
+
+Sitemap: https://docs.mattermost.com/sitemap.xml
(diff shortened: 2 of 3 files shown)


justinegeffen:
Thanks for this PR, @lawrencejohnson! A question about the 404 page change; does this mean that users will now hit a 404 page with an error instead of the current dead page? We have an open Jira issue for a custom 404 page [here](https://mattermost.atlassian.net/browse/MM-29614) but if this has the same/a similar result then I don't see any issues with using it. 

Are you open to elaborating on the other changes? And is there anything we'd need to validate/check on our side with regard to building docs?

lawrencejohnson:
> Thanks for this PR, @lawrencejohnson! A question about the 404 page change; does this mean that users will now hit a 404 page with an error instead of the current dead page? We have an open Jira issue for a custom 404 page [here](https://mattermost.atlassian.net/browse/MM-29614) but if this has the same/a similar result then I don't see any issues with using it.
> 
> Are you open to elaborating on the other changes? And is there anything we'd need to validate/check on our side with regard to building docs?

Hi Justine, the 404 update won't correct the current 404 issue although I think the transition to read the docs will correct that. I just happened to notice that the hard link in that 404 file was mapped to a non-https URL.

The addition of a static file called `robots.txt` is a file that will live in the root of the website. It's used by search engines to help crawlers determine what links are good or bad. The SEO team that is currently assigned to improving SEO provided that file as is.

The configuration file has a small change which simply tells the compiler to include the new robots.txt in the output.

None of the changes will have an affect on any existing pages. All there is to do in terms of testing is to make sure the project builds as it normally does and verify that there is also a robots.txt in the foot of the site.

wiersgallak (approved):

jasonblais (approved):

justinegeffen (approved):

justinegeffen:
Thanks, @lawrencejohnson! I appreciate all the feedback and the changes. :)

PR #4132, opened 2020-11-25
Remove `ExperimentalDataPrefetch` from system console

sudheerDev:
#### Summary
Removing the section about data prefetch from docs

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27989


justinegeffen (approved):

esethna:
@amyblais @justinegeffen how do we normally deal with config removes in the docs?

amyblais:
@esethna This PR looks good to me. I can then later add a note to changelog, upgrade notes, and [deprecation webpage](https://about.mattermost.com/deprecated-features/) when we know which release this is going to.

esethna (approved):

justinegeffen:
Thanks for raising that, @esethna. I approved the deletion but should also have confirmed with Amy whether it needs to remain in the doc with a deprecation note as opposed to being deleted. Thanks for confirming, @amyblais. Going forward, is this the right process or should a note be added instead of the content being removed entirely?

amyblais:
@justinegeffen Ah yes, that makes sense. I think adding a deprecation note instead of removing the content completely sounds better. We can work on it when we know which release this will happen in.

justinegeffen:
I've re-added the content and we can add a deprecation note similar to the others in this doc once we know which release it's going out in.

hmhealey:
Since it's missed 5.30, this'll be going out in 5.31 then. I can take over these PRs to finish them up

justinegeffen:
Changed the milestone and label for now so it doesn't go out with this release inadvertently.

PR #4137, opened 2020-11-29
Replace embedded scripts with google tag manager

lawrencejohnson:
#### Summary
As requested by SEO team, all 3rd-party javascript snippets that were previously embedded within the application have been transferred to a GTM container. The scripts were replaced with the GTM snippet and tested locally to ensure they were firing as intended.

Note: The `conf.py` setting `html_context.googleanalytics_enabled` was being used as a condition for only one of the three Analytics tracking snippets. A sitewide search yielded no results for this setting being modified within the app itself. It's possible that this setting can be removed, or it could be that the setting is adjusted during a deployment process. Recommending that a reviewer spot-check this. We may need to add conditional triggers in GTM to move this properly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MAR-93

### DIFF
--- a/source/404.html
+++ b/source/404.html
@@ -19,8 +19,19 @@
     }
 
   </style>
+  <!-- Google Tag Manager -->
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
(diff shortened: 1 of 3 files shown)


justinegeffen:
Thanks, @lawrencejohnson! I've added @asaadmahmood and @jaydeland as reviewers as they're more familiar with the build process. As far as I can see this looks fine.

justinegeffen (approved):

asaadmahmood (approved):
We can try out the `html_context.googleanalytics_enabled` when it goes live, and if it causes a problem, we can think about changing that.

jaydeland (approved):
I am not up on all these details but the changes look clean.

PR #4181, opened 2020-12-16
Add recommended postgres config

agnivade:
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-31392

### DIFF
--- a/source/deployment/cluster.rst
+++ b/source/deployment/cluster.rst
@@ -280,6 +280,31 @@ Transparent Failover
 
 The database can be configured for High Availability and transparent failover use the existing database technologies. We recommend MySQL Clustering, Postgres Clustering, or Amazon Aurora. Database transparent failover is beyond the scope of this documentation.
 
+Recommended Configuration Settings
+``````````````````````````````````
+
+If you are using Postgres as the choice of database, we recommend the following configuration optimizations which suits the workload of a Mattermost server.
+
+The following configuration was tested on an AWS Aurora r5.xlarge instance of Postgres 11.7.
+
+1. **max_connections**: If you are using read-replicas set reader connections to 1024, and writer connections to 256. If you are using a single instance, then only setting it to 1024 should be sufficient. If the instance of lower capacity than r5.xlarge, then set it to a lower number. Also tune the `MaxOpenConns` setting under the `SqlSettings` of Mattermost app accordingly.
+
+2. **random_page_cost**: Set it to 1.1, unless the DB is using spinning disks.
+
+3. **work_mem**: Set it to 16MB for readers, and 32MB for writers. If it's a single instance, then just 16MB should be sufficient. If the instance of lower capacity than r5.xlarge, then set it to a lower number.
+
+4. **effective_cache_size**: Set it to 65% of total memory. For a 32GB instance, it should be 21GB.
+
+5. **shared_buffers**: Set it to 65% of total memory. For a 32GB instance, it should be 21GB.
+
+6. **tcp_keepalives_count**: 5
+
+7. **tcp_keepalives_idle**: 5
+
+8. **tcp_keepalives_interval**: 1
+
+Note that if you are using pgbouncer or any similar connection pooling proxy in front of your DB, then the keepalive settings should be applied to the proxy instead and revert the keepalive settings for the DB back to defaults.
+
 Leader Election
 ^^^^^^^^^^^^^^^^
 
--- a/source/deployment/cluster.rst
+++ b/source/deployment/cluster.rst
@@ -295,7 +295,7 @@ The following configuration was tested on an AWS Aurora r5.xlarge instance of Po
 
 4. **effective_cache_size**: Set it to 65% of total memory. For a 32 GB instance, it should be 21 GB.
 
-5. **shared_buffers**: Set it to 65% of total memory. For a 32GB instance, it should be 21GB.
+5. **shared_buffers**: Set it to 65% of total memory. For a 32 GB instance, it should be 21 GB.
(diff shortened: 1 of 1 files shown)


source/deployment/cluster.rst
```diff
@@ -280,6 +280,31 @@ Transparent Failover
 
 The database can be configured for High Availability and transparent failover use the existing database technologies. We recommend MySQL Clustering, Postgres Clustering, or Amazon Aurora. Database transparent failover is beyond the scope of this documentation.
 
+Recommended Configuration Settings
+``````````````````````````````````
+
+If you are using Postgres as the choice of database, we recommend the following configuration optimizations which suits the workload of a Mattermost server.
```
justinegeffen, line 286:
```suggestion
If you're using Postgres as the choice of database, we recommend the following configuration optimizations on your Mattermost server.
```

source/deployment/cluster.rst
```diff
@@ -280,6 +280,31 @@ Transparent Failover
...
+
+The following configuration was tested on an AWS Aurora r5.xlarge instance of Postgres 11.7.
+
+1. **max_connections**: If you are using read-replicas set reader connections to 1024, and writer connections to 256. If you are using a single instance, then only setting it to 1024 should be sufficient. If the instance of lower capacity than r5.xlarge, then set it to a lower number. Also tune the `MaxOpenConns` setting under the `SqlSettings` of Mattermost app accordingly.
+
+2. **random_page_cost**: Set it to 1.1, unless the DB is using spinning disks.
+
+3. **work_mem**: Set it to 16MB for readers, and 32MB for writers. If it's a single instance, then just 16MB should be sufficient. If the instance of lower capacity than r5.xlarge, then set it to a lower number.
```
justinegeffen, line 294:
```suggestion
3. **work_mem**: Set it to 16 MB for readers, and 32 MB for writers. If it's a single instance, then just 16 MB should be sufficient. If the instance of lower capacity than r5.xlarge, then set it to a lower number.
```

source/deployment/cluster.rst
```diff
@@ -280,6 +280,31 @@ Transparent Failover
...
+
+4. **effective_cache_size**: Set it to 65% of total memory. For a 32GB instance, it should be 21GB.
```
justinegeffen, line 296:
```suggestion
4. **effective_cache_size**: Set it to 65% of total memory. For a 32 GB instance, it should be 21 GB.
```

source/deployment/cluster.rst
```diff
@@ -280,6 +280,31 @@ Transparent Failover
...
+
+5. **shared_buffers**: Set it to 65% of total memory. For a 32GB instance, it should be 21GB.
```
justinegeffen, line 298:
```suggestion
5. **shared_buffers**: Set it to 65% of total memory. For a 32 GB instance, it should be 21 GB.
```

justinegeffen (approved):

source/deployment/cluster.rst
```diff
@@ -280,6 +280,31 @@ Transparent Failover
...
+Recommended Configuration Settings
+``````````````````````````````````
+
+If you're using Postgres as the choice of database, we recommend the following configuration optimizations on your Mattermost server.
+
+The following configuration was tested on an AWS Aurora r5.xlarge instance of Postgres 11.7.
+
+1. **max_connections**: If you are using read-replicas set reader connections to 1024, and writer connections to 256. If you are using a single instance, then only setting it to 1024 should be sufficient. If the instance of lower capacity than r5.xlarge, then set it to a lower number. Also tune the `MaxOpenConns` setting under the `SqlSettings` of Mattermost app accordingly.
+
+2. **random_page_cost**: Set it to 1.1, unless the DB is using spinning disks.
```
streamer45, line 292: Is this the same as saying, "do this if the DB is running on a solid state disk"?

streamer45 (approved):
Looks very good, brilliant work on these :+1: 

My only question would be how confident are we that there are no unwanted side effects. In other words, before recommending values maybe we should have a testing period (e.g. a couple of weeks).

source/deployment/cluster.rst
agnivade: Lol yea :)

agnivade:
> My only question would be how confident are we that there are no unwanted side effects. In other words, before recommending values maybe we should have a testing period (e.g. a couple of weeks).

I have tested each setting individually and verified that it does what it's supposed to do, as you have already seen in the cloud-engineering channel. All these are very basic and safe to modify IMO. I have erred on the side of caution and tried to be as conservative as possible.

Although I think adding a line about the issues with cargo-culting won't hurt ;)

Totally fine with a testing period though, can you clarify what you mean by this? Who will do the testing? And how?

streamer45:
> I have tested each setting individually and verified that it does what it's supposed to do, as you have already seen in the cloud-engineering channel. All these are very basic and safe to modify IMO. I have erred on the side of caution and tried to be as conservative as possible.

Absolutely :+1: 

> Totally fine with a testing period though, can you clarify what you mean by this? Who will do the testing? And how?

I was thinking Cloud would be start applying these setting soon so that would be more than sufficient.

agnivade:
> I was thinking Cloud would be start applying these setting soon so that would be more than sufficient.

Ah I see. Sure, let's wait. Although it might take some time for Cloud to set all of these.

agnivade:
So to confirm, you are good with merging this. But hold off from recommending this to our customers right?

streamer45:
> So to confirm, you are good with merging this. But hold off from recommending this to our customers right?

Yes that works :+1:

PR #4195, opened 2020-12-21
Read-Only admin role name changed to Viewer

cwarnermm:
The name of the read-only admin role has been updated so that the product documentation matches the product UX.

Updated:
- Self-Managed Admin Guide > Onboard Users > Additional System Admin Roles (E20)


justinegeffen (approved):

wiersgallak (approved):
Thanks!

PR #4219, opened 2021-01-06

amyblais:
Updating note about TLS deprecation as it's not taking effect in v5.31.

### DIFF
--- a/source/administration/important-upgrade-notes.rst
+++ b/source/administration/important-upgrade-notes.rst
@@ -8,7 +8,7 @@ Important Upgrade Notes
    Support for Mattermost Server v5.19 `Extended Support Release <https://docs.mattermost.com/administration/extended-support-release.html>`_ has come to the end of its lifecycle. Upgrading to v5.25 or later is required.
 
 .. important::
-   TLS versions 1.0 and 1.1 have been deprecated by browser vendors. Starting in v5.31 (January 16, 2021) mmctl will return an error when connected to Mattermost servers deployed with those TLS versions. System Admins will need to explicitly add a flag in their commands to continue to use them. We recommend upgrading to TLS version 1.2 or higher.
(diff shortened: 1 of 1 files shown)


source/administration/important-upgrade-notes.rst
```diff
@@ -8,7 +8,7 @@ Important Upgrade Notes
    Support for Mattermost Server v5.19 `Extended Support Release <https://docs.mattermost.com/administration/extended-support-release.html>`_ has come to the end of its lifecycle. Upgrading to v5.25 or later is required.
 
 .. important::
-   TLS versions 1.0 and 1.1 have been deprecated by browser vendors. Starting in v5.31 (January 16, 2021) mmctl will return an error when connected to Mattermost servers deployed with those TLS versions. System Admins will need to explicitly add a flag in their commands to continue to use them. We recommend upgrading to TLS version 1.2 or higher.
+   TLS versions 1.0 and 1.1 have been deprecated by browser vendors. In an upcoming release (version TBD), mmctl will return an error when connected to Mattermost servers deployed with these TLS versions and System Admins will need to explicitly add a flag in their commands to continue to use them. We recommend upgrading to TLS version 1.2 or higher.
```
cwarnermm, line 11:
```suggestion
   TLS versions 1.0 and 1.1 have been deprecated by browser vendors. In an upcoming release (version TBD), ``mmctl`` will return an error when connected to Mattermost servers deployed with these TLS versions. System Admins will need to explicitly add a flag in their commands to continue to use these TLS versions. We recommend upgrading to TLS version 1.2 or higher.
```

wiersgallak (approved):
Thanks, @amyblais

source/administration/important-upgrade-notes.rst
cwarnermm: If I'm reading this correctly, we're documenting that at some point in the future (currently TBD), mmctl will return errors and system admins will need to take action. It's a good forewarning to customers, but from the customer's perspective, there isn't a clear timeline for taking action.

What happens with mmctl right now in cases where a customer is using TLS 1.0 or 1.1? Is there anything we can recommend system admins do now to prepare?

cwarnermm (commented):
Comments/questions left inline

PR #4222, opened 2021-01-07
MM-30159 Allow moving multi-selected channels with channel context menu

cwarnermm:
Documentation for: https://mattermost.atlassian.net/browse/MM-30159

Updated:
- User Guide > Getting Started > Organizing Conversations > Experimental: Channel Organization Features
   - Expanded the bullet point about drag and drop to include additional options (available through any channel's 3-dot icon)

### DIFF
--- a/source/help/getting-started/organizing-conversations.rst
+++ b/source/help/getting-started/organizing-conversations.rst
@@ -154,6 +154,8 @@ When configured by the System Admin, users can enable the features in **Account 
 
 **Collapsible custom categories**: Create custom categories in the sidebar to group channels together for easier navigation (e.g. “Design” or “Marketing”). Collapse categories to display only unread channels and reduce unnecessary scrolling.
 
+**Move channels and categories**: Prioritize important conversations by dragging and dropping channels between or within categories, or drag channels to reorder entire categories. You can also select the **More options** icon, then select **Move to**, followed by a category destination. 
+
 **Drag and drop channels and categories**: Drag channels between or within categories, or drag to reorder entire categories to prioritize important conversations.
 
 **Sort Direct Messages by recent conversations**: Choose to sort Direct Messages alphabetically or by recent conversations first. 


source/help/getting-started/organizing-conversations.rst
```diff
@@ -154,6 +154,8 @@ When configured by the System Admin, users can enable the features in **Account
 
 **Collapsible custom categories**: Create custom categories in the sidebar to group channels together for easier navigation (e.g. “Design” or “Marketing”). Collapse categories to display only unread channels and reduce unnecessary scrolling.
 
+**Move channels and categories**: Prioritize important conversations by dragging and dropping channels between or within categories, or drag channels to reorder entire categories. You can also select the **More options** icon, then select **Move to**, followed by a category destination.
```
justinegeffen, line 157:
```suggestion
**Move channels and categories**: Prioritize important conversations by dragging and dropping channels between or within categories, or drag channels to reorder entire categories. You can also select the **More options** icon, then select **Move to**, followed by a category destination.
```

justinegeffen (approved):

esethna:
@cwarnermm I think we need to highlight the multi-select feature in general, rather than the move to function.

Specifically, users can now ctrl (CMD on Mac) or shift click to select groups of channels, then drag to reorder or "move to" categories

esethna (approved):

source/help/getting-started/organizing-conversations.rst
```diff
@@ -154,6 +154,8 @@ When configured by the System Admin, users can enable the features in **Account
...
```
amyblais, line 157: There should be a mention that this is only available in Cloud and will be available for on-prem later in v5.32.

cwarnermm: Agreed! Thanks! Worked with the PMs to define a process for such cases. The documentation now clarifies when the feature will be available for self-hosted deployments.

source/help/getting-started/organizing-conversations.rst
```diff
@@ -154,6 +154,8 @@ When configured by the System Admin, users can enable the features in **Account
 
 **Collapsible custom categories**: Create custom categories in the sidebar to group channels together for easier navigation (e.g. “Design” or “Marketing”). Collapse categories to display only unread channels and reduce unnecessary scrolling.
 
+**Move channels and categories**: Specify a destination for selected channels using the **Move to** under the **More options** icon. This feature is available today for Cloud deployments and will be available for self-hosted deployments in Mattermost Server v5.32 and later.
```
esethna, line 157: @cwarnermm I think the way the bullets are ordered here lessens the impact of drag and drop which is the primary way for people to move channels. I think of "move to" as a secondary way to move channels, hence why it may be better as a sub point under drag and drop. Also move to existed before v5.32, it's just that you can now multi-select > move to that's the new feature, so probably not even worth complicating the docs for that. 

Propose:

**Collapsible custom categories**: Create custom categories in the sidebar to group channels together for easier navigation (e.g. “Design” or “Marketing”). Collapse categories to display only unread channels and reduce unnecessary scrolling.

**Drag and drop channels and categories**: Drag selected channels between or within categories, or drag to reorder entire categories to prioritize important conversations. You can also specify a destination for selected channels using the **Move to** under the **More options** icon.

esethna (changes requested):

source/help/getting-started/organizing-conversations.rst
cwarnermm: Thanks, @esethna. This section underwent a number of individual changes recently, and it's good to step back and look at the section as a whole.

esethna (approved):

PR #4223, opened 2021-01-07
MM-31085 Create db init CLI command

cwarnermm:
Documentation for: https://mattermost.atlassian.net/browse/MM-31085

Updated:
- Self-Managed Admin Guide > Administration > Command Line Tools > Mattermost 3.6 and later
   - Added new mattermost db init command documentation

### DIFF
--- a/source/administration/command-line-tools.rst
+++ b/source/administration/command-line-tools.rst
@@ -743,7 +743,7 @@ mattermost db init
 ------------------
 
   Description
-    Initializes the database for a given DSN (data name source), executes migrations, and loads custom defaults when specified.
+    Initializes the database for a given data source name (DSN), executes migrations, and loads custom defaults when specified.
 
   Format
     .. code-block:: none
--- a/source/administration/command-line-tools.rst
+++ b/source/administration/command-line-tools.rst
@@ -738,6 +738,37 @@ mattermost config validate
       .. code-block:: none
 
         bin/mattermost config validate
+	
+mattermost db init
+------------------
+
+  Description
+    Initializes the database for a given DSN (data name source), executes migrations, and loads custom defaults when specified.
+
+  Format
+    .. code-block:: none
+
+      mattermost db init
+
+  Examples
+  
+    Use the ``config`` flag to pass the DSN:
+    
+    .. code-block:: none
+
+       mattermost db init --config postgres://localhost/mattermost
+       
+    Run this command to use the ``MM_CONFIG`` environment variable:
+    
+    .. code-block:: none
+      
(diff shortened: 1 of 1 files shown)


source/administration/command-line-tools.rst
```diff
@@ -738,6 +738,37 @@ mattermost config validate
       .. code-block:: none
 
         bin/mattermost config validate
+	
+mattermost db init
+------------------
+
+  Description
+    Initializes the database for a given DSN (data name source), executes migrations, and loads custom defaults when specified.
```
wiersgallak, line 746: Just confirming "DSN" = "data name source"? I would expect the acronym to follow the words.

cwarnermm: Good question, @wiersgallak! Common convention often puts an acronym's expansion ahead of the acronym itself, while the expansion of an initialism (such as DSN) follows. 
- Expansion is what the term means when written out in full. 
- Acronyms are abbreviations that use initials, but the initials are pronounced as a word rather than speaking the individual letters aloud. 
- Initialisms are when you abbreviate a word to only its initials.

All this being said, if we tend to position expansions first in MM docs, I can update this for consistency. I do recommend keeping both a short and long version to avoid ambiguity since DSN can mean different things to different audiences. CC @justinegeffen

mgdelacroix (approved):

source/administration/command-line-tools.rst
justinegeffen, line 746:
```suggestion
    Initializes the database for a given data source name (DSN), executes migrations, and loads custom defaults when specified.
```

cwarnermm: Circling back on this -- We're aligned on documenting acronyms and initialisms such that the expansion comes ahead of the acronym/initialism itself.

justinegeffen: That's a really helpful breakdown, @cwarnermm. I'm 4/5 on having it as "data source name (DSN)" purely for consistency. I'd like to add that breakdown to the style guide at some stage because it's very helpful.

justinegeffen (approved):

source/administration/command-line-tools.rst
cwarnermm: Agreed, @justinegeffen  - Consistency with existing docs is best in this case. I'm not so sure that the differences between acronyms and initialisms is widely understood enough ;)

justinegeffen:
Gentle reminder on this, @wiersgallak. :)

wiersgallak (approved):

PR #4225, opened 2021-01-08
MM-30048 Implement new websocket events for collapsed threads

cwarnermm:
Documentation for: https://mattermost.atlassian.net/browse/MM-30048

Updated:
-  Self-Managed Admin Guide > Configure Mattermost > Configuration Settings > Experimental Settings only in config.json > Service Settings
   - Added new Collapsed Threads configuration setting


amyblais:
May also need to be documented in the Telemetry doc.

cwarnermm:
@reflog - In the Telemetry documentation, CollapsedThreads is currently documented as a bool, rather than a string. Please let me know if that's technically correct or if changes are needed. Thanks!

reflog:
@cwarnermm - it was a boolean initially, but not it's a tri-state: "off/default_on/default_off".
off - turned off completely
default_off - each user has to opt-in
default_on - each user has to opt-out

reflog (approved):

cwarnermm:
Thanks, @reflog -- are tri-states also documented as bools in the Telemetry documentation? (they appear to be - just want to confirm).

reflog:
Yup, it's the same in telemetry

esethna:
@cwarnermm let's hold off on any docs for PRs for collapsed reply threads for now. In fact we should probably figure out our approach to documenting the entire feature once we are closer to implementation rather than piecemeal

amyblais:
I'll close this for now so that it doesn't get introduced to the v5.31-documentation branch.

PR #4238, opened 2021-01-13
Add Sphinx_reredirects extension

jaydeland:
#### Summary
looking to install a Sphinx extension that'll allow us to use 301 redirects. I've found one but am stumped in terms of how to install it. These instructions are pretty clear for local installation/use: https://pypi.org/project/sphinx-reredirects/. But I need to install the extension globally. I don't know if it needs to be added to a file in the repo with the extension's URL or somehow uploaded into the repo.

#### Ticket Link
https://mattermost.atlassian.net/browse/DOPS-274

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -v -E
 SPHINXBUILD   = pipenv run sphinx-build
 PAPER         =
 BUILDDIR      = build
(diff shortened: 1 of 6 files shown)


justinegeffen (approved):

justinegeffen:
This makes sense to me. So from looking at the redirect you added, if this works as expected, that URL should land up at cloud/cloud-integrations.html?

jaydeland:
> This makes sense to me. So from looking at the redirect you added, if this works as expected, that URL should land up at cloud/cloud-integrations.html?

I had success using an existing docname with the redirect being generated.

khos2ow (approved):
I'm not all too familiar with Sphinx here, feel free to proceed on your discretion if manual testing passes.

PR #4254, opened 2021-01-20
Update troubleshooting.rst

amyblais:
For https://forum.mattermost.org/t/mattermost-resets-config-file-periodically-on-startup/11116/8.

**Notes for reviewers:**
1. Need guidance on what other details should be added.
2. Suggestions for other locations are welcome. I added this to the troubleshooting doc since this issue is not related to a specific release, so I wasn't sure where to add it in the important upgrade notes.

### DIFF
--- a/source/install/troubleshooting.rst
+++ b/source/install/troubleshooting.rst
@@ -133,6 +133,11 @@ Multi-Factor Authentication (MFA)
 
 If MFA has stopped working, confirm that your server clock is not out of sync. On Unix systems you can view the hardware time of your system using ``# hwclock -r``. If you're using systemd you can also use ``# timedatectl``.
 
+Config.json Resets to Default Values
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


source/install/troubleshooting.rst
```diff
@@ -133,6 +133,11 @@ Multi-Factor Authentication (MFA)
 
 If MFA has stopped working, confirm that your server clock is not out of sync. On Unix systems you can view the hardware time of your system using ``# hwclock -r``. If you're using systemd you can also use ``# timedatectl``.
 
+config.json Resets to Default Values
```
agnivade, line 136: Actually, I might have been a bit hasty with this one without actually verifying it myself. I am trying to reproduce this but unable to. I was thinking that running the binary as root will take a different config file location, but that's not what's happening from what I can see.

agnivade:
@amyblais - I think we can close this. I opened a ticket to investigate this first and get a better understanding of what's happening. We can then see if it's a bug or something to be documented.

PR #4170, opened 2020-12-11
SEO Updates - Find and Replace Chained Redirects

lawrencejohnson:
#### Summary
Per SEO Request, existing links that resulted in redirects were examined and updated to point to their final destination.

#### Ticket Link
https://mattermost.atlassian.net/browse/MAR-187

PR #4271, opened 2021-01-27
Fix broken links

amyblais:
Unsure if this is the correct fix, but https://developers.mattermost.com/integrate/ seems to be a broken page


justinegeffen (approved):

justinegeffen:
Thanks @amyblais. The links were updated as part of https://github.com/mattermost/docs/pull/4170/files and as they were existing links I didn't think to validate them. Thanks for picking this up!

PR #4286, opened 2021-02-03
Replaced inherited layout with full layout to add article markup

lawrencejohnson:
#### Summary

Pulled full theme layout from theme repo (link below) to customize the full markup output, primarily for adding article tag markup per SEO team.

https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/layout.html

#### Ticket Link

https://mattermost.atlassian.net/browse/MAR-222


justinegeffen:
Hi @lawrencejohnson! I've taken a look at this and it looks good to me. However, I'd like to confirm what has actually changed as it's not possible to check via preview (previews don't spin up if the PR is from a fork). Alternatively you can let me know what to look for, and I'll clone your fork to do a local build.  :)

lawrencejohnson:
> Hi @lawrencejohnson! I've taken a look at this and it looks good to me. However, I'd like to confirm what has actually changed as it's not possible to check via preview (previews don't spin up if the PR is from a fork). Alternatively you can let me know what to look for, and I'll clone your fork to do a local build. :)

Hi @justinegeffen, so in terms of appearance, everything is actually supposed to look and work exactly the same. There should be no difference in content even meta data. Even the rendered HTML is nearly identical except one particular difference. There is a wrapping div with role "main" that we are swapping out to an article tag.

Old HTML output
`<div role="main" class="document" itemscope="itemscope" itemtype="http://schema.org/Article">`

New HTML output
`<article class="document" itemscope="itemscope" itemtype="https://schema.org/Article">`

Unfortunately, there was no way to make this change to the markup without using our own layout, so what I did was just download the layout that was being inherited and tweak it as needed.

If I was to make a recommendation, it would be to clone and build locally for review. It affects almost every page, but the only three page types I've found that were actually "unique" are the home page (not affected), the search page, and all the other pages. I was pretty thorough in my QA, but a second set of eyes would probably be for the best.

justinegeffen:
@lawrencejohnson, I'm unable to build this locally (I completely broke my entire local environment and I'm battling to get it up and running on this cloned repo for some reason). @cwarnermm, would you be open to running a local build of this? I don't want to block this from being merged. :) Thank you!

lawrencejohnson:
@justinegeffen oh that's a bummer. I'm not sure how much help I can offer since I definitely had some pain points on my initial setup as well, but if you think it would be helpful, I may be able to see if I can spot the issue you're having via screen-share. My setup is on a windows device.

justinegeffen:
Thanks @lawrencejohnson! I managed to get it working - had to do a bunch of updates on my virtual environments. This looks great, and all seems to be working fine. :)

justinegeffen (approved):

PR #4214, opened 2021-01-05
MM-27312 Allow uploading 10 files per post

cwarnermm:
Documentation for: https://mattermost.atlassian.net/browse/MM-27312

Updated:
- User's Guide > Messaging > Sharing Files
   - Clarified references to "attachments" to indicate file attachments (rather than Slack attachments)
   - Updated "File Previewer" section title to "Preview File Attachments" 
   - Updated instances of "click" to "select"
   - Updated Tip to provide correct System Console route to Public Links setting
   - Under Attachment Limits and Sizes, added Cloud support for up to 10 file attachments, and clarified that a max of 5 file attachments remains as-is for TE and EE

PR #4301, opened 2021-02-08
Cloud - File Attachment Updates

cwarnermm:
Docs ticket: https://mattermost.atlassian.net/browse/MM-32617

Updated:
- User's Guide > Messaging > Sharing Files > Attachment Limits and Sizes
   - max of 10 files can be attached per post (available in both Cloud & self-managed as of v5.32)
   - default max file size is now 100MB (from 50MB) for Cloud; will be available for self-managed as of v5.33

### DIFF
--- a/source/help/messaging/attaching-files.rst
+++ b/source/help/messaging/attaching-files.rst
@@ -51,6 +51,6 @@ Other document previews (such as Word, Excel, or PPT) are not yet supported.
 Attachment Limits and Sizes
 ---------------------------
 
-The ability to attach a maximum of 10 files per post is available today for Cloud deployments and will be available for self-hosted deployments in Mattermost Server v5.32 and later.
+A maximum of 10 files can be attached per post.
 
-The default maximum file size is 100 MB, but this can be changed by the System Admin.
+The default maximum file size for Cloud deployments is 100 MB, and this file size increase will be available for self-hosted deployments from Mattermost Server v5.33. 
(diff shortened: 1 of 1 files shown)


source/help/messaging/attaching-files.rst
```diff
@@ -51,6 +51,6 @@ Other document previews (such as Word, Excel, or PPT) are not yet supported.
 Attachment Limits and Sizes
 ---------------------------
 
-The ability to attach a maximum of 10 files per post is available today for Cloud deployments and will be available for self-hosted deployments in Mattermost Server v5.32 and later.
+A maximum of 10 files can be attached per post.
 
-The default maximum file size is 100 MB, but this can be changed by the System Admin.
+The default maximum file size for Cloud deployments is 100 MB, and this file size increase will be available for self-hosted deployments from Mattermost Server v5.33.
```
esethna, line 56: > and this file size increase will be available for self-hosted deployments from Mattermost Server v5.33.

Admins can already change it today on-prem, so perhaps it would be better to state that instead of the "file size increase being available" to on-prem

esethna (changes requested):

esethna (approved):

amyblais (approved):

amyblais:
This feature is getting reverted https://github.com/mattermost/mattermost-server/pull/16907 due to performance reasons, so we need to revert this section to what it used to be before https://github.com/mattermost/docs/pull/4214/files was merged.

amyblais:
I believe this is replaced with https://github.com/mattermost/docs/pull/4305.

PR #4323, opened 2021-02-18
[MM-33009] Set correct HTTP version for websocket upgrade on NGINX

streamer45:
#### Summary

We were failing to set the HTTP version to 1.1 for NGINX deployments. This started failing after the recent websocket revamp in https://github.com/mattermost/mattermost-server/pull/16620.

#### Ticket

https://mattermost.atlassian.net/browse/MM-33009

@amyblais We might also want to add an upgrade note for the release that will include the linked PR changes since some users/customers might be using the stock config as suggested and upgrading would most probably break websocket connections.


agnivade (approved):
Great catch!

amyblais (approved):

amyblais:
Adding an upgrade note via https://github.com/mattermost/docs/pull/4355.

agnivade:
I don't think there is any harm in merging this. The directive can stay for prior versions and there shouldn't be any issue.

streamer45:
> I don't think there is any harm in merging this. The directive can stay for prior versions and there shouldn't be any issue.

Yes, it was wrong before as well just that the previous implementation was a bit more lenient.

PR #4340, opened 2021-02-24
v5.33 changelog

amyblais:


esethna (approved):

source/administration/changelog.md
```diff
@@ -4,6 +4,127 @@ This changelog summarizes updates to [Mattermost Team Edition](https://mattermos
...
+### Breaking Changes
+ - Deleting a reaction is now a soft delete in the ``Reactions`` table. A schema update is required and may take up to 15 seconds on first run with large data sets.
+ - Improved the websocket implementation by using epoll to manually read from a websocket connection. As a result, the number of goroutines is expected to go down by half. This implementation is only available on Linux and FreeBSD-based distributions. If you are using NGINX as a proxy to Mattermost, please ensure to have ``proxy_http_version 1.1;`` in the block that handles the websocket path.
+ 
+**IMPORTANT:** If you upgrade from a release earlier than v5.32, please read the other [Important Upgrade Notes](https://docs.mattermost.com/administration/important-upgrade-notes.html).
+
+### Highlights
+
+#### OpenID Connect for OAuth 2.0 Authentication (E20 Edition)
+ - OpenID Connect enables authentication to Mattermost using any OAuth 2.0 provider that adheres to the OpenID Connect specification. This feature is available for Mobile Apps in v1.40 release.
```
cwarnermm, line 20: @esethna We have mobile support for OpenID Connect coming in a future release. Should this changelog or the product documentation be changed?

source/administration/changelog.md
```diff
@@ -4,6 +4,127 @@ This changelog summarizes updates to [Mattermost Team Edition](https://mattermos
...
+
+#### Support Packet Generation (E10 & E20 Editions)
+ - Mattermost provides the ability to download configuration details, logs, and other deployment information when requesting commercial support for Mattermost self-managed E10 or E20 Enterprise editions, or Mattermost Cloud editions.
+
+#### Updated Incident Collaboration plugin to 1.4.0 (E20 Edition)
+ - Added an Incident Timeline, to add support for status updates and other key events to be shown in the right-hand sidebar in a chronological order, allowing users to easily gather information for post-incident reports.
```
cwarnermm, line 26:
```suggestion
 - Added an Incident Timeline to add support for status updates and other key events to be shown in the right-hand sidebar in a chronological order. The timeline enables users to easily gather information for post-incident reports.
```

source/administration/changelog.md
amyblais: @cwarnermm Would you have more details on which future release? I thought it's already released for Mobile.

cwarnermm (commented):
Reads very well! Non-blocking suggestion provided inline.

source/administration/changelog.md
```diff
@@ -4,6 +4,127 @@ This changelog summarizes updates to [Mattermost Team Edition](https://mattermos
 
 Also see [changelog in progress](https://bit.ly/2nK3cVf) for the next release.
 
+## Release v5.33 - [Feature Release](https://docs.mattermost.com/administration/release-definitions.html#feature-release)
+
+**Release Day: 2021-03-16**
+
+### Breaking Changes
+ - Deleting a reaction is now a soft delete in the ``Reactions`` table. A schema update is required and may take up to 15 seconds on first run with large data sets.
```
esethna, line 12: Not sure I understand the impact of this breaking change, why it exists, or what action items I have to take, or what happens if I don't do anything (from the perspective of a sys admin)

source/administration/changelog.md
```diff
@@ -4,6 +4,127 @@ This changelog summarizes updates to [Mattermost Team Edition](https://mattermos
...
+ - Improved the websocket implementation by using epoll to manually read from a websocket connection. As a result, the number of goroutines is expected to go down by half. This implementation is only available on Linux and FreeBSD-based distributions. If you are using NGINX as a proxy to Mattermost, please ensure to have ``proxy_http_version 1.1;`` in the block that handles the websocket path.
```
esethna, line 13: Not sure I understand what is the breaking change here?

"Improved the websocket implementation by using epoll to manually read from a websocket connection. As a result, the number of goroutines is expected to go down by half. This implementation is only available on Linux and FreeBSD-based distributions."

Perhaps we need to rephrase this bullet to identify the breaking change and why it exists

esethna (changes requested):

source/administration/changelog.md
amyblais: @wiggin77 - please help share details here.

source/administration/changelog.md
amyblais: @streamer45 Please help with details here.

streamer45: Maybe that goes a bit into too much detail for a release note. The gist is that we improved the performance of handling websocket connections on servers running on Linux/Unix platforms (more or less anything except Windows). 

As part of this we discovered that we were suggesting incorrect NGINX configurations that will break websocket connectivity if used with the new release so the only action required is for admins deploying Mattermost behind NGINX. If that's the case they should review and possibly update their NGINX configuration by adding the suggested line to the websocket related block. 

cc @agnivade

agnivade: The breaking change is only from an nginx perspective. The new code changes require nginx to send the http version header, whereas previous versions did not. The first two sentences don't really explain the need for the breaking change. It's better for a release note. I would rephrase to something like:

```
A new websocket implementation requires the HTTP version to be 1.1. If you are using NGINX as a proxy to Mattermost, please ensure to have ``proxy_http_version 1.1;`` in the block that handles the websocket path.
```

source/administration/changelog.md
wiggin77: I don't know what has been defined as a breaking change, but this is just a schema change and doesn't change how any APIs behave.  The need for a release notice is simply to warn the admin that the database migration may take longer than usual.

source/administration/changelog.md
```diff
@@ -170,6 +294,7 @@ Mattermost v5.31.0 contains a low level security fix. [Upgrading](https://docs.m
  
 ### Known Issues
  - The Database Schema Version is displayed as 5.30.0 in the About Mattermost modal.
+ - Config.json can reset when running any ``sudo -u mattermost ./bin/mattermost`` commands or when running ``systemctl restart mattermost`` [MM-32390](https://mattermost.atlassian.net/browse/MM-32390).
```
streamer45, line 297: This is not necessarily true. Only affected commands would be those that write to config (e.g. `config` or `plugin`). For example, we [fixed](https://github.com/mattermost/mattermost-server/pull/16805) other frequently used commands like `version` to be read-only so they won't cause this issue. Also, I am curious about the `systemctl restart mattermost` case. This wasn't reported originally and I am not sure how it could be causing the issue.

source/administration/changelog.md
amyblais: I've added the suggested note in the breaking changes section, and moved the first two sentences to the Websocket section below.

source/administration/changelog.md
amyblais: Thanks, I've fixed the text.

streamer45 (approved):

adamjclarkson (approved):

source/administration/changelog.md
amyblais: @streamer45  I added this ticket to known issues: https://mattermost.atlassian.net/browse/MM-33752

Issue #4342, opened 2021-02-24
UseGossip setting should be UseExperimentalGossip


amyblais:
More context at: https://community.mattermost.com/core/pl/q5uw3iig9i8niesfd6gjtewo3c

The docs for https://docs.mattermost.com/administration/config-settings.html#use-gossip needs updating. The change from ``UseExperimentalGossip`` to ``UseGossip`` was done only to the UI. The config setting is still called ``UseExperimentalGossip``.

This is part of a two step process we are taking in transitioning the feature from beta to stable. We have kept the config field as is to avoid breaking changes, and have just flipped the default to true now. And changed the text in UI.

Later, we will be entirely removing the ``UseExperimentalGossip`` setting altogether and make it the only mode.

cwarnermm:
Documentation updated via: https://github.com/mattermost/docs/pull/4347

PR #4347, opened 2021-02-24
UseGossip setting should be UseExperimentalGossip

cwarnermm:
Documentation for: https://github.com/mattermost/docs/issues/4342

Updated:
- Self-Managed Admin Guide > Configure Mattermost > Configuration Settings > Use Gossip
   - Updated ``config.json`` configuration setting from ``UseGossip`` to ``UseExperimentalGossip``

### DIFF
--- a/source/administration/config-settings.rst
+++ b/source/administration/config-settings.rst
@@ -987,9 +987,9 @@ Use Gossip
 
 Note that the gossip port and gossip protocol are used to determine cluster health even when this setting is ``false``.
 
-+-------------------------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


amyblais (approved):

agnivade (approved):
I am wondering what led to the change in the first place? Just checking if there was any confusion on my end.

cwarnermm:
@agnivade - I made an assumption that the word "experimental" would be removed from both the UI and from the config.json configuration setting. That assumption resulted in this error. Going forward, instead of making such an assumption, I'll confirm my understanding prior to publishing updated documentation.

agnivade:
Ahh I see. Gotcha. No worries @cwarnermm ! Thanks for the clarification!

PR #4350, opened 2021-02-26
add mmctl user verify

coltoneshaw:
#### Summary
Updated the docs for mmctl user verify

https://github.com/mattermost/mmctl/blob/release-5.32/commands/user.go#L168


coltoneshaw:
Copied and changed the docs from our [mattermost user verify command](https://docs.mattermost.com/administration/command-line-tools.html#mattermost-user-verify)

source/administration/mmctl-cli-tool.rst
```diff
@@ -3938,6 +3938,7 @@ Child Commands
   -  `mmctl user delete`_ - Delete users
   -  `mmctl user deleteall`_ - Delete all users and all posts (local command only)
   -  `mmctl user email`_ - Set user email
+  -  `mmctl user verify`_ - Verify the email address of a new user.
```
cwarnermm, line 3941:
```suggestion
  -  `mmctl user verify`_ - Verify the email address of a new user
```

source/administration/mmctl-cli-tool.rst
```diff
@@ -4170,6 +4171,41 @@ mmctl user email
...
+
+.. code-block:: sh
+
+    -h, --help       help for email
+
+**Options inherited from parent commands**
+
+.. code-block:: sh
+
+   --format string               the format of the command output [plain, json] (default "plain")
```
cwarnermm, line 4204: Based on the [mmctl repository](https://github.com/mattermost/mmctl/blob/master/docs/mmctl_user_verify.rst), there are 6 inherited options from parent commands:

--config-path string           path to the configuration directory. If "$HOME/.mmctl" exists it will take precedence over the default value (default "$XDG_CONFIG_HOME")
--format string                the format of the command output [plain, json] (default "plain")
--insecure-sha1-intermediate   allows to use insecure TLS protocols, such as SHA-1
--insecure-tls-version         allows to use TLS versions 1.0 and 1.1
--local                        allows communicating with the server through a unix socket
--strict                       will only run commands if the mmctl version matches the server one

Can you update this list to match above, please? Thanks so much!

cwarnermm (changes requested):
Thank you so much @coltoneshaw, for updating the product documentation with this missing mmctl command. Very much appreciated! A couple of options inherited from parent commands are missing, but otherwise looks great!

source/administration/mmctl-cli-tool.rst
coltoneshaw: I noticed in the other commands we did not list all the options that were included, so I copied that. I can push the update for this one. Should we also update all the other commands to match?

coltoneshaw: Just updated the mmctl user verify command

cwarnermm: Great question! Many of the commands in the product documentation are currently outdated when compared against the `mmctl` repository -- and this is one of the most common areas of the docs that need updates. I'm in the process of reviewing all documented commands against the `mmctl` repo and updating accordingly.

cwarnermm (approved):
Thanks so much, @coltoneshaw!

PR #4373, opened 2021-03-08
Update Pipfile.lock

justinegeffen:
Sphinx upgrade moved to staging for testing. This PR contains the upgrade from Sphinx 1.84 to 3.4.3. 

Adding editorial, PM, and build reviewers for sanity check and 360 feedback.


justinegeffen:
@lawrencejohnson, I was wondering if you'd be able to help with this: When I run a search on the preview server ([here's an example](https://mattermost.github.io/docs/4373/search.html?q=gitlab&check_keywords=yes&area=default#), I don't get any preview text returned. The screenshots attached are the error I see, as well as the content in the template "layout notoc" that is apparently problematic. 

Any idea how we can fix this? If not, I'm also going to ask @asaadmahmood  if he can run his eye over it. :)

Screen Shot 2021-03-16 at 22 07 06

Screen Shot 2021-03-16 at 22 04 28

lawrencejohnson:
Try changing

`docContent = htmlElement.find('[role=main]')[0];`

to

`docContent = htmlElement.find('article.document')[0];`

justinegeffen:
Thanks, @lawrencejohnson! I'm not actually sure where to change that. The only place I can see a reference to "role" is in this section. Would I remove "role=main" with "role=article.document"?

Screen Shot 2021-03-17 at 11 58 30

I'm also wondering if this is related to the opensearch setting which is currently not set. This is what it's meant to do:
"# If true, an OpenSearch description file will be output, and all pages will
# contain a <link> tag referring to it.  The value of this option must be the
# base URL from which the finished HTML is served.
#
# html_use_opensearch = ''"

Apparently it works like this: "If nonempty, an OpenSearch description file will be output, and all pages will contain a <link> tag referring to it. Since OpenSearch doesn’t support relative URLs for its search page location, the value of this option must be the base URL from which these documents are served (without trailing slash), e.g. "https://docs.python.org". The default is ''."

Should I set the opensearch option? I don't want to cause conflicts with your work!

lawrencejohnson:
If you're not able to edit the code that was in the screenshot you provided (the `htmlToText` function), then you could also try updating `layout.html` so that this section looks like this:

```
<article class="document" itemscope="itemscope" itemtype="https://schema.org/Article">
{%- block document %}
  <div itemprop="articleBody" role="main">
  {% block body %}{% endblock %}
  </div>
  {% if self.comments()|trim %}
  <div class="articleComments">
  {% block comments %}{% endblock %}
  </div>
  {% endif%}
</article>
```

justinegeffen:
Thanks, @lawrencejohnson! That worked and completely resolved the issue. Next issue I have is likely more directly related to you. This is the error I get on all the pages (this is an example page: https://mattermost.github.io/docs/4373/overview/compliance.html). 

Screen Shot 2021-03-18 at 15 17 47

What can I do to resolve that?

justinegeffen:
One more thing, @lawrencejohnson. I see there's a consistent error with modernizr: "Loading failed for the <script> with source “https://mattermost.github.io/docs/4373/_static/js/modernizr.min.js”." In the ``layout.notoc`` file we have the following:

`` {# Keep modernizr in head - https://modernizr.com/docs/#installing #}
  <script src="{{ pathto('_static/js/modernizr.min.js', 1) }}"></script>``

Should we replace Modernizr with something else? Do we need to keep it or can we remove that?

lawrencejohnson:
@justinegeffen - my preference would be to remove modernizr. I tried removing it previously, but it created some issues with the old framework/theme (I believe I'm the one who left that comment to keep it in `head`). If the new build is working as intended while you are getting that 404 error, then that means we don't need it anymore.

justinegeffen:
Thanks Lawrence. The build is working as intended so I'll go ahead and remove it - it'll be pretty obvious whether it's a breaking change fairly quickly. :) Thank you!

justinegeffen:
Awesome, looks like it can be removed safely. This is all I am seeing, error-wise now, @lawrencejohnson:
Screen Shot 2021-03-18 at 20 26 42

lawrencejohnson:
I'm not seeing that message, and I'm still seeing the 404 in the preview environment. I'm assuming there is a latency on push; about how long does it take?

justinegeffen:
Ah - could you hard refresh the page? It doesn't auto-refresh AFAIK. :)

lawrencejohnson:
Cache is fully cleared. On the home page I'm not seeing the 404 (or the warnings you posted), but on sub pages I'm still seeing the 404 (but not the warnings)

https://mattermost.github.io/docs/4373/index.html

What browser are you using to debug?

cwarnermm (commented):
Reviewed; however, my current familiarity/knowledge of this code is low.

justinegeffen:
Thanks, @cwarnermm! You've helped test the changes in the staging preview which was all I needed from your side. :)

source/deployment/sso-ldap.rst
```diff
@@ -90,167 +98,187 @@ Using filters assigns roles to specified users on login. To access AD/LDAP filte
...
 4. Choose **Save**.
 
 When a guest logs in for the first time they are presented with a default landing page until they are added to channels.
 
-See the [Guest Accounts documentation](https://docs.mattermost.com/deployment/guest-accounts.html) for more information about this feature.
+See the `Guest Accounts documentation <https://docs.mattermost.com/deployment/guest-accounts.html>`_ for more information about this feature.
 
-##### Admin Filter
+Admin Filter
+''''''''''''
```
amyblais, line 120: When clicking View File, this page/file looks broken, maybe something with the header formatting here and in other places needs fixing.

justinegeffen: Thanks, @amyblais, that's a good point. I actually meant to remove this change from the PR as it's no longer needed. I'll sort that out. :)

jasonblais:
@justinegeffen Thanks for these changes! I don't have much to comment on here, having this deployed to staging sounds like a good next step.

Adding @wiersgallak if any feedback from her side.

justinegeffen:
Thanks, @jasonblais. That's a good point - this is only going to be merged to staging for now, not prod just yet. That's the next step. So one approving review would be great but won't affect prod. :)

jasonblais (approved):

PR #4378, opened 2021-03-10
Quick improvment for config in db

JtheBAB:
Update the sentence because it is strange how it is described now


source/administration/config-in-database.rst
```diff
@@ -41,7 +41,7 @@ Example output:
 
 Another way is to view your ``config.json`` file and get the value in ``SqlSettings.DataSource``.
 
-If ``SqlSettings.DataSource`` does not start with ``postgres://`` then you have to add ``mysql://`` to the beginning. Also, if you see ``\u0026`` replace it with ``&``.
+If ``SqlSettings.DataSource`` does not start with ``postgres://`` or ``mysql://`` then you have to add it to the beginning depending which database is in use. Also, if you see ``\u0026`` replace it with ``&``.
```
cwarnermm, line 44:
```suggestion
If ``SqlSettings.DataSource`` does not start with ``postgres://`` or ``mysql://``, then you have to add this line to the beginning based on the database in use. Also, if you see ``\u0026`` replace it with ``&``.
```

cwarnermm (changes requested):
Thanks so much, @JtheBAB, for this great contribution towards better docs clarity. I've provided feedback inline.

cwarnermm (approved):
Thanks so much!

PR #4392, opened 2021-03-16
Update changelog.md

amyblais:


source/administration/changelog.md
```diff
@@ -4,6 +4,10 @@ This changelog summarizes updates to [Mattermost Team Edition](https://mattermos
 
 Also see [changelog in progress](https://bit.ly/2nK3cVf) for the next release.
 
+## Release v5.33 - [Feature Release](https://docs.mattermost.com/administration/release-definitions.html#feature-release)
+
+**Note: v5.33.0 is currently delayed until tomorrow as additional release testing time is needed.**
```
amyblais, line 9: I know that this is not the full reason why it's delayed, but I tried to think of something that would make sense from customer/user perspective.

cwarnermm: "...additional release testing time is needed" feels quite fair and reasonable from a customer's perspective. I imagine customers would prefer us perform additional testing rather than simply pushing out a release for the sake of remaining on schedule.

cwarnermm (approved):
Reads well!

lindalumitchell (approved):
I agree that it sounds good as written. Only question I might have is should it say tomorrow's date instead of "tomorrow," just to help avoid ambiguity? Or even "tomorrow, 2021-03-17"? Either way I think it's fine, though.

PR #4394, opened 2021-03-17
fix and update the silent deployment script for the Windows app

flexo3001:
#### Summary
This fixes a timing problem with the silent install batch script and updates it the config.json to a recent version.


Willyfrog (changes requested):
Thank you for taking the time to improve our docs.

I like a lot of the changes you made, which make sense for this script. Can you also verify that the values prosed match the [default ones in the docs](https://docs.mattermost.com/help/apps/desktop-guide.html#app-options)? I can see that minimize to tray should be false

flexo3001:
> Thank you for taking the time to improve our docs.
> 
> I like a lot of the changes you made, which make sense for this script. Can you also verify that the values prosed match the [default ones in the docs](https://docs.mattermost.com/help/apps/desktop-guide.html#app-options)? I can see that minimize to tray should be false

I copied it from the config.json generated at the first opening of the desktop app. We can change the defaults in the docs too or change them in the desktop app i guess.

Screen Shot 2021-03-25 at 17 39 21

Willyfrog:
my bad, the documentation is outdated, I just reviewed and your changes reflect the latest changes in code which were in fact modified to a new default.

thanks!

Willyfrog (approved):

justinegeffen (approved):

PR #4413, opened 2021-03-24
WIP: Backport docs

amyblais:


source/administration/important-upgrade-notes.rst
```diff
@@ -21,6 +21,7 @@ Important Upgrade Notes
 |                                                    | WebSocket handshakes done with HTTP version lower than 1.1 will result in a warning, and the server will transparently upgrade the version to 1.1 to comply with |
 |                                                    | the WebSocket RFC. This is done to work around incorrect Nginx (and other proxy) configs that do not set the ``proxy_http_version`` directive to 1.1. This       |
 |                                                    | facility will be removed in a future Mattermost version and it is strongly recommended to fix the proxy configuration to correctly use the WebSocket protocol.   |
+|                                                    | **Note:** This requirement was removed in v5.33.2 dot release because ....                                                                                       |
```
amyblais, line 24: @agnivade Please help with details on why the v5.33.0 websocket changes is reverted in v5.33.2.

I thought we could update the upgrade notes and changelog docs to mention that it was reverted, and add details on why.

agnivade: It was reverted mainly due to https://github.com/mattermost/mattermost-server/issues/17213. There are lots of unofficial Mattermost builds out there in several different platforms which broke due to the websocket changes. We realized it was causing much more harm than benefit.

But in any case, I would suggest that customers still apply the HTTP header change to 1.1, to correctly adhere to the spec.

amyblais: @agnivade Makes sense, I just edited the changelog entry for the dot release text.

agnivade (approved):

PR #4414, opened 2021-03-25

justinegeffen:
Update for: https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/495


source/administration/devops-command-center.rst
```diff
@@ -152,6 +152,8 @@ Starting an incident opens an interactive dialog. Provide an incident name and s
 
 The creator of an incident is automatically added as the first member and becomes the commander.
 
+When the incident is created, it is announced in the configured channel. If the channel does not exist or is archived/deleted, the bot posts a message in the channel warning the user about it.
```
justinegeffen, line 155: What's the use case for a channel not existing and being announced? Would it be if a user tries to access it? If so, how would they access it if doesn't exist? Or is it if the channel did exist and was archived/deleted only?

justinegeffen:
@agarciamontoro, for the DB changes, would that be in the telemetry doc only or somewhere else?

source/administration/devops-command-center.rst
agarciamontoro: This is an edge case that may not need docs, actually. The case is: a user configures channel A as the announcement channel in the playbook. For some external reason, channel A is later archived/deleted, but the playbook is not updated with another channel. Then, an incident is created with that playbook, and the system tries to announce the incident creation in channel A: as channel A no longer exists, the bot posts a message in the channel warning the user that the announcement could not be made.

agarciamontoro:
> @agarciamontoro, for the DB changes, would that be in the telemetry doc only or somewhere else?

@justinegeffen, no need to document the DB changes. I wrote that only to make the dev review easier, it's an internal change.

source/administration/devops-command-center.rst
justinegeffen, line 155:
```suggestion
When the incident is created, it is announced in the configured channel. If that channel is archived or deleted, the bot indicates this in a Direct Message channel.
```

agarciamontoro (approved):
Yay!

PR #4450, opened 2021-04-07
Update architecture.rst

justinegeffen:
Minor cleanup of formatting and punctuation.

Dev review: @grundleborg, if you could confirm whether this is all still accurate (for self-managed) that'd be awesome.


cwarnermm (approved):
Great updates! Thanks!

PR #4466, opened 2021-04-09
Update guest-accounts.rst

justinegeffen:
Some minor content tweaks and clarification to roles. Process wording adjusted. 

@cwarnermm, @wiersgallak I am wondering whether the reference to and capitalization of Guest Accounts is correct in this doc as the feature is called Guest Access. The implication is that the name of the feature is Guest Accounts when it's actually called "Guest Access" in the System Console. 

As I wrote this doc I'm happy to review this but it has a Cloud docs impact so let me know your thoughts and whether this would work better.

### DIFF
--- a/source/deployment/guest-accounts.rst
+++ b/source/deployment/guest-accounts.rst
@@ -22,7 +22,7 @@ Guests cannot:
 - Create direct messages or group messages with members who aren’t within the same channel
 - Invite people
 
-Additionally, guests are not automatically added to the default ``Town-square`` and ``Off-topic`` channels upon logging in and need to be invited/added to them manually.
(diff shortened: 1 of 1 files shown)


cwarnermm:
I'd like to see the docs align more closely with the product UI over time. This is a good example of a case where content was initially developed with self-managed deployments in mind, but has equal/similar Cloud impact. I envision content like this being written once in common files, and then shared across both the self-managed and cloud admin guides.

As we work through UI adjustments for consistency with the UX team, we'll be set up well to ensure the docs match UI. And with a modern version of Sphinx, we can finally start correcting the various topic titles and URLs throughout the doc set that have changed.

cwarnermm (approved):
Thanks!

wiersgallak (approved):

PR #4479, opened 2021-04-13

justinegeffen:
Updated terminology and process flow.


source/administration/devops-command-center.rst
```diff
@@ -189,15 +194,16 @@ Incident status updates ensure that stakeholders remain informed about the progr
 
 * From the desktop or browser:
 1. Select the **Incidents** icon in the channel header to open the right-hand sidebar.
-2. Select the **Summary** tab.
-3. Within the **Summary** tab, select **Update Status**.
-4. Add a Markdown-formatted message.
+2. Find the incident you're looking for and select **Go to Incident Channel**.
```
agarciamontoro, line 197: This only happens if you're not in an incident channel, or if you've clicked the back button at the RHS when in an incident channel. Is it worth to differentiate that here in the docs?
(Same comment for the other two repetitions of this line)

justinegeffen: Very good question. Let me look into that for the next PR. Thank you!

PR #4491, opened 2021-04-15
v5.34.1 dot release changelog note

amyblais:


streamer45:
Not sure if this is the place but given the latest changes there's a chance that upgrading will also execute missing migrations that were scheduled for 5.32. To be specific these are:

- Altering of `Posts.FileIds` type (PostgreSQL only)
- Added new column `ThreadMemberships.UnreadMentions`
- Added new column `Channels.Shared`
- Added new column `Reactions.UpdateAt`
- Added new column `Reactions.DeleteAt`

These addition can be lengthy on very big MySQL (version 5.x) installations.

streamer45 (approved):

PR #4514, opened 2021-04-22
Update security overview page following customer questions

jasonblais:
1. Add note that Mattermost is not affected by session fixation, with a link to relevant place in the source code.
2. Add a note that user sessions can be revoked remotely by a System Admin


justinegeffen (approved):
Very minor, non-blocking, suggestion. Otherwise LGTM.

source/overview/security.rst
```diff
@@ -56,7 +56,8 @@ Authentication Safeguards
...
    - Session length, session cache, and idle timeout can be `configured according to your internal policies <https://docs.mattermost.com/administration/config-settings.html#sessions>`__, automatically forcing a user to re-login after a specified period of time.
-   - Remotely `revoke user sessions <https://docs.mattermost.com/help/settings/account-settings.html#view-and-logout-of-active-sessions>`__ across web, mobile devices, and native desktop apps.
+   - Remotely `revoke user sessions <https://docs.mattermost.com/help/settings/account-settings.html#view-and-logout-of-active-sessions>`__ across web, mobile devices, and native desktop apps. User sessions can also be revoked remotely by a System Admin in **System Console > Users*.
+   - Session fixation, where an attacker can trick the user to authenticate with a known session cookie, does not affect Mattermost users as a `new session cookie is set at each login <https://github.com/mattermost/mattermost-server/blob/master/api4/user.go#L1815>`_.
```
jupenur, line 60: Long-lived docs shouldn't link to master because master will probably already look different next week. But I'm not sure if the link to source code is needed at all -- seems like it's more likely to just cause confusion than anything else.

Say we link to this line in the current release branch: A few months go by and a new release comes out. The link is now stale and the code in the latest release might actually not even be recognizable any more. Someone reading the docs could follow the link, realize the same code doesn't exist in the latest release, and think of that as an issue. It's not, because the code just got refactored and works the same way, but the person reading the docs wouldn't know that.

jasonblais: @jupenur fair point, removed link to the source code

source/overview/security.rst
```diff
@@ -56,7 +56,8 @@ Authentication Safeguards
 
    - To protect against brute force attacks, you can set `rate limiting on APIs <https://docs.mattermost.com/administration/config-settings.html#id55>`__, varied by query frequency, memory store size, remote address, and headers.
    - Session length, session cache, and idle timeout can be `configured according to your internal policies <https://docs.mattermost.com/administration/config-settings.html#sessions>`__, automatically forcing a user to re-login after a specified period of time.
-   - Remotely `revoke user sessions <https://docs.mattermost.com/help/settings/account-settings.html#view-and-logout-of-active-sessions>`__ across web, mobile devices, and native desktop apps.
+   - Remotely `revoke user sessions <https://docs.mattermost.com/help/settings/account-settings.html#view-and-logout-of-active-sessions>`__ across web, mobile devices, and native desktop apps. User sessions can also be revoked remotely by a System Admin in **System Console > Users*.
```
jupenur, line 59: Missing asterisk.

```suggestion
   - Remotely `revoke user sessions <https://docs.mattermost.com/help/settings/account-settings.html#view-and-logout-of-active-sessions>`__ across web, mobile devices, and native desktop apps. User sessions can also be revoked remotely by a System Admin in **System Console > Users**.
```

jupenur (changes requested):

jupenur (approved):

PR #4531, opened 2021-04-28
rename network.PNG to network.png

cpanato:
#### Summary
rename network.PNG to network.png

when cloning the repo it complains the name of the file


justinegeffen:
Thanks, Carlos! I've seen this happen before when a file is replaced with the same name. GitHub complains about it. It looks good on my side, so I'm approving it. LMK if it resolves the issue!

justinegeffen (approved):

cwarnermm:
I've also seen the same when there's 2 files with the same filename and only the extension case differs.

cpanato:
found others, I will open a PR to get those fixed

PR #4556, opened 2021-05-06
Allow Preview Environment GitHubAction to run only for Upstream PRs

cpanato:
Allow Preview Environment GitHubAction to run only for Upstream PRs 

I will work on a solution to enable the Preview env for forks PRs. this is is just a quick fix to not have the fail check in the PR


harshilsharma63 (approved):
LGTM 👍

PR #4596, opened 2021-05-19

justinegeffen:
Added note about ad-hoc tasks.

@agarciamontoro, do you think this is sufficient or should we add more detail?

### DIFF
--- a/source/administration/devops-command-center.rst
+++ b/source/administration/devops-command-center.rst
@@ -178,12 +178,16 @@ If the incident channel is private, an existing member of the incident channel m
 Working with tasks
 ~~~~~~~~~~~~~~~~~~
 
-Tasks can be part of pre-configured task templates in playbooks and they can also be added, edited, and removed as needed during an active incident. Any member of the incident channel can work with tasks:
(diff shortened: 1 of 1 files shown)


source/administration/devops-command-center.rst
```diff
@@ -178,12 +178,16 @@ If the incident channel is private, an existing member of the incident channel m
 Working with tasks
 ~~~~~~~~~~~~~~~~~~
 
-Tasks can be part of pre-configured task templates in playbooks and they can also be added, edited, and removed as needed during an active incident. Any member of the incident channel can work with tasks:
+Tasks can be part of pre-configured task templates in playbooks and they can also be added, edited, and removed as needed during an active incident. Any member of the incident channel can work with tasks.
```
cwarnermm, line 181:
```suggestion
Tasks can be part of pre-configured task templates in playbooks, and they can also be added, edited, and removed as needed during an active incident. Any member of the incident channel can work with tasks.
```

cwarnermm (approved):
Reads really well! Added one minor, non-blocking suggestion.

agarciamontoro (approved):
Looks good! Thank you, @justinegeffen :)

PR #4643, opened 2021-06-07
Coltoneshaw postgres13 docs

coltoneshaw:
#### Summary
Updated docs for postgres versions across all platforms since we no longer support a version lower than postgres 10.

### DIFF
--- a/source/install/install-rhel-6-postgresql.rst
+++ b/source/install/install-rhel-6-postgresql.rst
@@ -5,29 +5,25 @@ Installing PostgreSQL Database
 
 1. Log in to the server that will host the database, and open a terminal window.
 
-2. Download the PostgreSQL 9.4 Yum repository.
+2. Download the PostgreSQL 13 Yum repository. For more detailed install instructions visit the PostgreSQL docs site [here](https://www.postgresql.org/download/linux/redhat/).
 
-  ``wget https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-6-x86_64/pgdg-redhat94-9.4-3.noarch.rpm``
+  ``sudo yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-6-x86_64/pgdg-redhat-repo-latest.noarch.rpm``
 
-3. Install the Yum repository from the file that you downloaded.
+3. Install PostgreSQL.
 
-  ``sudo yum localinstall pgdg-redhat94-9.4-3.noarch.rpm``
+  ``sudo yum install -y postgresql13-server``
 
-4. Install PostgreSQL.
+4. Initialize the database.
 
-  ``sudo yum install postgresql94-server postgresql94-contrib``
+  ``sudo service postgresql-13 initdb``
 
-5. Initialize the database.
+5. Set PostgreSQL to start on boot.
 
-  ``sudo service postgresql-9.4 initdb``
-
-6. Set PostgreSQL to start on boot.
-
-  ``sudo chkconfig postgresql-9.4 on``
+  ``sudo chkconfig postgresql-13 on``
 
 7. Start the PostgreSQL server.
 
-  ``sudo service postgresql-9.4 start``
+  ``sudo service postgresql-13 start``
 
 8. Switch to the *postgres* Linux user account that was created during the installation.
 
@@ -62,7 +58,7 @@ Installing PostgreSQL Database
 
 15. Allow Postgres to listen on all assigned IP Addresses.
 
-  a. Open ``/etc/postgresql/9.4/main/postgresql.conf`` as root in a text editor.
+  a. Open ``/var/lib/pgsql/13/data/postgresql.conf`` as root in a text editor.
 
   b. Find the following line:
 
@@ -80,7 +76,7 @@ Installing PostgreSQL Database
 
   **If the Mattermost server and the database are on the same machine**:
 
-    a. Open ``/var/lib/pgsql/9.4/data/pg_hba.conf`` as root in a text editor.
+    a. Open ``/var/lib/pgsql/13/data/pg_hba.conf`` as root in a text editor.
 
     b. Find the following line:
 
@@ -92,7 +88,7 @@ Installing PostgreSQL Database
 
   **If the Mattermost server and the database are on different machines**:
 
-    a. Open ``/var/lib/pgsql/9.4/data/pg_hba.conf`` as root in a text editor.
+    a. Open ``/var/lib/pgsql/13/data/pg_hba.conf`` as root in a text editor.
 
     b. Add the following line to the end of the file, where *{mattermost-server-IP}* is the IP address of the machine that contains the Mattermost server.
 
@@ -100,7 +96,7 @@ Installing PostgreSQL Database
 
 17. Reload PostgreSQL:
 
-  ``sudo service postgresql-9.4 reload``
+  ``sudo service postgresql-13 reload``
 
 18. Verify that you can connect with the user *mmuser*.
 
--- a/source/install/install-rhel-7-postgresql.rst
+++ b/source/install/install-rhel-7-postgresql.rst
@@ -5,38 +5,27 @@ Installing PostgreSQL Database
 
 1. Log in to the server that will host the database, and open a terminal window.
 
-2. Download the PostgreSQL 9.4 Yum repository.
+2. Download the PostgreSQL 13 Yum repository. For more detailed install instructions visit the PostgreSQL docs site [here](https://www.postgresql.org/download/linux/redhat/).
 
-  For RHEL 7
-    ``curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-redhat94-9.4-3.noarch.rpm``
-  For CentOS 7
-    ``curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-centos94-9.4-3.noarch.rpm``
-  For Scientific Linux
-    ``curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-sl94-9.4-3.noarch.rpm``
-  For Oracle
-    ``curl -O https://download.postgresql.org/pub/repos/yum/9.4/redhat/rhel-7-x86_64/pgdg-oraclelinux-9.4-3.noarch.rpm``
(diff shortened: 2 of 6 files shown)


cwarnermm:
Thank you so much, @coltoneshaw, for these amazing docs updates! I've added @streamer45 as a technical reviewer since he and I have been recently discussing the need for these important updates.

I've noticed that most of the PostgreSQL replacements made in this PR are changing older 9.x PostgreSQL references to PostgreSQL 13. I'm wondering if, instead of referencing version 13 specifically, taking a {version} approach would be reasonable? If we were to move to a more generic approach, are there any sub-steps that would be problematic using such an approach from your perspective?

coltoneshaw:
@cwarnermm - I don't disagree, maybe a version approach. It's already in the docs for ubuntu, and RHEL 8. I only added the steps where the default repo didn't seem to include postgres 10+. But we can point them to the postgres redhat link and tell what version to install. Circumventing the install steps in our guide.

streamer45:
My thinking is that the version agnostic approach should be preferred as long as it doesn't hinder the ability of the end user to follow and execute. So, in the case where an OS (or distribution) only has a specific version supported it makes sense to go with that as using anything else would either not work or add unnecessary complexity.

cwarnermm:
@coltoneshaw - Are you open to applying updates using a version-agnostic approach?

coltoneshaw:
I definitely can update these. Just so I understand though. Let's say that the default postgres on a distro is **less than** 10.x, we want to just point them to the postgres docs and say "follow docs to get the latest" then follow along? Right now I just have those docs for the distros that i know have a lower version by default.

coltoneshaw:
The reason I'm trying to be clear is that for:

> So, in the case where an OS (or distribution) only has a specific version supported it makes sense to go with that as using anything else would either not work or add unnecessary complexity.

If this distro uses postgres 9.x they cannot use Mattermost, so we'd want to ensure they're not using that. For example, centos 7 you can do the below which is not what we want them to do since they'd have to then uninstall and upgrade.

```bash
[vagrant@localhost ~]$ sudo yum install postgresql-server postgresql-contrib
...
install stuff
...
[vagrant@localhost ~]$ psql -V
psql (PostgreSQL) 9.2.24
[vagrant@localhost ~]$
```

streamer45:
In the specific case, if `https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm` provides multiple versions then it should be fine to leave the version as a variable with a clear note that they'll need >= 10. 

This is to make things simpler on our side (maintenance) of course but also leave the user with the choice (they might not always want the newest possible version).

wiersgallak (approved):
Thanks!

source/install/install-ubuntu-1804-postgresql.rst
```diff
@@ -14,6 +14,7 @@ Assume that the IP address of this server is 10.10.10.1.
   ``sudo apt install postgresql postgresql-contrib``
 
   When the installation is complete, the PostgreSQL server is running, and a Linux user account called *postgres* has been created.
+  Confirm this version is PostgresSQL version 10 or higher. You can do this by running ``psql -V``. If this is not version 10 or higher you can reference the PostgreSQL documentation [here](https://www.postgresql.org/download/linux/ubuntu/) to install an updated version of PostgreSQL.
```
justinegeffen, line 17:
```suggestion
Confirm this version is PostgreSQL version 10 or higher. You can do this by running ``psql -V``. If this is not version 10 or higher you can reference the PostgreSQL documentation [here](https://www.postgresql.org/download/linux/ubuntu/) to install an updated version of PostgreSQL.
```

source/install/install-ubuntu-1604-postgresql.rst
```diff
@@ -13,7 +13,9 @@ Assume that the IP address of this server is 10.10.10.1
 
   ``sudo apt-get install postgresql postgresql-contrib``
 
-  When the installation is complete, the PostgreSQL server is running, and a Linux user account called *postgres* has been created.
+  When the installation is complete, the PostgreSQL server is running, and a Linux user account called *postgres* has been created. 
+  
+  Confirm this version is PostgresSQL version 10 or higher. You can do this by running ``psql -V``. If this is not version 10 or higher you can reference the PostgreSQL documentation [here](https://www.postgresql.org/download/linux/ubuntu/) to install an updated version of PostgreSQL.
```
justinegeffen, line 18:
```suggestion
Confirm this version is PostgreSQL version 10 or higher. You can do this by running ``psql -V``. If this is not version 10 or higher you can reference the PostgreSQL documentation [here](https://www.postgresql.org/download/linux/ubuntu/) to install an updated version of PostgreSQL.
```

source/install/install-ubuntu-2004-postgresql.rst
```diff
@@ -14,6 +14,8 @@ Assume that the IP address of this server is 10.10.10.1.
   ``sudo apt install postgresql postgresql-contrib``
 
   When the installation is complete, the PostgreSQL server is running, and a Linux user account called *postgres* has been created.
+  
+  Confirm this version is PostgresSQL version 10 or higher. You can do this by running ``psql -V``. If this is not version 10 or higher you can reference the PostgreSQL documentation [here](https://www.postgresql.org/download/linux/ubuntu/) to install an updated version of PostgreSQL.
```
justinegeffen, line 18:
```suggestion
Confirm this version is PostgreSQL version 10 or higher. You can do this by running ``psql -V``. If this is not version 10 or higher you can reference the PostgreSQL documentation [here](https://www.postgresql.org/download/linux/ubuntu/) to install an updated version of PostgreSQL.
```

source/install/install-rhel-6-postgresql.rst
```diff
@@ -5,29 +5,25 @@ Installing PostgreSQL Database
 
 1. Log in to the server that will host the database, and open a terminal window.
 
-2. Download the PostgreSQL 9.4 Yum repository.
+2. Download the PostgreSQL 13 Yum repository. For more detailed install instructions visit the PostgreSQL docs site [here](https://www.postgresql.org/download/linux/redhat/).
```
justinegeffen, line 8:
```suggestion
2. Download the PostgreSQL 13 Yum repository. For more detailed install instructions visit the PostgreSQL docs site `here <https://www.postgresql.org/download/linux/redhat/>`_.
```

source/install/install-rhel-7-postgresql.rst
```diff
@@ -5,38 +5,27 @@ Installing PostgreSQL Database
 
 1. Log in to the server that will host the database, and open a terminal window.
 
-2. Download the PostgreSQL 9.4 Yum repository.
+2. Download the PostgreSQL 13 Yum repository. For more detailed install instructions visit the PostgreSQL docs site [here](https://www.postgresql.org/download/linux/redhat/).
```
justinegeffen, line 8:
```suggestion
2. Download the PostgreSQL 13 Yum repository. For more detailed install instructions visit the PostgreSQL docs site `here <https://www.postgresql.org/download/linux/redhat/>`_.
```

source/install/install-rhel-7-postgresql.rst
```diff
@@ -67,9 +56,9 @@ Installing PostgreSQL Database
 
   ``exit``
 
-15. (Optional) If you use a different server for your database and the Mattermost app server, you may allow PostgreSQL to listen on all assigned IP Addresses. To do so, open ``/etc/postgresql/9.5/main/postgresql.conf`` as root in a text editor. As a best practice, ensure that only the Mattermost server is able to connect to the PostgreSQL port using a firewall.
+15. (Optional) If you use a different server for your database and the Mattermost app server, you may allow PostgreSQL to listen on all assigned IP Addresses. To do so, open ``/var/lib/pgsql/13/data/postgresql.conf`` as root in a text editor. As a best practice, ensure that only the Mattermost server is able to connect to the PostgreSQL port using a firewall.
```
justinegeffen, line 59:
```suggestion
15. (Optional) If you use a different server for your database and the Mattermost app server, you may allow PostgreSQL to listen on all assigned IP addresses. To do so, open ``/var/lib/pgsql/13/data/postgresql.conf`` as root in a text editor. As a best practice, ensure that only the Mattermost server is able to connect to the PostgreSQL port using a firewall.
```

source/install/install-rhel-8-postgresql.rst
```diff
@@ -9,6 +9,8 @@ Installing PostgreSQL Database
 2. Install PostgreSQL.
 
   ``sudo yum install postgresql-server postgresql-contrib``
+  
+  Confirm this version is PostgresSQL version 10 or higher. You can do this by running ``psql -V``. If this is not version 10 or higher you can reference the PostgreSQL documentation [here](https://www.postgresql.org/download/linux/redhat/) to install an updated version of PostgreSQL.
```
justinegeffen, line 13:
```suggestion
Confirm this version is PostgreSQL version 10 or higher. You can do this by running ``psql -V``. If this is not version 10 or higher you can reference the PostgreSQL documentation [here](https://www.postgresql.org/download/linux/redhat/) to install an updated version of PostgreSQL.
```

justinegeffen (commented):
Thanks for this PR. :) I've added some comments - LMK if you'd like another review when you're done. :)

cwarnermm (approved):
Thank you so much for this great documentation update!

cwarnermm:
@streamer45 - Are you open to reviewing this PR for technical accuracy?

coltoneshaw:
@cwarnermm - Thanks! Let me switch some of the language to the RHEL 6 and CentOS 7 docs to be version agnostic since they use the 'latest' download.

coltoneshaw:
@cwarnermm - Just fixed RHEL 6 / 7 to be version agnostic.

Looking at RHEL 8 / Ubuntu the only other way to future proof these docs would be to replace the stand install with the steps to add the postgres repo and specify a version (like rhel 6/7 or [centos](https://www.postgresql.org/download/linux/ubuntu/) ). That way when the version from the OS gets outdated we don't have to many any changes. Not sure what process is more clear.

cwarnermm:
@coltoneshaw - Your updates feel like a very good step forward more generic approaches to our install documentation. @streamer45 - Do you have any additional feedback or recommendations on how to proceed with these specific updates?

streamer45 (approved):
Thanks for keep iterating on this, looks great :+1:

coltoneshaw:
@cwarnermm - Anything left on this before we can merge it?

PR #4679, opened 2021-06-28
clean up upgrade documentation and files created on upgrade

coltoneshaw:
- Added a check for an existing mattermost tar file before unzipping. This can cause issues where it unzips an old version and you push this for the upgrade.
- Tried to clean up the file purge command. This was a wall of text that confused customers. Now it walks through steps to verify custom directories and remove. It may need some additional cleaning up, but waiting to get a few more eyes on this.
- Added an additional delete with confirmation of mattermost*.gz files so no one has an issue like the first bullet in the future.

### DIFF
--- a/source/administration/upgrade.rst
+++ b/source/administration/upgrade.rst
@@ -54,7 +54,7 @@ Location of your local storage directory
 
      wget https://releases.mattermost.com/X.X.X/mattermost-X.X.X-linux-amd64.tar.gz
 
-#. Confirm no other Mattermost zip folders exist in your ``/tmp`` directory. If another version's zip file does exist, delete or rename the file
+#. Confirm no other Mattermost zip folders exist in your ``/tmp`` directory. If another version's zip file does exist, delete or rename the file.
 
    .. code-block:: sh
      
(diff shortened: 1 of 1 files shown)


source/administration/upgrade.rst
```diff
@@ -54,10 +54,18 @@ Location of your local storage directory
 
      wget https://releases.mattermost.com/X.X.X/mattermost-X.X.X-linux-amd64.tar.gz
 
-#. Extract the Mattermost Server files.
+#. Confirm no other Mattermost zip folders exist in your ``/tmp`` directory. If another version's zip file does exist, delete or rename the file
```
cwarnermm, line 57:
```suggestion
#. Confirm no other Mattermost zip folders exist in your ``/tmp`` directory. If another version's zip file does exist, delete or rename the file.
```

coltoneshaw:
@cwarnermm  - Fixed the numbering issue I was having and removed the Important block since it didn't flow. The `note` block I think can be added/removed based on how you think it reads but the content itself valuable. Also threw some grammar fixes in as well for the whole doc and my updates.

Let me know if you have any feedback.

Screen Shot 2021-06-29 at 10 51 21 AM

cwarnermm (approved):
This is really great, @coltoneshaw! Thank you so much!

PR #4681, opened 2021-06-29
Add note about new mmctl CLI

bbodenmiller:
#### Summary
Note how currently CLI will be replaced with mmctl CLI in the future. This is inline with comments throughout current CLI docs.


cwarnermm (approved):
Thank you!

PR #4743, opened 2021-07-16
Fixing a few broken docs

amyblais:
Also adjusting one of the upgrade notes for 5.37.


source/messaging/managing-channels.rst
```diff
@@ -143,7 +143,10 @@ To unmute the channel, select the channel name again to access the drop-down men
 Switching channels
 ------------------
 
-Use the **Find channel** option in the channel sidebar, or press CMD/CTRL+K, to visit other channels, or to review `member availability <https://docs.mattermost.com/help/getting-started/setting-your-status-availability.html#setting-your-availability>`__ at a glance. 
+Use the **Find channel** option in the channel sidebar, or press CMD/CTRL+K, to visit other channels, or to review `member availability <https://docs.mattermost.com/help/getting-started/setting-your-status-availability.html#setting-your-availability>`__ at a glance.
+
+    .. image:: ../../images/switch-channels.png
```
amyblais, line 148: Not sure if this image was intentionally removed, but it doesn't work in preview.

PR for reference: https://github.com/mattermost/docs/pull/4698/files

cwarnermm: @amyblais Image links need to be updated to ``../images/filename.png`` rather than ``../../images``

cwarnermm, line 148:
```suggestion
    .. image:: ../images/switch-channels.png
```

cwarnermm:
```suggestion
    .. image:: ../images/switch-channels.png
```

cwarnermm: I've made these updates.

source/install/self-managed-changelog.md
```diff
@@ -18,7 +18,7 @@ Lastest Mattermost Releases:
 **Release Day: 2021-07-16**
 
 ### Deprecations
- - The ``platform`` binary and “--platform” flag have been deprecated. If you are using the “--platform” flag or are using the ``platform`` binary directly to run the Mattermost server application via a systemd file or custom script, you will be required to use only the mattermost binary.
+ - The ``platform`` binary and “--platform” flag have been deprecated.
```
streamer45, line 21: I was actually thinking of replacing the word "deprecated" in favor of "removed". The rest of the message is fine. 
I guess the point I'd like to pass is that this is not a simple deprecation warning. That went out at the begin of the year if I remember correctly. Now we acted on the deprecation and completely removed the binary.

cwarnermm:
@amyblais - Users who click on existing docs links with now-outdated URLs will be automatically redirected to the new page. We've been updating the older links as we encounter them, but we included the 301 redirects too to assist.

cwarnermm (approved):
Updated the image file paths

streamer45 (approved):
Awesome, thanks Amy :+1:

PR #4750, opened 2021-07-19
Update getting-started.rst

Sayanta66:
these words not needed on the doc. Keeping the article simple


justinegeffen:
Thanks for this PR, @Sayanta66! The suggested change isn't one we can implement for this document, but we appreciate your input.

PR #4764, opened 2021-07-21
Make format of action ID clear

thierrymarianne:
#### Summary

In accordance with the server-side route matching an action depending on a post,
an action should contain letters or numbers only.

See https://github.com/mattermost/mattermost-server/blob/33edd645c765ffed2951427c0ca0879fe41c4321/api4/integration_action.go#L14

I've added the action ID format next to it.

#### Ticket Link

There is no ticket link

### DIFF
--- a/source/developer/interactive-messages.rst
+++ b/source/developer/interactive-messages.rst
@@ -301,7 +301,7 @@ Context
         }
       }
 
-  Then, when the integration receives the request, it can act based on the action ID.
+  Then, when the integration receives the request, it can act based on the action ID, which should match the `[a-zA-Z0-9]+` regular expression.
 
 Tips and Best Practices
 ------------------------


source/developer/interactive-messages.rst
```diff
@@ -301,7 +301,7 @@ Context
         }
       }
 
-  Then, when the integration receives the request, it can act based on the action ID.
+  Then, when the integration receives the request, it can act based on the action ID, which should match the `[a-zA-Z0-9]+` RegExp pattern.
```
justinegeffen, line 304: Just want to check whether this should be "RegExp" or "RegEx" and whether it's a separate pattern to "regex". Thank you!

justinegeffen (commented):
Thanks so much for this PR! It's a really helpful addition. I've added a comment - let me know if you'd like to chat further.

source/developer/interactive-messages.rst
thierrymarianne: I was thinking of regular expression so regex would just fine. I might have been under [influence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) here :sweat_smile: 
As you prefer :smiley:

cwarnermm: The term "regex" is the one I'm personally more familiar with. However, Google uses the term "RegExp". To avoid unintentional ambiguity, perhaps going with "regular expression" would work? 

@justinegeffen - Thoughts?

thierrymarianne: I've replaced `RegExp pattern` with `regular expression` in https://github.com/mattermost/docs/commit/69b1b923761bd441f5c84aeafe42f2161cd2a5c3. 
Please let me know if you would like to see something else instead.

cwarnermm (approved):
Thank you so much for your docs contribution!

source/developer/interactive-messages.rst
justinegeffen: That works for me, thank you @thierrymarianne! And thanks for clarifying, @cwarnermm.

justinegeffen (approved):

thierrymarianne:
Many thanks 😊

PR #4792, opened 2021-07-27
Update links to the exact section and update broken links aswell

arjitc:
#### Summary

- A few links were broken (ie, redirected back to the home page of the documentation section) 
- A few links redirected to a the relevant page but not to the exact section (ie, if someone wanted to know how about _Integrity and Audit Controls_, the link would go to "security.html" and not "security.html#integrity-and-audit-controls" (I suspect this happened as the pages/path were renamed from deployment to deploy for example which breaks the # value after the redirect happens).

#### Ticket Link

N/A


justinegeffen (commented):
Thank you for this extremely helpful PR, @arjitc! This really makes our lives a lot easier and we appreciate you fixing the links.

justinegeffen (approved):

PR #4795, opened 2021-07-27
Update links and link to the exact section

arjitc:
#### Summary

- A few links redirected to a the relevant page but not to the exact section (ie, if someone wanted to know how about _Integrity and Audit Controls_, the link would go to "security.html" and not "security.html#integrity-and-audit-controls" (I suspect this happened as the pages/path were renamed from deployment to deploy for example which breaks the # value after the redirect happens).

#### Ticket Link

N/A

Similar to PR #4792

### DIFF
--- a/source/about/product.rst
+++ b/source/about/product.rst
@@ -5,7 +5,7 @@ Product Overview
 
 Thousands of organizations use Mattermost around the world in 20 languages for its unmatched benefits:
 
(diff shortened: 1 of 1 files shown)


justinegeffen (approved):

justinegeffen:
Thanks for this PR! We recently updated our site architecture and some of the links inadvertently got deprecated. We really appreciate your work on this. :)

Issue #4712, opened 2021-07-07
Resolve documentation build warnings: Inline emphasis start-string without end-string


cwarnermm:
Mattermost product documentation is written in RST, generated using Sphinx, and formatted using a Sphinx theme called ReadTheDocs.

When the documentation builds, the following common Sphinx build warning is returned:
- ``Inline emphasis start-string without end-string``

The Technical Writing team welcomes some assistance identifying what's causing this warning in the content source files, how to resolve the warning, and how to avoid the warning altogether going forward with new content development.

VS Code-based linters and other RST format checking tools compatible with Sphinx 3.4.3 and later are also most welcome additions to our existing toolset.

haardikdharma10:
Hey @cwarnermm, I am no Sphinx expert but don't know if you have gone through this discussion already, if not, this can be of help:

https://github.com/fish-shell/fish-shell/pull/7520

cwarnermm:
Thanks, @haardikdharma10! We'll check this out!

larkox:
May I take this one?

justinegeffen:
> May I take this one?

Absolutely! :)

larkox:
This was a proper warning that was rendered fine and two false positives due to something that also was rendering fine.

- Proper warning:
  In some place we had something like `*t.Testing`. The `*` is used as the emphasis marker. Since there are no other `*` to close the marker, hence the warning. We can escape it with a backslash. Looking at this, we should check for other * that are not properly escaped, since we may have more problems with those.
- False positive
The code blocks should have the `::` as a single paragraph (i.e. one empty line before and one empty line after). These ones did not have an empty line after. Therefore the parser did not recognize it as code, and started considering the `*` later as emphasis markers.

A PR is coming soon.

larkox:
I correct myself, the `::` problem was causing an improper rendering.

PR #4812, opened 2021-07-30
Fix "Inline emphasis start-string without end-string" warnings on build

larkox:
#### Summary
This was a proper warning that was rendered fine and two false positives that uncovered a more pressing problem.

- Proper warning:
  In some place we had something like `*t.Testing`. The `*` is used as the emphasis marker. Since there are no other `*` to close the marker, hence the warning. We can escape it with a backslash. Looking at this, we should check for other * that are not properly escaped, since we may have more problems with those.
- False positive
The code blocks should have the `::` as a single paragraph (i.e. one empty line before and one empty line after). These ones did not have an empty line after. Therefore the parser did not recognize it as code, and started considering the `*` later as emphasis markers.

#### Ticket Link
Fix https://github.com/mattermost/docs/issues/4712

### DIFF
--- a/source/deploy/mobile-apps-faq.rst
+++ b/source/deploy/mobile-apps-faq.rst
@@ -213,6 +213,7 @@ The app checks for platform-specific configuration on app install. If no configu
 2. In order to handle deep links, paste the following ``JSON`` into the ``apple-app-site-association`` file. Make sure to place your app ID in the ``appID`` property:
 
 ::
+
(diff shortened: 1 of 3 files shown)


larkox:
I also piggybacked one typo correction

justinegeffen:
This is an awesome PR, thank you!! Really great work, and solves a very annoying problem. :)

Issue #3326, opened 2020-01-22
List of Supported Languages for Syntax Highlighting is out of date


christian-steinmeyer:
I have just noticed that not all languages given in the [documentation](https://github.com/mattermost/docs/blob/master/source/help/messaging/formatting-text.rst) do render correctly but some that are not given actually do. I assume there have been some updates in the rogue library that have been integrated but not made their way into the documentation.

For example, when writing:

    ```ps1
    some-powershell-command "some string" #comment
    ```

This should look like
```powershell
some-powershell-command "some string" #comment
```

But actually looks like
```
some-powershell-command "some string" #comment
```
However, using `powershell` instead of `ps1` as language identifier works like a charm.

The current list of identifiers and supported languages by rogue can be found [here](https://github.com/rouge-ruby/rouge/wiki/List-of-supported-languages-and-lexers).

Sidenote: Interestingly, Github supports the union of identifiers for those examples, I've tested.

justinegeffen:
Hi @christian-steinmeyer. Thanks for raising this, it's very helpful. Would you be open to updating the documentation with your findings? If you have any additional documentation examples that would also be great!

christian-steinmeyer:
Hi!

I can certainly update the list in the existing documentation to the one published by rogue. And perhaps add a link to their list for future changes. However, I believe that it would make sense to test or varify that list first, for which I lack the time, unfortunately.

justinegeffen:
Thanks, @christian-steinmeyer that's a good point. I'm cc'ing @lieut-data: is there a way we can verify the languages listed in the "Syntax Highlighting" section? Or is there a list we can link to?

lieut-data:
@justinegeffen, I /think/ we depend on https://highlightjs.org/ for the highlighting. @hmhealey, is that correct?

I wonder if we shouldn't just link out to their documentation, identifying the version we're using for reference?

hmhealey:
We use Highlight JS, but the languages are enabled manually, so we haven't enabled them all. If we were to add something like lazy loading of languages, we could turn them all on and just link out to their docs, but that's not possible currently.

We shouldn't be linking to Rogue's list of languages though since that's not the library that we use.

justinegeffen:
Perhaps we should add a disclaimer such as: We use Highlightjs.org for highlighting. Not all languages are enabled however they can be manually enabled as required."

If that is accurate, could we add some steps for manually enabling them or, if inaccurate, any alternative suggestions would be appreciated.

[...]

PR #4813, opened 2021-07-30
updating supported languages and converting display to a table

ewwollesen:
#### Summary
I updated the list of supported languages based on the list in [constants.jsx](https://github.com/mattermost/mattermost-webapp/blob/master/utils/constants.jsx)  which I believe to be accurate. I converted it from just a list of aliases to a table that now lists all of the available aliases for each supported language. I put the table in a separate file to make the main doc page a little cleaner, and I basically just copied the format from a table in a different page.

The table was maybe out of scope of the linked issue, but it seems a little more useful than the previous list format.

#### Ticket Link
https://github.com/mattermost/docs/issues/3326


justinegeffen (approved):

justinegeffen:
This is a great improvement, thank you! :)

PR #4836, opened 2021-08-02
Fixed deprecated link in audit-log.rst

Rutam21:
#### Summary

Fixed deprecated link in audit-log.rst

#### Ticket Link

NA

### DIFF
--- a/source/comply/audit-log.rst
+++ b/source/comply/audit-log.rst
@@ -12,7 +12,7 @@ Configuring the Mattermost server to use the new audit log requires editing the 
 
 **Accessing configuration options for audit log**
 
-Open ``config.json`` and navigate to the audit settings. The following `configuration options <https://docs.mattermost.com/administration/config-settings.html#audit-settings>`_ are available:
+Open ``config.json`` and navigate to the audit settings. The following `configuration options <https://docs.mattermost.com/configure/configuration-settings.html#audit-settings>`_ are available:
 
 - Syslog configuration options:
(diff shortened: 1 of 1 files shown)


source/comply/audit-log.rst
```diff
@@ -26,7 +26,7 @@ Open ``config.json`` and navigate to the audit settings. The following `configur
 
     - Enable settings to write audit files locally
     - Specify file size
-    - Specify backup interval
+    - Specify the backup interval
     - Add compression
     - Set maximum age to manage file rotation
```
justinegeffen, line 32: This is a helpful change, thank you! I am curious as to why only this point got edited and not points two and five? Looking at this suggestion it seems like the same could apply to those. :)

justinegeffen:
Thank you for this PR! I've added a comment about one of the changes. Look forward to discussing further. :)

source/comply/audit-log.rst
Rutam21: This is because backup intervals can be crucial so we are trying to give it the extra emphasis by supplying the definite article 'THE'. File Size can be of any amount, doesn't necessarily have to be a definite one. But still, there's no hard and first rule, so we can either leave it for the other two or else can add for them as well.

justinegeffen: Thank you - that makes sense. I'm happy with it as-is.

justinegeffen (approved):

PR #4843, opened 2021-08-02
Environment Cleanup (Docathon 2021)

neflyte:
# Mattermost Documentation - Build Environment Cleanup

### Primary
- Lock down versions of Sphinx and its dependencies to mitigate issues due to unexpected updates
- Upgrade Sphinx extensions to recent releases
- Update Sphinx scripts to reflect recent Sphinx releases

### Secondary
- Enable parallel reading and writing support in Sphinx
- Document the `pipenv` dependencies file (`Pipfile`)
- Enable unlimited depth in navigation sidebar
- Fix missing HTML tags in 404 template

#### Removed

- `recommonmark` since it was deprecated
- `sphinx-press-theme` since it was not used

#### Added

- `myst-parser` to replace `recommonmark`
- `typing_extensions` to support `myst-parser` on macOS with Python 3.7

#### Moved

- `sphinx_reredirects` into the `extensions/` directory and enabled parallel read and write support
- `sphinx-sitemap` into the `extensions/` directory and enabled parallel read and write support

### Dependency Versions

Sphinx dependencies were locked down to the following versions:

| Dependency | Version |
| --- | --- | 
| `sphinx` | 3.5.4 |
| `sphinx-rtd-theme` | 0.5.2 |
| `myst-parser` | 0.15.1 |
| `sphinx-autobuild` | 2021.3.14 |
| `typing-extensions` | 3.10.0.0 |
| `awscli` | 1.20.11 |

### Sphinx Configuration

- The `conf.py` file was updated to reflect recent releases of Sphinx
- Configuration for builders other than HTML have been removed
- A minimum Sphinx version of `3.4.3` was set to align with the recent Sphinx upgrade in this repo
- Sidebar navigation depth was set to `-1` to allow for unlimited depth

#### macOS

- Sphinx's parallel read and write support is not available when using Python 3.8+ on macOS
- Parallel read and write support is available when using Python 3.7
  - See [www.python.org](https://www.python.org/) for more information on installing Python
- If you do not need Sphinx parallel read or write support, skip the instructions in this section.
- If you have already run `pipenv sync` in your environment, you will need to remove the virtual environment that was generated
  - Note that some [VSCode](https://code.visualstudio.com/) extensions may be impacted by recreating the virtual environment
    - VSCode should prompt you to re-install any missing extensions and their components
  - Remove the existing virtual environment by running the following command:
    - `pipenv --rm`
- Create a new `pipenv` virtual environment with Python 3.7 using the following command:
  - `pipenv --python 3.7 sync`
- The documentation can now be built using the following command:
  - `make html`

### Sphinx Autobuild

The `sphinx-autobuild` extension is enabled in the Sphinx Makefile and make.bat scripts. This extension allows rapid
development of documentation by watching for modified files and immediately re-building them. The extension also starts
a local HTTP server to host the built documentation.

To start the autobuild, run the following command:
```shell
make livehtml
```

When you have finished using autobuild, press `CTRL-C` to stop it.

For more information, see the sphinx-autobuild repo: [sphinx-autobuild](https://github.com/executablebooks/sphinx-autobuild)

### Missing HTML tags in 404 template

The 404 template was updated to contain complete HTML tags and a defined document language.


cwarnermm:
@justinegeffen - Full disclosure: This PR is a contribution from my partner, Alan Lew. We teamed up together to tackle some of our tool stack cleanup work. In addition to Alan's excellent PR notes, I've noted the following during our testing of these PR changes:

MAC instructions ONLY; can't be used for Windows or Linux

Alan's available to walk us through any part of this PR to clarify as needed.

Switched from Python 3.9 to Python 3.7
We currently have Python 3.9 installed and working; however, we need Python 3.7 to take advantage of performance improvements during docs builds.
run ``brew list``
- lists all packages that Brew has installed on the local machine
- if you see python@3.9, that's the version you have installed. You'll be installing another parallel version of Python, rather than downgrading the one you have.
- Note: You can have multiple versions of Python installed on a machine.

Install Python 3.7
run ``brew install python@3.7``
- installs 3.7 on the local machine
- you may see updates come through for 3.9 too and pip env, that's okay
- Note: the following error may occur: "Error: The 'brew link' step did not complete successfully"; this error is unrelated and shouldn't affect anything we're trying to do
- Python 3.8 and 3.9 don't work with the parallel building performance improvements; there was a change in how Python does parallel processing on Mac that Sphinx isn't compatible with.
    - If this gets resolved at some point, we can use a later version of Python again, but in the meantime, the performance enhancements are worth using Python 3.7
    - not only will we see local builds generating faster, but all of our our pipeline build processes will also be much faster, and our previews will be available sooner

Switch to the branch where you want to build the docs, and rebuild your pipenv environment
run ``pipenv --rm``
- if there wasn't already one there, that's okay; the command will be aborted.

Recreate a new pipenv environment
run ``pipenv --python 3.7 sync``
    - this didn't work the first time on Carrie's machine because Python 3.7 wasn't in path environment variable
    - use any text editor to add ``python@3.7`` to the export PATH
    - Not all writers may be using bash; different shells will require different steps to correct path environment variables

Build the docs locally
run ``make html``
- with any luck, the docs build will look and feel A LOT FASTER because Alan has updated our tech stack to take advantage of all cores on a given machine when building
- all warnings and errors are now piped into files ``warnings.log`` and ``errors.log`` located in the ``build`` directory locally (Sphinx seems to treat errors and warnings the same way; we may see duplicate warnings across the two files)
    - they're no longer displayed on the screen; and they no longer roll off the screen!
    - running a ``make clean html`` will clear these log files as well as remove all build artifacts (really handy when switching branches, etc.)
    - if a build fails, review the contents of these log files for clues as to what has happened.

In docs builds
- LHS depth has been configured as ``-1`` (was ``4``) to enable unlimited depth (goes as far down as our topic hierarchy does)
- highlighting appears to be improved down to at least H4 if not lower
- some ways to make this better:
    - a LHS that is resizable on demand (by the user); requires some CSS and potentially browser cookie code to remember width
    - some of our topics, when selected in the LHS, auto-collapse; this may be a formatting issue on our end causing this
    - the LHS tends to scroll up beyond what's visible on the screen as you select deeper and deeper. There may be ways to address this.
- If this LHS depth isn't great: revert the depth back

New markdown parser
- updated to a new implementation (myst-parser) since the old one we were using has since been deprecated (recommonmark)
    - different parser altogether
    - we should review our MD content to ensure it's being converted as expected
- the myst-parser is relatively new technology, from the makers of recommonmark; it was released recently
    - with this new parser, we can use directives with MD files in addition to RST files
    - Alan has added doc links in the ``Pipfile`` for each of our active extensions so that we can learn more about how each of them work

Dependency versions are now locked down via the ``Pipfile``
- our Sphinx version is no longer 3.4.3; we're using 3.5.4 (which is pretty close to bleeding edge)
    - that jump from 3.4.3 to 3.5.4 happened because we didn't lock our dependencies down
- we always know what versions of our dependencies we're getting and using
- we can upgrade when we're ready, as opposed to brew or other tooling upgrading us without us knowing or preparing for it
- we're saying we want version X because we know version X works for our purposes
    - when we're ready to upgrade to Sphinx 4, we can plan for this in test environments to know what's going to work, what's going to break, etc. That will also work for all of our depencies, including parsers, linters, etc.

index.rst changes
- 3 links have been removed because they didn't appear to be used/doing anything; we'll want to confirm

justinegeffen:
Wow, @cwarnermm and @neflyte! This is an astounding PR. Thanks to both of you for putting this together. I'm going to leave this for review over the next day or so if that works for you. I'd like to take some time with this. But really really awesome work. :)

neflyte:
I've updated the `sphinx_sitemap` extension to enable proper parallel write support based on an existing GitHub PR: https://github.com/jdillard/sphinx-sitemap/pull/29

This resolves the mostly empty `sitemap.xml` file that was produced previously.

extensions/sitemap/__init__.py
```diff
@@ -0,0 +1,222 @@
...
+def setup(app):
+    """Setup connects events to the sitemap builder"""
+    app.add_config_value(
+        'site_url',
+        default=None,
+        rebuild=''
+    )
+    app.add_config_value(
+        'sitemap_url_scheme',
+        default="{lang}{version}{link}",
```
cwarnermm, line 62: @justinegeffen - This is where we can customize the content elements included in the generated ``sitemap.xml`` file. Removing ``{verison}`` here removes the ``version`` value set via the ``conf.py`` file.

cwarnermm:
@justinegeffen - With @neflyte's latest commit, sitemap generation looks restored. 
With the change noted [here](https://github.com/mattermost/docs/pull/4843/files#r687263998), the following ``conf.py`` code can safely be merged into Production:

```
# The short X.Y version.
version = '5.37'
# The full version, including alpha/beta/rc tags.
release = '5.37'
```

justinegeffen:
Whoohoo! That's awesome news! I'll review today and we can merge later today. :) Awesome job, @neflyte!!

PR #4854, opened 2021-08-03
Update config-tls-mattermost.rst

Sayanta66:
Updated a link and added a resource


source/install/config-tls-mattermost.rst
```diff
@@ -62,3 +62,4 @@ Other helpful resources:
 
 - `NGINX's SSL blog <https://www.nginx.com/blog/nginx-ssl/>`__
 - `NGINX's SSL guide <https://docs.nginx.com/nginx/admin-guide/security-controls/terminating-ssl-http/>`__
+- `All about Nginx’s SSL <http://nginx.org/en/docs/http/ngx_http_ssl_module.html>`__
```
cwarnermm, line 65: I'm inclined to leave this link out of the official product documentation given that it's not a secure site using HTTPS. @justinegeffen - Thoughts?

cwarnermm (commented):
Thank you for your docs contribution! I appreciate you adding an additional resource link for readers; however, it may pose a security risk to visitors.

source/install/config-tls-mattermost.rst
justinegeffen: Definitely agree with that.

PR #4808, opened 2021-07-29
Updated links and Fixed links

Sayanta66:
Updated the links and fixed the links in the exact location

PR #4859, opened 2021-08-03
Update slash-commands.rst

Sayanta66:
Updated a link. In reference to PR #4808


Sayanta66:
Okay I will update the new one.

PR #4896, opened 2021-08-04
Updated deprecated links in atlassian-integrations.rst

Rutam21:
#### Summary

Updated deprecated links in atlassian-integrations.rst

#### Ticket Link

NA


justinegeffen:
Thank you for your PR. This content is no longer in use which is why it's in the archive. We don't currently require changes to it. :)

PR #4899, opened 2021-08-04
Updated the deprecated links in bots.rst

Rutam21:
#### Summary

Updated the deprecated links in bots.rst

#### Ticket Link

NA


justinegeffen:
Thank you for this PR. The content is archived so we don't currently need it to be updated. Thank you!

Issue #4101, opened 2020-11-10
Document how to work with SELinux in enforcing mode


nigelhorne:
This line is missing from [https://docs.mattermost.com/install/install-rhel-8.html](url):

`sudo restorecon -R /opt/mattermost/`

It should come after the line saying

`sudo chmod -R g+w /opt/mattermost`

justinegeffen:
Thank you for letting us know, @nigelhorne! Would you be open to updating the doc and opening a PR for it?

PR #4925, opened 2021-08-04
Add sudo restorecon -R /opt/mattermost/ command line to the documenta…

Francois-D:
#### Summary
Add sudo restorecon -R /opt/mattermost/ command line to the documentation.  
Restorecon command will reset the SELinux security context for files and directories to its default values. 

#### Ticket Link
Fixes https://github.com/mattermost/docs/issues/4101


jespino (approved):
Technically looks good to me. But I'm not a SELinux user, so maybe we should include somebody else here. The reasoning behind the command makes sense.

jespino:
@DSchalla @jupenur Any of you knows better if this command is right in this context? I think it is, but I don't have the environment to test it and I'm not very familiarized with SELinux enough.

justinegeffen (approved):
Great contribution, thank you!

jupenur:
> @DSchalla @jupenur Any of you knows better if this command is right in this context? I think it is, but I don't have the environment to test it and I'm not very familiarized with SELinux enough.

Sorry, I have no idea. This is a bit too far outside my area of expertise. Maybe Daniel can pitch in though.

DSchalla (commented):
Generally the command makes sense in that context, the thing I am uncertain about is whether there are any other bits that may need adaption for SELinux - For example, if we add it here we may need to make a note in our upgrade guide as well to do a restorecon to fix the SELinux context. 

If we don't we would be inconsistent at least.

Francois-D:
> Generally the command makes sense in that context, the thing I am uncertain about is whether there are any other bits that may need adaption for SELinux - For example, if we add it here we may need to make a note in our upgrade guide as well to do a restorecon to fix the SELinux context.
> 
> If we don't we would be inconsistent at least.

Running restorecon would reset the files to the default setting.  If the user had made customized changes you would overwrite these changes possibly leading to a unexpected upgrade experience.  

Perhaps recommending to  run restorecon -n option to check the current settings to ensure that the settings are what is desired.   This would be added after point 12 in the document below.

https://docs.mattermost.com/upgrade/upgrading-mattermost-server.html#id3.

DSchalla:
Right, my proposal was to mention during the server upgrade that another `restorecon` would be necessary.

cwarnermm (approved):
Thank you for this docs contribution!

Francois-D:
> Right, my proposal was to mention during the server upgrade that another `restorecon` would be necessary.

"might" be necessary or is necessary?  Sorry I just want to clear on your point.  Thank you.

cwarnermm:
I've logged https://mattermost.atlassian.net/browse/MM-37745 to capture the follow-on work needed for the Upgrade documentation. Thanks @DSchalla and @Francois-D!

PR #4943, opened 2021-08-10
Fix Upgrade docs content issues

cwarnermm:
Clarifying ESR upgrade path
Ensured procedural steps are logical, ordered correctly, and don't contain duplicate/contradictory details


cwarnermm:
I'm in the process of confirming our ESR upgrade path support with @wiersgallak. The Upgrading to the Latest Version section will be updated accordingly. Let's start the technical reviews of the content in parallel. Thank you!

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -12,6 +12,8 @@ Upgrading to the Latest Version
 
 If you are upgrading from version 3.0 or later, these instructions apply to you. If you are upgrading from a version prior to 3.0.0, you must first upgrade to version 3.0.3.
 
+We only support one minor version difference between the server versions when performing a rolling upgrade (for example v5.27.1 + v5.27.2 or v5.26.4 + v5.27.1 is supported, whereas v5.25.5 + v5.27.0 is not supported). Running two different versions of Mattermost in your cluster should not be done outside of an upgrade scenario.
```
cwarnermm, line 15: @wiersgallak - This is the information we need to confirm as correct or update if needed. Is this still true? A customer has asked for clarification: "it's not clear for me whether I can upgrade from the 5.31 ESR to the 5.37 ESR, skipping everything in between or not.  The note at the end mentions minor versions and rolling releases - as opposed to what?"

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -24,9 +26,8 @@ Read these instructions carefully from start to finish. Make sure that you under
...
-        - If that command does not produce any results because your version is older, try ``whereis platform`` instead.
-      - Location of your local storage directory: The local storage directory contains all the files that users have attached to their messages. If you don't know its location, open the System Console and go to **Environment > File Storage**, then read the value in **Local Storage Directory**. Relative paths are relative to the ``mattermost`` directory. For example, if the local storage directory is ``./data/`` then the absolute path is ``{install-path}/mattermost/data``.
+      - **Existing install directory - {install-path}**: If you don't know where Mattermost Server is installed, use the ``whereis mattermost`` command. The output should be similar to ``/opt/mattermost/bin/mattermost``. The install directory is everything before the first occurrence of the string ``/mattermost``. In this example, the ``{install-path}`` is ``/opt``. If that command does not produce any results because your version is older, try ``whereis platform`` instead.
```
justinegeffen, line 29:
```suggestion
      - **Existing install directory - {install-path}**: If you don't know where Mattermost Server is installed, use the ``whereis mattermost`` command. The output should be similar to ``/opt/mattermost/bin/mattermost``. The install directory is everything before the first occurrence of the string ``/mattermost``. In this example, the ``{install-path}`` is ``/opt``. If that command does not produce any results it's likely because your version is older; try ``whereis platform`` instead.
```

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +82,81 @@ Upgrading Mattermost Server
...
-  **A default Mattermost installation has the following files and directories**:
-
-    .. code-block:: sh
-
-      $ ls /opt/mattermost
-      ENTERPRISE-EDITION-LICENSE.txt README.md  client  data   i18n  manifest.txt  prepackaged_plugins
-      NOTICE.txt                      bin        config  fonts  logs  plugins       templates
+7. Remove all files **except** data and custom directories from within the current ``mattermost`` directory. 
 
+  **What's Preserved?**
```
justinegeffen, line 87:
```suggestion
  **What's preserved?**
```

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +82,81 @@ Upgrading Mattermost Server
...
-8. Identify if any custom directories from the above step need to be preserved. For each custom directory within the Mattermost folder that you wish to preserve, ensure you add ``-o -path  mattermost/yourFolderHere`` to the following command. See the example below where the folder ``yourFolderHere`` is preserved by adding ``-o -path  mattermost/yourFolderHere``.
+  .. code-block:: sh
 
-.. code-block:: sh
+    $ ls /opt/mattermost
+    ENTERPRISE-EDITION-LICENSE.txt README.md  client  data   i18n  manifest.txt  prepackaged_plugins
+    NOTICE.txt                      bin        config  fonts  logs  plugins       templates
 
-  sudo find mattermost/ mattermost/client/ -mindepth 1 -maxdepth 1 \! \( -type d \( -path mattermost/client -o -path mattermost/client/plugins -o -path mattermost/config -o -path mattermost/logs -o -path mattermost/plugins -o -path mattermost/data -o -path  mattermost/yourFolderHere \) -prune \) | sort | xargs echo rm -r
+  **Clear Mattermost Contents**
```
justinegeffen, line 101:
```suggestion
  **Clear Mattermost contents**
```

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +82,81 @@ Upgrading Mattermost Server
...
 
-9. You should first modify the last part to ``xargs echo rm -r`` to verify what will be executed. If you've added custom directories to the command in step b, then add those to the following command. For example:
-    
-.. code-block:: sh
+  The following command clears the contents of mattermost, preserving only those directories and their contents. You should first modify the last part to ``xargs echo rm -r`` to verify what will be executed.
```
justinegeffen, line 103:
```suggestion
  The following command clears the contents of the ``mattermost`` folder, preserving only the specified directories and their contents. You should first modify the last part to ``xargs echo rm -r`` to verify what will be executed.
```

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +82,81 @@ Upgrading Mattermost Server
...
 
-  sudo find mattermost/ mattermost/client/ -mindepth 1 -maxdepth 1 \! \( -type d \( -path mattermost/client -o -path mattermost/client/plugins -o -path mattermost/config -o -path mattermost/logs -o -path mattermost/plugins -o -path mattermost/data \) -prune \) | sort | xargs echo rm -r
+  If you store TLSCert/TLSKey files or other information within your ``/opt/mattermost`` folder, you should add ``-o -path mattermost/yourFolderHere`` to the below command or you will have to manually copy the TLSCert/TLSKey files from the backup into the new install.
```
justinegeffen, line 105:
```suggestion
  If you store TLSCert/TLSKey files or other information within your ``/opt/mattermost`` folder, you should add ``-o -path mattermost/yourFolderHere`` to the below command or you'll have to manually copy the TLSCert/TLSKey files from the backup into the new install.
```

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +82,81 @@ Upgrading Mattermost Server
...
-10. Clear the contents of this directory. If you've added custom directories to the command, be sure to add those to the following command. For example:
+  .. code-block:: sh
 
-.. code-block:: sh
+    sudo find mattermost/ mattermost/client/ -mindepth 1 -maxdepth 1 \! \( -type d \( -path mattermost/client -o -path mattermost/client/plugins -o -path mattermost/config -o -path mattermost/logs -o -path mattermost/plugins -o -path mattermost/data \) -prune \) | sort | sudo xargs rm -r  
+  
+  **Using Bleve Search**
 
-  sudo find mattermost/ mattermost/client/ -mindepth 1 -maxdepth 1 \! \( -type d \( -path mattermost/client -o -path mattermost/client/plugins -o -path mattermost/config -o -path mattermost/logs -o -path mattermost/plugins -o -path mattermost/data \) -prune \) | sort | sudo xargs rm -r
+  If using `Bleve Search <https://docs.mattermost.com/deploy/bleve-search.html>`__, the index directory path won't be preserved with the commands below if the directory exists *within* the ``mattermost`` directory.
```
justinegeffen, line 113:
```suggestion
  If using `Bleve Search <https://docs.mattermost.com/deploy/bleve-search.html>`__, and the directory exists *within* the ``mattermost`` directory, the index directory path won't be preserved with the commands below. 
```

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +82,81 @@ Upgrading Mattermost Server
...
+11. Start your Mattermost server.
 
-.. code-block:: sh
+  .. code-block:: sh
 
-  sudo systemctl start mattermost
+    sudo systemctl start mattermost
 
-15. If you're using a High Availability deployment you need to apply the steps above on all the nodes in your cluster. Once complete, the **Config File MD5** columns in the High Availability section of the system console should be green. If they're yellow, please ensure that all nodes have the same server version and the same configuration.
+12. If you're using a High Availability deployment, you need to apply the steps above on every nodes in your cluster. Once complete, the **Config File MD5** columns in the High Availability section of the system console should be green. If they're yellow, please ensure that all nodes have the same server version and the same configuration.
```
justinegeffen, line 154:
```suggestion
12. If you're using a High Availability deployment, you need to apply the steps above on every node in your cluster. Once complete, the **Config File MD5** columns in the High Availability section of the System Console should be green. If they're yellow, please ensure that all nodes have the same server version and the same configuration.
```

source/upgrade/upgrading-mattermost-server.rst
mrckndt, line 29: `whereis` only looks up standard binary places and $PATH. If e.g. **/opt/mattermost/bin** wasn't added to the PATH `whereis` will return nothing. To find the *mattermost* binary something like `find / -executable -type f -iname mattermost 2> /dev/null` would be more generic/useful.

justinegeffen (commented):
Great update! I've added some comments and minor suggestions. :)

source/upgrade/upgrading-mattermost-server.rst
cwarnermm: This is great feedback, @mrckndt! Thank you!

source/upgrade/upgrading-mattermost-server.rst
cwarnermm: @wiersgallak I've clarified all upgrade path details under "Upgrading to the Latest Version".

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +84,81 @@ Upgrading Mattermost Server
...
+   sudo rm -r /tmp/mattermost-upgrade/
+   sudo rm -i /tmp/mattermost*.gz
 
-.. note::
-  The ``n`` (no-clobber) flag and trailing ``.`` on source are very important.
+  .. note::
+    The ``n`` (no-clobber) flag and trailing ``.`` on source are very important.
 
-13. If you want to use port 80 to serve your server, or if you have TLS set up on your Mattermost server, you *must* activate the CAP_NET_BIND_SERVICE capability to allow the new Mattermost binary to bind to low ports. For example:
+10. If you want to use port 80 to serve your server, or if you have TLS set up on your Mattermost server, you **must** activate the CAP_NET_BIND_SERVICE capability to allow the new Mattermost binary to bind to low ports. For example:
```
mrckndt, line 143: Maybe add an explicit note on port 443 (instead of or in addition to TLS) as this is a common question which arises sometimes and not everyone knows the terminology/tech behind this. Also low ports isn't really specific and could be changed to *ports lower than 1024*. This leaves less open questions to the users imo.

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +84,81 @@ Upgrading Mattermost Server
...
+  If using `Bleve Search <https://docs.mattermost.com/deploy/bleve-search.html>`__, and the directory exists *within* the ``mattermost`` directory, the index directory path won't be preserved with the commands below. 
+  
+  - You can either move the bleve index directory out from the ``mattermost`` directory before upgrading or, following an upgrade, you can copy the contents of the bleve index directory from the ``backup`` directory. 
+  - You can then store that directory or re-index as preferred. 
+  - The bleve indexes can be migrated without reindexing between Mattermost versions. See our `Configuration Settings <https://docs.mattermost.com/configure/configuration-settings.html#bleve-settings-experimental>`__ documentation for details on setting the bleve index directory.
       
-11. Change ownership of the new files before copying them. For example:
+8. Change ownership of the new files before copying them. For example:
```
mrckndt, line 121: I would put this before point 11 (right before starting the server). In case the user managed to change the permissions while doing the upgrade (e.g. by copying files/folders with sudo) it's not enough to change the owner and group of just the install files.

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +84,81 @@ Upgrading Mattermost Server
...
 
-.. code-block:: sh
+  .. code-block:: sh
          
-  sudo chown -hR mattermost:mattermost /tmp/mattermost-upgrade/
+    sudo chown -hR mattermost:mattermost /tmp/mattermost-upgrade/
```
mrckndt, line 126: In case this moves down: `sudo chown -R mattermost:mattermost {install-path}`. The `-h` option can be omitted and probably introduces more issues than it solves.

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +84,81 @@ Upgrading Mattermost Server
...
-  sudo rm -r /tmp/mattermost-upgrade/
-  sudo rm -i /tmp/mattermost*.gz
+   sudo cp -an /tmp/mattermost-upgrade/. mattermost/
+   sudo rm -r /tmp/mattermost-upgrade/
+   sudo rm -i /tmp/mattermost*.gz
 
-.. note::
-  The ``n`` (no-clobber) flag and trailing ``.`` on source are very important.
+  .. note::
+    The ``n`` (no-clobber) flag and trailing ``.`` on source are very important.
```
mrckndt, line 142: *The ``n`` (no-clobber) flag preserves existing configs and logs in your installation path and the trailing ``.`` on source ensures all install files are copied.*? - This makes it more clear and less opaque.

source/upgrade/upgrading-mattermost-server.rst
```diff
@@ -81,87 +84,81 @@ Upgrading Mattermost Server
...
+    ENTERPRISE-EDITION-LICENSE.txt README.md  client  data   i18n  manifest.txt  prepackaged_plugins
+    NOTICE.txt                      bin        config  fonts  logs  plugins       templates
 
-  sudo find mattermost/ mattermost/client/ -mindepth 1 -maxdepth 1 \! \( -type d \( -path mattermost/client -o -path mattermost/client/plugins -o -path mattermost/config -o -path mattermost/logs -o -path mattermost/plugins -o -path mattermost/data -o -path  mattermost/yourFolderHere \) -prune \) | sort | xargs echo rm -r
+  **Clear Mattermost contents**
 
-9. You should first modify the last part to ``xargs echo rm -r`` to verify what will be executed. If you've added custom directories to the command in step b, then add those to the following command. For example:
-    
-.. code-block:: sh
+  The following command clears the contents of the ``mattermost`` folder, preserving only the specified directories and their contents. You should first modify the last part to ``xargs echo rm -r`` to verify what will be executed.
```
mrckndt, line 105: I think it's a good idea to change this to a 'test' run with `sudo find mattermost/ mattermost/client/ -mindepth 1 -maxdepth 1 \! \( -type d \( -path mattermost/client -o -path mattermost/client/plugins -o -path mattermost/config -o -path mattermost/logs -o -path mattermost/plugins -o -path mattermost/data \) -prune \)` instead of the remove command without sudo. In case a user switched to **root** or **mattermost** and forgot about it, `xargs echo rm -r` will remove the data anyway.

mrckndt (commented):
Looks good to me and changes are minor improvements

cwarnermm:
Your review feedback and suggestions for improvements are fantastic, @mrckndt! Thank you so much! I believe I've incorporated all of your feedback. Can you please review the latest end-to-end?

cwarnermm:
@svelle - @mrckndt has done an outstanding job reviewing and revising these upgrade details. Can you please review end-to-end to ensure technical completeness and accuracy?

svelle (approved):
Looks good to me, thanks @cwarnermm and @mrckndt !

wiersgallak (approved):

PR #4945, opened 2021-08-10

amyblais:
Add a link to the v6.0 upgrade note forum post.


source/upgrade/important-upgrade-notes.rst
```diff
@@ -4,7 +4,7 @@ Important Upgrade Notes
...
    - In the v5.38 release (August 16, 2021), we will deprecate "config watcher" (the mechanism that automatically reloads the ``config.json file``), in favor of an mmctl command that will need to be run to apply configuration changes after they are made. This change will improve configuration performance and robustness.
-   - The deprecations `listed here <https://docs.mattermost.com/administration/changelog.html#upcoming-deprecations-in-mattermost-v6-0>`_ are planned for the Mattermost v6.0 release, which is scheduled for September 15, 2021. This list is subject to change prior to the release.
+   - The deprecations `listed here <https://docs.mattermost.com/administration/changelog.html#upcoming-deprecations-in-mattermost-v6-0>`_ are planned for the Mattermost v6.0 release, which is scheduled for September 15, 2021. This list is subject to change prior to the release. Also please note that with Mattermost v6.0, nodes from different versions in a cluster can no longer be run and longer migration times can be expected - see this `Forum post <https://forum.mattermost.org/t/6-0-important-upgrade-notes/12037>`_ for details.
```
streamer45, line 7: @agnivade

My understanding is that this limitation only applies to the 5.x --> 6.x migration process. After that we are back to normal (unless explicitly specified). Just want to make sure there is no room for interpretation here.

agnivade: That is correct.
- Nodes from 5.x cannot run with 6.x nodes in a cluster. For an admin, it effectively means an HA upgrade is not possible when upgrading to 6.0.
- After that, we are back to normal - _unless explicitly specified in the upgrade notes_.

wiersgallak (approved):

streamer45 (approved):

agnivade (approved):

PR #4955, opened 2021-08-12
Fixed a few broken links

emilyacook:
I stumbled upon a few broken links in the Messaging Attachment page!

I'm pretty sure to fix the links it just needs `/help/` removed, but I'm not 100% sure how to check it

https://docs.mattermost.com/developer/message-attachments.html

### DIFF
--- a/source/developer/message-attachments.rst
+++ b/source/developer/message-attachments.rst
@@ -1,7 +1,7 @@
 Message Attachments
 ===================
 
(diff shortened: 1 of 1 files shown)


justinegeffen (approved):

justinegeffen:
Thanks, Emily! These pages are in the process of being deprecated and I'll be putting 301 redirects in place for them today. Thanks for picking this up though, as it means the pages that have been moved likely need an update. :)

PR #4956, opened 2021-08-13
Update conf.py (added 301 redirects)

justinegeffen:
Added 301 redirects for developer pages that have been moved.


cwarnermm (approved):
Thanks!

PR #4969, opened 2021-08-16
Fix mentions of `mmctl config reload`

elyscape:
#### Summary
The release and upgrade notes for 5.38.0 mention that the config watcher has been removed, and that in its place a command has to be run to notify the server that the configuration has changed. This change includes the actual command that has to be run.

#### Ticket Link
N/A


amyblais:
@elyscape Would you be open to updating the same text at https://docs.mattermost.com/upgrade/important-upgrade-notes.html? I can also push the update if it's easier!

elyscape:
Sure. I'll make it happen shortly.

elyscape:
Updated!

amyblais (approved):
Thank you!

PR #4987, opened 2021-08-23
New Deployment Landing Page

BenLloydPearson:
#### Summary
This turns all of the links from the Install, Deploy, Upgrade, and Scale toctree into a structured guide. It also creates a redirect from the toctree page to point here.

#### Ticket Link
N/A

### DIFF
--- a/source/guides/deployment.rst
+++ b/source/guides/deployment.rst
@@ -1,17 +1,14 @@
 :orphan:
 
 Deploy Mattermost
-===================
+=================
 
 Learn how to install, deploy, and scale Mattermost for teams and organizations of any size.
-
 
 Get Started
 -----------
 
 These guides will get you up and running with Mattermost in minutes.
-
-
 
 * :doc:`Run Mattermost via Docker </install/setting-up-local-machine-using-docker>` - Launch a Mattermost server instantly to test functionality and build integrations.
 * :doc:`Quick Install Guide </getting-started/light-install>` - Deploy in minutes via Mattermost Omnibus on Ubuntu.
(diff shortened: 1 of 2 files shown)


source/guides/deployment.rst
```diff
@@ -0,0 +1,162 @@
...
+**Server Installation**
+
+
+
+* :doc:`Software and Hardware Requirements </install/software-hardware-requirements>`
+* :doc:`Local Docker Setup </install/setting-up-local-machine-using-docker>`
+* :doc:`Mattermost Omnibus </install/installing-mattermost-omnibus>`
+* :doc:`Ubuntu 20.04 LTS </install/installing-ubuntu-2004-LTS>`
+* :doc:`Kubernetes </install/install-kubernetes>`
+* `Helm <https://docs.mattermost.com/install/install-kubernetes.html#installing-the-operators-via-helm>`__
```
BenLloydPearson, line 39: I couldn't figure out how to create this anchor link without doing an external link. I'm open to alternatives if someone knows how to properly do this.

cwarnermm: This feels like a reasonable approach. Any concerns from your end, @justinegeffen?

cwarnermm:
@BenLloydPearson @justinegeffen - This is a change we'll need to confirm working in Production. Previews won't be able to show us that a 301 redirect to a new page is working because that new page isn't available in Production yet.

cwarnermm (approved):
This is a good update, Ben! Thank you so much!

source/guides/deployment.rst
justinegeffen: No concerns on my end - I know we can use a :ref: [as mentioned here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#ref-role) and we might want to look into that going forward but I don't think it's a blocker. :)

BenLloydPearson: I tried every combination of :ref: I could think of and nothing seemed to work properly. I have a suspicion that it could be related to how the target doc page is generated (it's 4 docs pages in one), but I wasn't able to diagnose the problem.

justinegeffen: Ah ok that makes sense. Thanks Ben! :)

justinegeffen:
@cwarnermm, would the orphan tag on the page create any build issues?

cwarnermm:
@justinegeffen I believe the orphan directive lets a document be built, but doesn't notify Sphinx that it isn't reachable via a toctree. All source files need to be part of some toctree, else Sphinx will spit out a warning on build because it means that the file in question won't be reachable via standard LHS navigation. 

We should use the orphan directive on any source page we intentionally don't want to show up in the LHS (such as install subsections of content that we reuse across topics), but should not be used on any page the user would expect to get to via LHS navigation.

justinegeffen (approved):

PR #4999, opened 2021-08-25
Unintentional no-op in rm command after improvement

Roy-Orbison:
##### Fixes a bug that could result in no upgrade being performed at all.

The existing example was already a dry-run/no-op. After 09c34e28ac5a1c88ca6b9243830df60432ec4058, the actual deletion was moved to a separate command, but the `echo` was not removed from it, so it does nothing. This will result in no upgrade being performed at all because the later copy command uses the no-clobber flag, therefore will not overwrite the old files at all.


Roy-Orbison:
@cwarnermm @justinegeffen There's a bug in #4943.

Roy-Orbison:
This was ignored and fixed again in a later PR #5007.

PR #5020, opened 2021-09-01
Desktop 4.7.2 docs

amyblais:


source/install/desktop-app-changelog.rst
```diff
@@ -15,6 +15,11 @@ Release v4.7
 
 **Download Binaries:** `Mattermost Desktop on GitHub <https://github.com/mattermost/desktop/releases/latest>`_
 
+- **v4.7.2, released 2021-09-02**
+ - Upgraded to Electron v12.0.16 to fix an issue where grabbing the horizontal scrollbar dragged the app window.
```
devinbinnie, line 19: So this issue came back...I have no idea why.
I think we live with it at this point because it seems to be intermittent, and going back to Electron v12.0.1 causes other issues.

But I think we remove it from the release note to not draw attention to it.

devinbinnie (changes requested):

source/install/desktop-app-changelog.rst
amyblais: Thanks, I left the Electron upgrade but removed the bug fix.

devinbinnie (approved):

PR #5022, opened 2021-09-02
Added link to example webhook payloads

icelander:
#### Summary

Because many devs can't read Go code, I recorded some example webhook requests and posted them to Gist so people can more easily write integrations


justinegeffen (approved):
This is awesome, thanks so much @icelander!

agarciamontoro (approved):
Oh, great addition, @icelander, thank you! :)

PR #5027, opened 2021-09-07
Update playbooks.rst

justinegeffen:
- Changing headers to align with Messaging and Boards sections (i.e., "working with", "accessing" etc).
- Revised the actions section to remove laundry list of actions and include only information that's not immediately accessible in the UI.


source/guides/playbooks.rst
```diff
@@ -12,24 +12,24 @@ Before You Begin
    /playbooks/overview.rst
    /playbooks/getting-started.rst
    
-Planning
---------
+Work with Playbooks
```
cwarnermm, line 15: Work or Working?

source/guides/playbooks.rst
```diff
@@ -12,24 +12,24 @@ Before You Begin
...
+-------------------
 
 .. toctree::
    :maxdepth: 2
 
    /playbooks/setting-up-playbooks.rst
 
-Launching
----------
+Manage Runs
```
cwarnermm, line 23: Manage or Managing?

cwarnermm (approved):
Thanks! Non-blocking questions provided inline.

source/guides/playbooks.rst
justinegeffen: Work, to match Messaging although I am open to "Working" as well if updating Messaging is an option?

source/guides/playbooks.rst
justinegeffen: See above. :)

source/guides/playbooks.rst
cwarnermm: Pre v6.0 is a great time for you and I to align on the use of gerunds going forward in titles for all new content development. 

Do we want to define gerund use per title type (H1 through H4)? Or do we want to use one versus the other consistently for all title types? I'm open to either option? I'm strongly leaning towards simplicity and applying either gerunds or no gerunds across the doc set going forward. Thoughts?

justinegeffen: I'm 5/5 on consistency. We've always leaned toward gerunds and I think it'd make life easier if we used them throughout as opposed to one standard for H1 and a different standard for H2 onwards.

It would be less work to change these to gerunds than the opposite from what I can see from a brief glance through the docs. That said, not all our titles lend themselves to gerunds so I guess the rule should be "when a title has an action, use a gerund" so that pages like "Release notes" won't have to be changed to "Releasing notes". ;)

source/playbooks/setting-up-playbooks.rst
```diff
@@ -42,39 +42,23 @@ Defining actions
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
- * **Send a welcome message**: Enable the toggle and enter text that new channel members will see when they join the channel.
- * **Add the channel to a sidebar category**: Enable the toggle so that the channel will be automatically categorized and listed under **Playbook runs** in the left-hand sidebar.
- 
-When a run is archived
-~~~~~~~~~~~~~~~~~~~~~~
-
- * **Export channel**: (E20 and Cloud only) When a run finishes, this triggers the playbook to automatically save a copy of all channel messages in CSV format. This provides a record of the discussion for future reference or other use such as audits.
+- For information about the webhook payload for run start , see the `PlaybookRunWebhookPayload struct <https://github.com/mattermost/mattermost-plugin-playbooks/blob/b4c8058d8660efe35050bc7eb080e3819c7ab09c/server/app/playbook_run_service.go#L176-L185>`_. An example of the JSON payload for a run start `is available here <https://gist.github.com/icelander/b68f2bf2b4ffefec93400cb050211cf1>`_.
```
lieut-data, line 60: nit, space before comma:
```suggestion
- For information about the webhook payload for run start, see the `PlaybookRunWebhookPayload struct <https://github.com/mattermost/mattermost-plugin-playbooks/blob/b4c8058d8660efe35050bc7eb080e3819c7ab09c/server/app/playbook_run_service.go#L176-L185>`_. An example of the JSON payload for a run start `is available here <https://gist.github.com/icelander/b68f2bf2b4ffefec93400cb050211cf1>`_.
- ```

lieut-data: Both of these permalinks feel like they will quickly become stale, despite their intent to be otherwise. This is fine to merge, but can you help create a follow-up ticket to document this more explicitly in our code that will serve as a better permalink target?

lieut-data (approved):

cwarnermm (approved):
Thanks!

source/playbooks/setting-up-playbooks.rst
justinegeffen: Great point; will do.

PR #5038, opened 2021-09-10
Fix environment pickling

neflyte:
I noticed that my previous PR had introduced an issue with the Sphinx "environment pickling" process. The net result of the issue is the loss of incremental builds, although the build speed has improved. Please accept this PR as a fix to the issue including further code cleanups and documentation. :)

--

# Fix Environment Pickling

### Primary

- Repair the recently-broken Sphinx environment pickling process

### Secondary

- Improve documentation of the `reredirects` and `sitemap` extensions

### Sphinx Environment Pickling

- The Sphinx build process consists of [5 phases](https://www.sphinx-doc.org/en/master/extdev/index.html#build-phases):
  - Initialization
  - Reading
  - Consistency checks
  - Resolving
  - Writing
- During the _Reading_ phase, Sphinx gathers a list of all documents to be built
- This list of documents, among other build information, is persisted in the `build/doctrees` directory as a "pickle" file named `environment.pickle`
  - [pickle](https://docs.python.org/3.7/library/pickle.html) is Python's built-in object serialization library
  - It allows arbitrary objects to be serialized into binary and deserialized back from binary
- Prior to Sphinx's _Reading_ phase the existing `environment.pickle` file is loaded if it exists
  - The list of documents read during previous builds is included with this information
- When Sphinx determines the list of files to build, the existing list of documents is taken into account
  - Only added documents, deleted documents, and modified documents are affected
  - This dramatically reduces the time to complete the build as only necessary changes are built

### Pickling Issue

- The recent [environment-cleanup](https://github.com/mattermost/docs/pull/4843) PR introduced parallel read and write support to the `sitemap` extension
- The parallel read support, while technically functional, inadvertently caused the environment picking process to fail when reading the existing `environment.pickle` file
- This issue results in a full build of all documents regardless of their current build status
  - In effect, it is the same as executing `make clean html` for each build

#### Technical Details

- The parallel read solution was implemented using a first-class [multiprocessing.Queue](https://docs.python.org/3.7/library/multiprocessing.html#multiprocessing.Queue) object
- This `Queue` object, while technically suited to parallel processing, is not easily serializable with Python's `pickle` library

#### Added

- Documentation for all functions
- Status output in a format consistent with Sphinx
- Improved parallel read support, including:
  - Functionality to prune documents from the pickled environment when they are no longer included in the build
  - Functionality to merge document names collected from parallel readers

#### Changed

- Changed the `Queue` implementation of parallel read to [List](https://docs.python.org/3.7/library/stdtypes.html#list), a Python built-in object which is known to serialize without issue
- Changed the phase at which document information is collected from the _Writing_ phase to the _Reading_ phase

#### Added

- Documentation for all functions
- Status output in a format consistent with Sphinx

## Usage Notes

- Run `make clean` before using the code in this PR to ensure the existing pickled environment files are removed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,7 @@
 SOURCEDIR       = source
 BUILDDIR        = build
 WARNINGSFILE    = $(BUILDDIR)/warnings.log
-ERRORSFILE      = $(BUILDDIR)/errors.log
-SPHINXOPTS      ?= -j auto -w $(WARNINGSFILE)
+SPHINXOPTS      ?= -j auto
 SPHINXBUILD     ?= pipenv run sphinx-build
 SPHINXAUTOBUILD ?= pipenv run sphinx-autobuild
 
@@ -15,7 +14,7 @@ SPHINXAUTOBUILD ?= pipenv run sphinx-autobuild
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile livehtml clean
+.PHONY: help Makefile livehtml
 
 # Run `make livehtml` to start sphinx-autobuild
 livehtml:
@@ -26,7 +25,4 @@ livehtml:
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
 	@mkdir -p "$(BUILDDIR)"
-	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O) 2>>"$(ERRORSFILE)"
-
-clean: # clean the build directory
-	rm -rf build/
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O) 2>>"$(WARNINGSFILE)"
--- a/extensions/reredirects/__init__.py
+++ b/extensions/reredirects/__init__.py
@@ -1,50 +1,65 @@
 #
-# reredirects - A fork of the sphinx_reredirects extension with parallel read and write support enabled
+# reredirects - The sphinx_reredirects extension with parallel read and write support enabled
+#
+# Parallel read/write support by Alan Lew <alan@ethereal.cc> (https://github.com/neflyte/)
 #
 import re
 from fnmatch import fnmatch
 from pathlib import Path
 from string import Template
-from typing import Dict, Mapping, Tuple
-
+from typing import Dict, Mapping, Tuple, Any, List
 from sphinx.application import Sphinx
+from sphinx.environment import BuildEnvironment
 from sphinx.util import logging
 from sphinx.util.console import bold, colorize, term_width_line  # type: ignore
 
 OPTION_REDIRECTS = "redirects"
-OPTION_REDIRECTS_DEFAULT: Dict[str, str] = {}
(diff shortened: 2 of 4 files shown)


justinegeffen (approved):
Thank you for this awesome PR, @neflyte! 

I'm approving it based on your explanation but I would like to find out the best way to test and possibly troubleshoot the changes. In addition, is this tied to any dependencies, and will it be affected by any future upgrades? 

I have a lot to learn about Sphinx, so apologies for the newbie questions!

cwarnermm:
Great questions, @justinegeffen! 

**To test PR 5038**
1. Switch to your local branch containing Alan's initial environment cleanup PR. (Or, clone Alan's branch from his latest PR if that branch is no longer available).
2. Run ``make clean html`` to ensure any existing pickled environment files are removed. The build will run for approximately the same amount of time as any other build.
3. When the build is complete, confirm that the pickled environment loads without error. 
Screen Shot 2021-09-14 at 8 43 18 AM

Run ``make html`` to build the docs again. This time, the build should be very fast because Sphinx can detect (though pickling) that nothing has changed since the last build). The pickled environment should not throw an error.
Screen Shot 2021-09-14 at 8 44 46 AM

**Review the sitemap**
Compare the contents of the ``sitemap.xml`` you just generated locally against the sitemap.xml in Production.
- Ensure the sitemap doesn't contain version information
- It's okay that the contents and order of the sitemap differs due to site architecture updates over time

**Is this update tied to dependencies?**
- No. The sphinx-sitemap extension we're using is custom and is no longer tied to the public version. Updates to the public extension repository will no longer impact our extension in any way. 
    - This decoupling was deliberate and intentional since the public version of the extension was developed before parallel processing was supported in Sphinx. The public extension code hasn't been updated to take advantage of faster processing capabilities. 
    - Over time, as Sphinx functionality and popularity grew, a number of contributors have submitted attempts to address this limitation. 
    - Alan's initial PR for the Docathon applied those learnings; however, those community-driven fixes were applied used a method to organize the data that isn't compatible with the persistence method that Sphinx supports. 
        - The original fix Alan applied went too far by using a queue object which is good for parallel processing but isn't compatible with the way that Sphinx persists its build information. As a result, the pickle file was being written, but couldn't be read, and a pickling error would be thrown during builds.
        - PR-5038 stores the build data differently using a basic Python list object (a built-in data type in Python). This means that builds are now using facilities Sphinx provides to do the parallel reading instead of the queue object.
- The docs sitemap was being generated during the writing phase of the build process. With PR-5038, it's now being generated during the reading phase instead. This means that sitemap generation is no longer the last action performed during the build process; it's now the second action performed.
    - This aligns well with Sphinx's extension model because the sitemap is simply collecting a list of page names. It doesn't need to change the pages in any way -- it just needs the list of pages themselves. Because no writing is required, it's better to do this work during the reading phase where Sphinx has event hooks that retrieves the doc names, and this retrieval can be parallelized.  

**Future Upgrades**
- It's now our team's responsibility to maintain this custom code going forward. In the future, when we choose to upgrade Sphinx to a newer version, it will be our responsibility to ensure that all of our extensions continue to work.

**Important Note**
- In Python, information is organized in a specific way, and persisting data for later retrieval is a platform-dependent. PR-5038 has been tested on macOS and Unix-based platforms. It has not been tested on Windows platforms.

cwarnermm (approved):
Thank you for this great update and for explaining the changes so clearly.

justinegeffen:
Tested locally and working perfectly, thanks @neflyte and @cwarnermm!

PR #5041, opened 2021-09-14
Remove references to product names.

BenLloydPearson:
#### Summary
I removed most of these references when this guide was created, but missed these. I think this page should focus on features rather than packaging because someone who visits this page is probably more interested in what they can do rather than what they need to buy.

#### Ticket Link
N/A


cwarnermm (approved):
Thanks!

justinegeffen (approved):

PR #5043, opened 2021-09-14
Fixed formatting for 301 redirects causing build failures

cwarnermm:
A small handful of 301 redirects were causing build failures due to unexpected line breaks. Fixed these formatting issues, and removed one duplicate URL from the list.


justinegeffen (approved):
Awesome fixes, thank you!

Issue #4713, opened 2021-07-07
Detect broken documentation links


cwarnermm:
Mattermost product documentation is written in RST, generated using Sphinx, and formatted using a Sphinx theme called ReadTheDocs.

The Mattermost Technical Writing Team is looking for a way to detect broken links in Mattermost product documentation in their local development environments as part of documentation builds. Detecting broken links before content is published to Production ensures fewer broken links on the live documentation site at https://docs.mattermost.com.

VS Code-based tooling compatible with Sphinx 3.4.3 or later and RST source files are most welcome additions to our existing toolset.

Er-Riya:
Up for grabs

rootbid:
I think [CheckExternalLinksBuilder](http://www.sphinx-doc.org/en/master/builders.html#sphinx.builders.linkcheck.CheckExternalLinksBuilder) would be useful here.
It's basically used by calling 'sphinx-build' with '-b linkcheck' option. Please see [sphinx-build](http://www.sphinx-doc.org/en/master/man/sphinx-build.html?highlight=linkcheck) for more info.

cwarnermm:
Thanks, @rootbid! This sounds like exactly the thing we need to try out! Really appreciate the recommendation!

Issue #4488, opened 2021-04-15
Learn how to use the Sphinx link check feature


cwarnermm:
Learn how to use the link check feature that is available through Sphinx that identifies broken URL links. 
- [Introduction to the feature](https://www.writethedocs.org/guide/tools/testing/#sphinx)
- [Sphinx documentation](https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.linkcheck.CheckExternalLinksBuilder)

larkox:
I have been looking into this. The usage is pretty simple (`make linkcheck`). There would be several possible action points here:
- Adding this check to the workflow
  This seems complicated. There are tons of links, and many of them are expected to fail (github users in the changelogs that may have been disabled for whatever reason). We can ignore those, but it is for sure something we cannot do for every single new PR. Either way, we should do this after we hit 0 breaks.
- Fix what is broken
  We can spend some time going through the results fixing everything that is broken.
- Fix redirects
  Redirects also may mean something is broken. We should review the redirects and use the expected URL instead of the old one.

Do you want me to take care of any of these points?

cwarnermm:
Thank you SO MUCH for digging into this, @larkox. 

You make an excellent point about why not to add this check to our GitHub workflows. That alone is amazing to have confirmed!

The Docs team would benefit greatly from being able to run checks on demand and review the output to find both broken links and broken redirects. 

I just tried the ``make linkcheck`` myself locally. It works as advertised -- it spits out a LOT. In fact, it's been running the ENTIRE time I've been writing this response! :)

From your perspective, it is possible to take the output from this command, and split it out into multiple files we could then take action on, one at a time? I'm imagining something like an output file for links that return OK, one for broken links, one for redirect problems, etc? Even better if we could them pull those results into a tool like a spreadsheet where we can track each issue & fix (at least until we're at 0 issues, and **then** we can work on ignoring the rest and adding build-level process in to assist. What do you think?

larkox:
Hi @cwarnermm !

You can find there a breakdown of what I considered relevant from the links. https://docs.google.com/spreadsheets/d/1eNtyjKBO9IO0xZSwZG9FF_wCWfEY76Z2dWuwzOq3Wjc/edit?usp=sharing

All `http://github.com/` links are ignored.

Let me explain a bit what is there and why:
- broken: Pretty obvious. All the broken links. We should look into them to fix them.
[...]

PR #5058, opened 2021-09-21
Detect broken documentation links

rootbid:
#### Summary

Implements `linkcheck` based on the discussions #4713 and #4488 in order to detect broken documentation links.

Resolves: #4713

#### Usage
Running `make linkcheck` generates two output reports `build/output.txt` & `build/output.json`
- `output.txt` contains information on broken and redirected links as identified by linkcheck
- `output.json` contains information on all links checked by linkcheck. This output can be further filtered and processed based on the desired action. Possibly with the help of [jq](https://stedolan.github.io/jq/). For example; filter out broken links using `cat build/output.json | jq -c 'select( .status == "broken" )'
`

#### Configuration
A little context on the linkcheck configuration made with this PR,
1. Based on the discussion over #4488, `exclude_patterns` is overridden in `Makefile` to exclude archive, process directories from source as they are not used in the documentation. Additional files can also be excluded using this.
2. Based on the local execution of this implementation, some links were found to be generating false-positives and were configured to be ignored in `source/conf.py` under the 'linkcheck_ignore' option. You will also find a lot of anchor links for "https://docs.mattermost.com/" reported as broken (which appear to be redirecting to different urls), this can also be ignored if that's the expected behavior.

Please see [Options for the linkcheck builder](https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-the-linkcheck-builder) for more info.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,13 @@ livehtml:
 	@mkdir -p "$(BUILDDIR)"
 	@$(SPHINXAUTOBUILD) "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
+# Run `make linkcheck` to check external links
+# Overriding `exclude_patterns` configuration to exclude
+# directories or files not included in the documentation
+linkcheck:
+	@mkdir -p "$(BUILDDIR)"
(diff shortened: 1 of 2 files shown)


cwarnermm:
We're really excited to dig into this, @rootbid! Thank you so much!

jasonblais:
@rootbid thanks for the contribution!

PR #5059, opened 2021-09-21
Add swiftype annotations for better indexing

svelle:
This PR adds the needed annotations for swiftypes crawler so that it excludes navigational elements and only indexes the article body which should result in much more useful autocomplete previews.


justinegeffen (approved):
This looks good to me, and doesn't look like it touches any of the tabbing changes. I'd like @cwarnermm to confirm that though!

PR #5065, opened 2021-09-21
Update playbooks.rst: Revised landing page links.

justinegeffen:
Revision of the Playbooks landing page links to replace boards with playbooks.


BenLloydPearson (approved):
Oh wow, thanks for catching this!

PR #5075, opened 2021-09-24
Re-added missing upgrade command example

cwarnermm:
Added back missing command example to remove all files except special directories from within the current mattermost directory.


justinegeffen (approved):

justinegeffen:
Looks good to me but note that this isn't my forte and I have not tested the command.

PR #5078, opened 2021-09-27
Remove invisible paragraph icon from docs html

svelle:
This PR removes the invisible paragraph icon that's being added to the html.
This should also make it disappear from the swiftype autocomplete results returning more useful descriptions when searching.

Side by side
image
image

I've tried a blank space as a substitution but it didn't get registered by sphinx when rendering the html. 
We could think about substituting with a less intrusive icon if this cosmetic change is unwanted.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -574,7 +574,7 @@ html_theme_options = {
     "titles_only": False,
 }
 
-html_permalinks_icon = ""
+html_permalinks_icon = " "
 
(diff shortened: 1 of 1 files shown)


svelle:
image


@cwarnermm I changed up the whitespace character so this now should be barely a difference to the old look

cwarnermm (approved):
Thanks! No concerns from me on this change, especially if it gets us one step closer to federated search!

PR #5086, opened 2021-09-27
Updated links

Sayanta66:

### DIFF
--- a/source/deploy/consider-mobile-vpn-options.rst
+++ b/source/deploy/consider-mobile-vpn-options.rst
@@ -20,7 +20,7 @@ A Virtual Private Network (VPN) allows a device outside a firewall to access con
 
 We recommend one of two options: `per-app VPN <#id3>`_ or a `device VPN <#id4>`_ to secure your deployment. Both options are compatible with most EMM providers. 
 
-We also recommend following our `recommended steps to secure your deployment <https://docs.mattermost.com/mobile/mobile-appstore-install.html>`__ and to review the following commonly-asked questions about data security on mobile devices:
+We also recommend following our `recommended steps to secure your deployment <https://docs.mattermost.com/deploy/consider-mobile-vpn-options.html>`__ and to review the following commonly-asked questions about data security on mobile devices:
 
 - `How data is handled on a device after an account is deleted? <https://docs.mattermost.com/deploy/mobile-faq.html#how-is-data-handled-on-mobile-devices-after-a-user-account-is-deactivated>`__
 - `What post metadata is sent in mobile push notifications? <https://docs.mattermost.com/deploy/mobile-faq.html#what-post-metadata-is-sent-in-mobile-push-notifications>`__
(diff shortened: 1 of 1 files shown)


source/deploy/consider-mobile-vpn-options.rst
```diff
@@ -45,7 +45,7 @@ Connect via Corporate Proxy Server
...
 
-- `How do I receive mobile push notifications if my IT policy requires the use of a corporate proxy server? <https://docs.mattermost.com/deploy/mobile-faq.html#how-do-i-receive-mobile-push-notification-if-my-it-policy-requires-the-use-of-a-corporate-proxy-server>`__
-- `Deploy Mattermost with connection restricted post-proxy relay in DMZ or a trusted cloud environment <https://docs.mattermost.com/deploy/mobile-faq.html#deploy-mattermost-with-connection-restricted-post-proxy-relay-in-dmz-or-a-trusted-cloud-environment>`__
-- `Whitelist Mattermost push notification proxy to bypass your corporate proxy server <https://docs.mattermost.com/deploy/mobile-faq.html#whitelist-mattermost-push-notification-proxy-to-bypass-your-corporate-proxy-server>`__
-- `Run App Store versions of the Mattermost Mobile apps <https://docs.mattermost.com/deploy/mobile-faq.html#run-app-store-versions-of-the-mattermost-mobile-apps>`__
+- `How do I receive mobile push notifications if my IT policy requires the use of a corporate proxy server? <https://docs.mattermost.com/deploy/mobile-faq.html#how-do-i-receive-mobile-push-notifications-if-my-it-policy-requires-the-use-of-a-corporate-proxy-server>`__
```
justinegeffen, line 48: Thanks for these changes. Could you elaborate on changing the URLs from the docs pages to the developer pages? The docs pages appear to be correct but if the developer content is more current than the docs content please let us know!

source/deploy/consider-mobile-vpn-options.rst
```diff
@@ -20,7 +20,7 @@ A Virtual Private Network (VPN) allows a device outside a firewall to access con
 
 We recommend one of two options: `per-app VPN <#id3>`_ or a `device VPN <#id4>`_ to secure your deployment. Both options are compatible with most EMM providers. 
 
-We also recommend following our `recommended steps to secure your deployment <https://docs.mattermost.com/mobile/mobile-appstore-install.html>`__ and to review the following commonly-asked questions about data security on mobile devices:
+We also recommend following our `recommended steps to secure your deployment <https://docs.mattermost.com/deploy/consider-mobile-vpn-options.html>`__ and to review the following commonly-asked questions about data security on mobile devices:
```
justinegeffen, line 23: Thanks for the update! Could you let us know why this link was chosen to replace the previous on? The linked page doesn't seem to have any steps in it so I am wondering if there may be a better option than this link. :)

source/deploy/consider-mobile-vpn-options.rst
Sayanta66: The redirected link was coming like this. So changed it

justinegeffen: Apologies! I am referring to the links below this comment - they've been changed from docs links to links in the developer documentation. If the content in those docs is better or more accurate that'll be good to know so that we can plan to align them or consolidate them with the docs as opposed to having multiple sets of the same content. :)

justinegeffen: I've found the culprit, thanks for finding this @Sayanta66! Very good catch. The content has actually been moved to the docs set, so we'll retain all the links as they currently are. :)

justinegeffen: If you could revert all but the first one, that'd be awesome. :)

Sayanta66: Thanks a lot. Doing the needful

Sayanta66: Resolved

source/deploy/consider-mobile-vpn-options.rst
justinegeffen: Just wanted to follow up on this question as it would be great to confirm the replacement does contain the correct information so we don't merge inaccurate content.

PR #5089, opened 2021-09-29
Redesign 2021

emdecr:
Hi All,

Just pushing up some of the changes outlined in JuliAnn's Figma files to start the preview/review process.

Quick overview of the changes:
- Totally new homepage styling
- New favicon
- Base fonts changed to: Trade Gothic (headings), Fira Mono (accent), Noto Sans (body)
- Code blocks have a different background
- Note/Tip modules looks a bit different
- Footer on the home page is a copy of the one that will be live on mm.com for the relaunch
- Same things for the navigation, but this will be a global change on all docs pages.

Similar to the code blocks and note sections, let us know if you see any other types of content that need a bit of redesign-love.

### DIFF
--- a/source/about/certifications-and-compliance.rst
+++ b/source/about/certifications-and-compliance.rst
@@ -56,8 +56,8 @@ Data Management
 ^^^^^^^^^^^^^^^^
 
 - **Data Retention:** Use `data retention <https://docs.mattermost.com/comply/data-retention-policy.html>`__ to automatically erase data after a set period of time, a feature that meets the Right to Erasure principle. In Team Edition, you can use database scripts to achieve the same result.
-- **Profile Deletion:** Delete a user’s personal information via `a command line tool <https://docs.mattermost.com/manage/command-line-tools.html#mattermost-user-delete>`__. The command permanently deletes all user information including messages created by the user.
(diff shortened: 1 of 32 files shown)


BenLloydPearson:
@emdecr I noticed that the search bar is visible on the home page for the mobile layout, but not the desktop layout. I think this is an oversight on my end with the initial design, can we add the search bar to the home page desktop layout too?

BenLloydPearson (approved):
I updated one link on the home page. Other than the lack of a search bar on the home page, this looks good to me.

emdecr:
@BenLloydPearson @cwarnermm With my latest push (I don't think the preview site has updated at the time of me writing this comment), we're adding the search back on desktop and mobile, in a slightly different treatment.

JuliAnn's also added some new options for search on the actual docs pages (she'd prefer if the search icon were not in the nav next to to the logo). I'm gonna see what's possible and circle back.

emdecr:
@BenLloydPearson @cwarnermm 
I've implemented JuliAnn's proposals for search:

- The search bar appears on the homepage for both mobile and desktop
- For internal docs pages, the search icon has been removed from the top-left spot in the nav, and moved under the breadcrumb links beside the Edit link

I see a go-ahead/approval from Ben in a comment above, but Carrie obvs let us know if the team has any more pieces of feedback. I remember you mentioning it would be shared with more folks. As JuliAnn is out after tomorrow, I'm not sure any large design-related changes can be implemented, but we can see how we can assist. TY!

PR #5090, opened 2021-09-29
Update boards.rst

justinegeffen:
Removed "Work with views" from landing page list as this has been folded into another section.


cwarnermm (approved):
Thanks!

PR #5109, opened 2021-10-04
chown only the Mattermost directory not the parent

firasm:
#### Summary

Change 

```
sudo chown -R mattermost:mattermost {install-path}
```

to:
```
sudo chown -R mattermost:mattermost {install-path}/mattermost
```

Rationale

I followed these instructions to upgrade MM but I ended up accidentally breaking my installation because I have Mattermost installed in `/srv/` and that happens to be where `mysql` is also installed. So after the upgrade I couldn't launch mysql again until I chowned the permissions back.

Isn't it safer to only chown the Mattermost directory and everything in it instead?

### DIFF
--- a/source/upgrade/upgrading-mattermost-server.rst
+++ b/source/upgrade/upgrading-mattermost-server.rst
@@ -149,7 +149,7 @@ Upgrading Mattermost Server
 
   .. code-block:: sh
          
-    sudo chown -R mattermost:mattermost {install-path}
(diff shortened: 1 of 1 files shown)


cwarnermm:
Thanks, @firasm, and great catch!

cwarnermm (approved):
Thank you for this product documentation contribution!

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @firasm

PR #5123, opened 2021-10-07
Update intro to remove old packaging language.

BenLloydPearson:
#### Summary
This is a small fix to update the changelog intro to remove references to packaging. I'm assuming this page is only for self-hosted Mattermost, so feel free to remove that mention if I'm wrong.

#### Ticket Link
n/a


jasonblais:
@BenLloydPearson @cwarnermm Curious why are we removing the reference? It is referencing the two different binaries (Team Edition, Enterprise Edition) with two different licensing. It's not related to our packaging.

PR #5125, opened 2021-10-07
Fix example webhook wording

bbodenmiller:


emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @bbodenmiller

PR #5156, opened 2021-10-14
Updated the links

Sayanta66:
Updated the deprecated links


emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @Sayanta66

PR #5198, opened 2021-10-21
Terminology updates for Account Settings

cwarnermm:
As a follow-up to tutorial and onboarding language updates, also updating product documentation terminology to clearly differentiate between configurations available through Account Settings and Settings. 

These updates:
- ensure that terminology used is precise
- highlight the most recognized features from each setting
- provide users with accurate and direct paths to desired configuration options
- are applied consistently across the product documentation.

### DIFF
--- a/source/messaging/managing-account-settings.rst
+++ b/source/messaging/managing-account-settings.rst
@@ -202,7 +202,7 @@ Notification Duration
 
 *Removed in June 16th, 2018 release*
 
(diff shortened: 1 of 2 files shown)


esethna:
@cwarnermm are we going to make the change from Account Settings > Profile separately?

cwarnermm:
Agreed! Thanks for clarifying @esethna! This PR can remain as-is, then, and we'll revisit the terminology change of "Account Settings" to "Profile" in a future docs PR once the change has been addressed in-product.

Meanwhile, I'll start a Webapp PR for the in-product terminology change.

PR #5213, opened 2021-10-25
Terminology change from switch to navigate

cwarnermm:
Replaced existing instances of "switch" or "switcher" with "navigate" in contexts where users are moving from one area of Mattermost to another.


wiersgallak (approved):

source/boards/accessing-boards.rst
```diff
@@ -18,7 +18,7 @@ Access your Boards
   :target: https://mattermost.com/deploy
   :alt: Available for Mattermost Self-Hosted deployments.
 
-You can access all your boards as well as boards in channels you belong to in the Boards dashboard. Select **Boards** from the Product menu in the top-left corner of Mattermost to open the dashboard and switcher. You can also select the Boards icon in the channel header to open the board for the current channel, Direct Message, or Group Message.
+You can access all your boards as well as boards in channels you belong to in the Boards dashboard. Select **Boards** from the Product menu in the top-left corner of Mattermost to open the dashboard. You can also select the Boards icon in the channel header to open the board for the current channel, Direct Message, or Group Message.
```
justinegeffen, line 21: Just a note: This is actually called the switcher in-product which is why it's in here post 6.0. :) 

Screen Shot 2021-11-08 at 20 40 14

justinegeffen (approved):
Just one, non-blocking, note. :)

source/boards/accessing-boards.rst
cwarnermm: Thanks! I'll revert this change. I suspect that this in-product term will likely need to change in the future given similar terminology changes we're in the process of making.

cwarnermm, line 21:
```suggestion
You can access all your boards as well as boards in channels you belong to in the Boards dashboard. Select **Boards** from the Product menu in the top-left corner of Mattermost to open the dashboard and switcher. You can also select the Boards icon in the channel header to open the board for the current channel, Direct Message, or Group Message.
```

justinegeffen: Agreed! I'll discuss it with the Focalboard team.

justinegeffen: Raised here: https://community.mattermost.com/core/pl/j5stsoe43pymzcxjrgjpmjqbko

justinegeffen: As per the thread, I'll make this update but it will shortly be moot. :)

PR #5222, opened 2021-10-28
Page clarifications based on visitor feedback

cwarnermm:
Addressed visitor feedback about the title/content of this page being misleading. Visitors were looking for server upgrade details and found this section or page via Search instead which doesn't contain the details they're looking for. Clarified title and added direct link to the Upgrading Mattermost Server page.


justinegeffen (approved):
Nice change! :)

PR #5225, opened 2021-10-28
Dropped "Edition" references for Enterprise

cwarnermm:
To clarify the distinction between the new subscription plans available as of v6.0, and the legacy editions of Mattermost available, dropping references to "edition" when referencing the subscription plan name.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -816,6 +816,7 @@ redirects = {
     "messaging/accessing-your-workspace.html": "https://docs.mattermost.com/messaging/signing-in.html",
     "guides/messaging": "https://docs.mattermost.com/guides/channels.html",
     "focalboard/installing-boards": "https://docs.mattermost.com/guides/boards.html",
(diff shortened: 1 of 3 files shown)


jasonblais (changes requested):
It is very important to keep the distinction between Team Edition (TE) and Enterprise Edition (Starter, Professional, Enterprise plans), as they are two different binaries, with two separate terms of use, privacy policy and licensing.

If we want to create more clarity between the Enterprise Edition binary and the Enterprise plan, perhaps we call "Starter", "Professional", "Enterprise" as "plans" instead of "editions"?

wiersgallak:
Thanks @jasonblais.  @cwarnermm Let's cancel this PR and you and I can do a working session of reviewing how each ```Enterprise``` is being used in context and add the "plan" behind it as applicable to distinguish between the SKU vs. the Binary.

cwarnermm:
Thanks @wiersgallak and @jasonblais. I agree that a working session to clarify binary from SKU instances is a great way forward. I've added a ticket to our v6.1 docs board to track this. Closing this PR.

jasonblais:
Thanks both! Happy to also join the working session if helpful

PR #5232, opened 2021-11-01
Split FAQ into sections

cwarnermm:
The FAQ topics under About Mattermost have been split apart into smaller sections to make the content easier to read and navigate for users.


wiersgallak (approved):
I reviewed the split of questions based on the proposed headings.  Some of these we may want to consider adding SMEs to review the content and update it as it likely can use a refresh.  I didn't review from the detailed content perspective at this time. We should determine a process for that.

cwarnermm:
Thanks, @wiersgallak! I've added a ticket to our v6.1 docs board to develop a process for quarterly PM reviews of this content going forward.

PR #5254, opened 2021-11-09
Clarified config setting unit of measure

cwarnermm:
Updated the **Idle Timeout** config setting to clarify that the value specified is seconds. Addresses https://github.com/mattermost/docs/issues/5253


wiersgallak (approved):
Thanks!

PR #5260, opened 2021-11-10
Remove license key deletion instructions

BenLloydPearson:
#### Summary
The instructions to remove the license key and downgrade the server are out of place in the Install a License Key section. These instructions will be referenced in important trial flows so we want to make sure they're on point.

#### Ticket Link
N/a


source/upgrade/installing-license-key.rst
```diff
@@ -25,8 +25,7 @@ You can use the System Console, the mmctl, or the CLI to add or change a Matterm
     .. tab:: Using System Console
 
         1. Go go **System Console > About > Edition and License**.
-        2. Select **Remove Enterprise License and Downgrade Server**. This clears the license from the server and refreshes the System Console.
-        3. Upload the new license key file.
```
jasonblais, line 29: @BenLloydPearson agree with the importance of this page if used for trial flows, and that removing an Enterprise license is not relevant for a trial user.

I hate to throw a curveball ~~ This page also serves a second use case for replacing an existing license key. Thus, removing this step would break the instructions for an admin who needs to remove the existing license key first before installing a new one (e.g. installing a production license key after an active trial, or installing a license key after a renewal).

image

I'm 0/5 on approach, we could have two separate pages to cover these two use cases.

cwarnermm: Great point, @jasonblais. 

@BenLloydPearson I suggest we remove steps 2 & 3 as you have in this PR, then move the details of those steps to the **Changing an existing license key** subsection below on the same page. That way above "the fold" trial users instantly see what they need for success, and below the fold existing users will find details to change their license if required.

cwarnermm: One minor request: Can you correct 2 existing typos as well, please? Search for "command to to" and remove the second "to". There are 2 instances of this typo to correct. Thanks!

cwarnermm (approved):
Thanks!

PR #5269, opened 2021-11-17
Update nav and footer

emdecr:
#### Summary
Hi team, this PR syncs the nav with the one seen on mm.com. There's been a few changes recently under the Login and Developer drop-downs.

I've also adjusted the notification bar to accommodate the longer copy. I've taken out the wrap property that was causing the text to cut off and made the base height of the bar on mobile larger. If you change the copy in the future to something much shorter, we'll probably have to adjust styles a bit (quick change). With the way the nav is set up now, I need to state a set height for the notification bar so I know how much to offset the navigation bar from the top (on mobile).

Let me know if you have any questions. Thanks!

cc: @cwarnermm


cwarnermm:
Thanks, @emdecr! I'm seeing a few issues when reviewing the preview.

In normal browser mode (tested with Microsoft Edge Version 93.0.961.47 (Official build) (x86_64)
- Clicking the Mattermost logo in the top left corner gives the user a 404 error (this may only be an issue in the preview; currently production is working as expected).

When the browser is in Incognito mode (which may or may not be a valid use case for you):
- Some Login and Developer header images are missing
- The Solutions header components are partially cut off
(see attached images)
Screen Shot 2021-11-18 at 8 41 26 AM
Screen Shot 2021-11-18 at 8 41 15 AM
Screen Shot 2021-11-18 at 8 41 09 AM

cwarnermm (changes requested):
Thanks, @emdecr! I've noted some issues I'm seeing in the current preview.

emdecr:
Thanks for the review @cwarnermm – much appreciated! And great catches!

Just pushed some changes up:
- I believe we should be good once in production for the logo error
- Fixed the typos in the footer
- Mobile & Desktop Apps + Mattermost server links should go to the proper mm.com pages
- Enterprise Trial License and the Cloud Trial links go to their respective pages
- Terms of Use page title has been changed
- Fixed the paths for images in the nav for internal docs pages (also in incognito mode)

The only thing I can't replicate is the Solutions items being cut off in Edge in incog mode.

cwarnermm (approved):
Latest changes address all issues identified with the exception of the Solutions header cutoff when the browser is in Incognito mode. This issue can be reproduced consistently when the browser window is between standard mobile and larger screen sizes. Approving for expediency.

cwarnermm:
@emdecr - Happy to help provide any browser data that would help with repro steps.

emdecr:
@cwarnermm I believe the Solutions dropdown bug we were seeing has been fixed with my latest push. Thanks again for catching that!

PR #5294, opened 2021-12-02
Removed outdated configuration settings

cwarnermm:
Removed the following plugin settings from the `configurable only via ``config.json`` section of the documentation as per Engineering: GitHub, Custom User Attributes, Jira, and Net Promoter Score.  The inclusion of these plugins in the config settings documentation was causing reader confusion by mixing up server config settings with plugin config settings. 

Also updated the WelcomeBot plugin configuration documentation to clarify that its the settings for the plugin itself that aren't available in System Console that can only be set via ``config.json``.


mickmister (approved):
LGTM :+1: Thanks Carrie!

justinegeffen (approved):

aaronrothschild (approved):
LGTM

PR #5305, opened 2021-12-07
Added Developerflags config setting

cwarnermm:
Documentation for: https://github.com/mattermost/mattermost-server/pull/18801
- Added DeveloperFlags config setting to the product documentation as well as the Telemetry page.


source/configure/configuration-settings.rst
```diff
@@ -1924,9 +1924,9 @@ Enable Developer Mode
 
 **False**: Users are not alerted to Javascript errors.
 
-+-------------------------------------------------------------------------------------------------------------+
-| This feature's ``config.json`` setting is ``"EnableDeveloper": false`` with options ``true`` and ``false``. |
-+-------------------------------------------------------------------------------------------------------------+
++-----------------------------------------------------------------------------------------------------------------------------+
+| This feature's ``config.json`` setting is ``"ServiceSettings.EnableDeveloper": false`` with options ``true`` and ``false``. |
```
lieut-data, line 1928: Prefixing with `ServiceSettings.` isn't consistent with the rest of the documentation, and is technically incorrect: the setting isn't named `ServiceSettings.EnableDeveloper` but is actually nested as, `"ServiceSettings": { ..., "EnableDeveloper": false, ...}`, but we haven't traditionally expressed it this way in the documentation. (I definitely think we should find a way to fix this!)

source/configure/configuration-settings.rst
```diff
@@ -6052,6 +6052,19 @@ This setting has been added as a requirement to support `Collapsed Reply Threads
...
    Enabling this configuration setting doesn’t retroactively follow threads for older actions taken prior to the setting being enabled. For example, if a user comments on an old thread, they automatically follow the thread if this setting is enabled regardless of whether the thread existed when this setting was disabled.
 
+Developer Flags
+^^^^^^^^^^^^^^^
+
+|all-plans| |self-hosted|
+
+This setting isn't available in the System Console and can only be set in ``config.json``.
+
+This configuration setting specifies a list of strings where each string is a flag used to set the content security policy for the Mattermost web app. Each flag must be in the format ``flag=true`` (e.g. ``unsafe-eval=true,unsafe-inline=true``). This configuration setting is disabled by default and requires `Developer Mode <https://docs.mattermost.com/configure/configuration-settings.html#enable-developer-mode>`__ to be enabled.
```
lieut-data, line 6062: For now, only `unsafe-eval` and `unsafe-inline` are supported. Can we make that explicit? It might be worth documenting as:
```
unsafe-eval: Adds the `unsafe-eval` CSP directive to the root webapp, allowing increased debugging in developer environments.
unsafe-inline: Adds the `unsafe-inline` CSP directive to the root webapp, allowing increased debugging in developer environments.
```

We should probably emphasize that these values shouldn't typically be enabled in production.

lieut-data (changes requested):
Thanks, @cwarnermm! A few clarifications below.

cwarnermm:
Great feedback, @lieut-data! Thank you!

source/configure/configuration-settings.rst
justinegeffen, line 6062:
```suggestion
This configuration setting specifies a list of strings where each string is a flag used to set the content security policy for the Mattermost Web App. Each flag must be in the format ``flag=true`` (e.g. ``unsafe-eval=true,unsafe-inline=true``). This configuration setting is disabled by default and requires `Developer Mode <https://docs.mattermost.com/configure/configuration-settings.html#enable-developer-mode>`__ to be enabled.
```

justinegeffen (approved):
One small correction but otherwise looks good from my side.

source/configure/configuration-settings.rst
cwarnermm: I appreciate this feedback, @lieut-data, thank you. We had started introducing this convention a few releases ago as new config settings have been introduced; however, I'm reverting those recent additions in this PR as a result of your input. 

I've added notes to an existing card on our docs backlog to revisit how best to communicate where, in a ``config.json`` file, users can find a given config setting. Any additional input or suggestions you have would be most welcome.

source/configure/configuration-settings.rst
cwarnermm: Great clarification. Thanks!

lieut-data (approved):

PR #5307, opened 2021-12-09
Remove unused and problematic images

mrckndt:
This commit removes two duplicate files which can cause problems between Linux and macOS/Windows because of the case insensitivity of the latter. The third file starts with a whitespace and is a duplicate too.


cwarnermm:
@mrckndt - If we were to correct the mixed case in 2 of the filenames and removed the erroneous whitespace from the one filename, but left the images otherwise in the product documentation, would that address the issue you're experiencing?

mrckndt:
@cwarnermm There is no need to keep them because those are duplicates with no references to them. In the screenshot you can see there are only references to the uppercase file and the two files which are identical according to their checksum.

image

cwarnermm:
Thanks, @mrckndt. I was initially concerned because my Mac was only showing me a single file for the ``discordDarkTheme.png`` file (due to Mac being case insensitive where I was assuming it would behave closer to Linux). Looking at the docs repository, I can see the duplicates clearly. Thanks so much for this clean up effort! Much appreciated!

cwarnermm (approved):
Thank you!

PR #5325, opened 2021-12-15
Update working-with-boards.rst

justinegeffen:
Added Calendar view content.

@wuwinson, I'm not sure how much detail to add around the default sorting vs. custom properties and how it affects moving cards around. If it's something that users struggle with then happy to add more details but don't want to muddy the waters. LMK if I left anything out (from the demo it seems like I included everything).

### DIFF
--- a/source/boards/working-with-boards.rst
+++ b/source/boards/working-with-boards.rst
@@ -47,7 +47,7 @@ Calendar view
 
 To use this view, cards need to have the **Date** property added.
 
-If cards don't have a **Date** property, they'll be sorted and displayed by the card creation date (default). These cards can't be moved around the board until a **Date** property is added. This property can be a due date, milestone, and so on.
+If cards don't have a custom **Date** property, they'll be sorted and displayed by the card creation date (default). These cards can't be moved around the board until a custom **Date** property is added.
 
 .. note:: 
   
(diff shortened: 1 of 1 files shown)


source/boards/working-with-boards.rst
```diff
@@ -35,6 +35,33 @@ All changes you make to boards and cards are saved immediately.
...
+~~~~~~~~~~
+
+This is a vertical list view where cards are sorted by columns. The column names are editable and you can drag cards between columns.
+
+Calendar view
+~~~~~~~~~~~~~
+
+To use this view, cards need to have the **Date** property added.
+
+If cards don't have a **Date** property, they'll be sorted and displayed by the card creation date (default). These cards can't be moved around the board until a **Date** property is added. This property can be a due date, milestone, and so on.
```
justinegeffen, line 50:
```suggestion
If cards don't have a custom **Date** property, they'll be sorted and displayed by the card creation date (default). These cards can't be moved around the board until a custom **Date** property is added.
```

wuwinson:
> Added Calendar view content.
> 
> @wuwinson, I'm not sure how much detail to add around the default sorting vs. custom properties and how it affects moving cards around. If it's something that users struggle with then happy to add more details but don't want to muddy the waters. LMK if I left anything out (from the demo it seems like I included everything).

Agreed, what you have there looks great!

wuwinson (approved):

cwarnermm (approved):

PR #5341, opened 2021-12-28
Clarifying the functionality

icelander:
#### Summary

The original language could cause people to think this is only for corporate exports and does not apply to all Slack exports, regardless of plan.


cwarnermm (approved):
Thank you!

PR #5344, opened 2022-01-03
v6.3 docs + upgrade notes

amyblais:
Let me know if anything is missing from the important upgrade notes section for v6.3 https://github.com/mattermost/docs/pull/5344/files#diff-eba99c731b9e0b65d9912ebdbfe52962ffc5680007ff399d14a7f8d5ddb5de6f.


agnivade (approved):

ashishbhate (approved):

catalintomai (approved):

streamer45 (approved):

cwarnermm:
@amyblais Would you be open to updating the product documentation copyright notice as part of this PR? It's set via the ``conf.py`` file.

cwarnermm (approved):
Thanks!

amyblais:
> @amyblais Would you be open to updating the product documentation copyright notice as part of this PR? It's set via the conf.py file.

@cwarnermm I added a commit now, let me know if I missed anything.

cwarnermm:
Thanks, @amyblais! The latest preview looks great!

source/upgrade/important-upgrade-notes.rst
```diff
@@ -14,14 +14,18 @@ Important Upgrade Notes
...
 
 +----------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | If you’re upgrading from a version earlier than... | Then...                                                                                                                                                          |
 +====================================================+==================================================================================================================================================================+
+| v6.3                                               | An asynchronous job was added to fix an issue with collapsed reply threads that resulted in unexpected unread channels. The job may result in an increase in     |
```
esethna, line 25: Tweaking the wording for this: @ashishbhate @cwarnermm can you please review?

v6.3 adds an asynchronous job to fix an issue that resulted in channels with join/leave messages being unexpectedly marked as unread when enabling Collapsed Reply Threads. After the upgrade, the job is started automatically when the server restarts and the job executes in the background so the Mattermost server will not experience downtime. We recommend upgrading during off-peak hours as you may experience higher than average server and database resource utilization while the job is in progress. In our testing, the job takes approximately 3 hours per ~4 million rows in the `ChannelMembership` table using a PostgreSQL database.

You may check the progress of the job by executing this database query: `select * from Jobs where type = 'fix_channel_unreads_for_crt' order by lastactivityat desc;`. The data column will display the number of channel memberships fixed (BadChannelMembershipsFixed), the number of channel memberships checked (TotalChannelMembershipsChecked), and the current Channel and User Id pair that is being processed. 

If you are experiencing severely high server and database resource utilization, you may kill the job by executing `XXXXXX` < @ashishbhate 

Once the job is complete, it will not run again and resource usage will return to average levels.

esethna (changes requested):

source/upgrade/important-upgrade-notes.rst
esethna: @ashishbhate @agnivade @streamer45 do we have any idea what the expected increase in DB and server load will be based on when it was run in community? From the admin perspective I'm sure they'll want to know what to expect if this job will be running along side the server. Especially considering all the performance incidents recently

esethna: @ashishbhate I know the time estimate is rough, but do we also have an idea of the when the admin should "sound the alarm" if this takes over X hours / Y rows?

ashishbhate: >If you are experiencing severely high server and database resource utilization, you may kill the job by executing 


First update the Systems table to stop the job from restarting
`INSERT INTO Systems VALUES ('fix_crt_channel_unreads', true);`

Then restart the mattermost application to kill the existing job.


>Do we have any idea what the expected increase in DB and server load will be? Perhaps we can use the datapoint from when it was run in community? From the admin perspective I'm sure they'll want to know what to expect if this job will be running along side the server - especially considering all the performance incidents recently

>I know the time estimate is rough, but do we also have an idea of the when the admin should "sound the alarm" if this takes over X hours / Y rows?


Its hard to measure this because the job depends on a lot of factors like number of channelmember entries, how many are broken and need fixing. For fixing ChannelMembers its also needs to scan the posts table. The performance, and time taken also depends on the server load and specs. During high load times its possible the job will slow down as well. 
I don't think there is a need to "sound the alarm" unless the application performance starts to suffer, however slow the job may be proceeding.

>If the job is killed (or perhaps paused during high usage times), can it be restarted at the point it left off? If so what are the commands to do that?

Restarting the mattermost server will automatically reschedule the job to run after a few minutes, if the `fix_crt_channel_unreads` key is not set to `true` in the Systems table. If that key is set, then deleting it from the Systems table will reschedule the job after a few minutes.

esethna: Okay, thanks @ashishbhate. I updated the proposed text above. Please re-review and confirm it's accuracy?  https://github.com/mattermost/docs/pull/5344#discussion_r781660452

ashishbhate: This section will need to be removed. We've reverted the job from master and release-6.3.

PR #5346, opened 2022-01-04
Add example PNS payload for id-only notifications

sadohert:
A customer infosec team was requesting an example.  Publishing this same example here.


sadohert:
For reference, I did a `tcpdump` on my MPNS host to view the payload contents:

```
sudo tcpdump -i eth0 -s 0 -A 'tcp dst port 8066 and tcp[((tcp[12:1] & 0xf0) >> 2):4] = 0x504F5354'
```

justinegeffen (approved):
Thanks, Stu!

PR #5353, opened 2022-01-07
Move Enable Reliable Websockets config setting to GA

cwarnermm:
Documentation for: https://github.com/mattermost/mattermost-server/pull/19040

Moved **Enable Reliable Websockets** config setting to deprecated config settings page with a note that it's now generally available (and set to true for backwards compatibility with older clients).

### DIFF
--- a/source/configure/configuration-settings.rst
+++ b/source/configure/configuration-settings.rst
@@ -6906,19 +6906,6 @@ Audit settings
 
 The audit settings output audit records to syslog (local or remote server via TLS) and/or to a local file. Both are disabled by default. They can be enabled simultaneously.
 
-Enable Reliable Websockets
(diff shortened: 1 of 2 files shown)


cwarnermm:
@agnivade - The config setting ``EnableReliableWebsockets`` doesn't appear to be listed on the [Telemetry](https://docs.mattermost.com/manage/telemetry.html) docs page. Not sure if this is intentional or an oversight. Given that it's moving to GA (enabled for all), is this a piece of telemetry data we're tracking and should doc?

justinegeffen (approved):

agnivade:
@cwarnermm - The config setting has been removed now that it's moved to GA.

agnivade (approved):

PR #5358, opened 2022-01-10
Fix links in changelog table of contents

hmhealey:
The format of the anchors for the latest few releases are incorrect, so clicking on them doesn't do anything


cwarnermm (approved):
Thank you!

PR #5379, opened 2022-01-22
Update common-navigating-mattermost-products.rst

varghesejose2020:


justinegeffen:
Hi @varghesejose2020. Thanks for your contribution! The current documentation is correct and so this change will not be applied. We look forward to working with you going forward!

PR #5396, opened 2022-02-01
v6.3.3 / 6.2.3 / 6.1.3 / 5.37.8 docs for CRT fixes (and other dot release fixes)

amyblais:
Note: I'll add the release dates once ready, the texts for the CRT fixes can be reviewed.

Changelog:
 - The default for ``ThreadAutoFollow`` has been changed to ``false``. This does not affect existing configurations where this value is already set to ``true`` [MM-41351](https://mattermost.atlassian.net/browse/MM-41351). 
 - Collapsed Reply Threads will only be kept up to date if ``ThreadAutoFollow`` config is set to true [MM-41350](https://mattermost.atlassian.net/browse/MM-41350).

Important Upgrade Notes:
 - In version __, the default for ``ThreadAutoFollow`` has been changed to ``false``. This does not affect existing configurations where this value is already set to ``true``.


esethna (changes requested):
@amyblais one update:

> Collapsed Reply Threads will only be kept up to date if ``ThreadAutoFollow`` config is set to true [MM-41350](https://mattermost.atlassian.net/browse/MM-41350).

Change to: 

Prevented some instances where operations relating to Collapsed Reply Threads were adding load to the database server even when the ``ThreadAutoFollow`` and ``CollapsedThreads`` configs were disabled. [MM-41350](https://mattermost.atlassian.net/browse/MM-41350).

esethna (approved):

PR #5398, opened 2022-02-02
MMCTL updates

cwarnermm:
Incorporated community feedback to clarify mmctl usage (particularly for those customers familiar with running CLI commands). 
- Updated Mattermost Administration > Other Resources > mmctl Command Line Tool > Usage Notes to clarify when prefixing commands with `./` is necessary.

### DIFF
--- a/source/manage/mmctl-command-line-tool.rst
+++ b/source/manage/mmctl-command-line-tool.rst
@@ -74,7 +74,8 @@ mmctl usage notes
 - ``mmctl`` comes bundled with the Mattermost distribution, and is located in the ``bin`` folder of the installation, next to the ``CLI``.
 
   - We recommend you add the path to the Mattermost ``bin`` folder into your ``$PATH`` environment variable. This ensures that you can run mmctl commands locally regardless of your current directory location.
-  - If the ``bin`` folder is missing from your ``$PATH``, you must specify the full path to mmctl when running mmctl commands. Otherwise, you must be in the ``bin`` directory to run mmctl commands, and the commands must be prefixed with ``./``.
+  - The most convenient way to use mmctl is to add the path to the Mattermost ``bin`` directory to your ``$PATH`` environment variable. This ensures that you can run mmctl commands locally regardless of your current directory location.
+  - If the ``bin`` directory is not added to the ``$PATH`` environment variable, each time you use mmctl you must be in the ``bin`` directory to run mmctl commands, and the commands must be prefixed with ``./``. If you're working from a different directory, make sure you specify the full path to mmctl when running mmctl commands.
 - Parameters in CLI commands are order-specific.
 - If special characters (``!``, ``|``, ``(``, ``)``, ``\``, ``'``, and ``"``) are used, the entire argument needs to be surrounded by single quotes (e.g. ``-password 'mypassword!'``, or the individual characters need to be escaped out (e.g. ``password mypassword\!``).
 - Team name and channel name refer to the handles, not the display names. So in the URL ``https://community.mattermost.com/core/channels/town-square`` team name would be ``core`` and channel name would be ``town-square``.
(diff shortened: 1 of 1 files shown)


isacikgoz (approved):
Nice addition 👍

source/manage/mmctl-command-line-tool.rst
```diff
@@ -72,6 +72,9 @@ mmctl usage notes
 
 - System Admins have two ways to run ``mmctl`` commands: by downloading ``mmctl`` from the repository, or by building it directly. See the `mmctl readme <https://github.com/mattermost/mmctl#install>`__ for details.
 - ``mmctl`` comes bundled with the Mattermost distribution, and is located in the ``bin`` folder of the installation, next to the ``CLI``.
+
+  - We recommend you add the path to the Mattermost ``bin`` folder into your ``$PATH`` environment variable. This ensures that you can run mmctl commands locally regardless of your current directory location.
+  - If the ``bin`` folder is missing from your ``$PATH``, you must specify the full path to mmctl when running mmctl commands. Otherwise, you must be in the ``bin`` directory to run mmctl commands, and the commands must be prefixed with ``./``.
```
justinegeffen, line 77: Some questions about this:
1. How would the ``bin`` folder be missing? Would it be missing in general or missing from the env. variable? Can it be added? If it can be, is this step needed?
2. So if the ``bin`` folder is missing, the user needs to be in the ``bin`` directory to run commands? 

If my assumptions are correct, would this work?

- We recommend you add the path to the Mattermost ``bin`` folder into your ``$PATH`` environment variable. This ensures that you can run mmctl commands locally regardless of your current directory location. If the ``bin`` folder isn't added, you'll need to use the full path when running mmctl commands. 
- If the path isn't added, you will need to be in the ``bin`` directory to run mmctl commands and they must be prefixed with ``/``.

LMKWYT. :)

justinegeffen (commented):
I've added some feedback and a suggestion. :)

source/manage/mmctl-command-line-tool.rst
cwarnermm: Good questions, @justinegeffen. Some programs and apps automatically add path details to the user's $PATH env variable, and others don't. I'm not sure if Mattermost attempts to do so. Not all users will be starting with mmctl from the same tool setup and with the same platform-level knowledge. And this isn't something we'll necessarily want to be prescriptive about because $PATH management is a user-centric choice.

This PR is the result of a recent exchange on the Community Mattermost server, where a recently upgraded user was struggling to run mmctl commands, having been a CLI user previously. The missing piece of the puzzle was prefixing ``./`` to the mmctl commands (and knowing to review the $PATH).

If the Mattermost path is missing from the user's $PATH environment variable, the user needs to run the mmctl command using different syntax depending on where they are at the time they try to run the command.

- The first proposed bullet is a firm recommendation to add the Mattermost path to the $PATH env variable. 
- The second proposed bullet tells the user what they need to do if they choose not to follow the recommendation: either specify full path for all commands, or be in the ``bin`` folder and prefix all commands with ``./``.

Given this context, are there changes we could make to simplify or clarify the docs to support a wider group of audiences?

justinegeffen: Thanks for the context, that makes a lot of sense. So the user can take either option but the first is the golden path because then it "just works"?

If so, I recommend this wording:

```suggestion
  - The most convenient way to use mmctl is to add the path to the Mattermost ``bin`` folder to your ``$PATH`` environment variable. This ensures that you can run mmctl commands locally regardless of your current directory location.
  - If the ``bin`` directory is not added to the ``$PATH`` environment variable, each time you use mmctl you must be in the ``bin`` directory to run mmctl commands, and the commands must be prefixed with ``./``. If you're working from a different directory, make sure you specify the full path to mmctl when running mmctl commands.
```

justinegeffen (approved):

PR #5400, opened 2022-02-02
Hide default emojis from a Mattermost self-hosted deployment

cwarnermm:
Added admin documentation on how to hide default emojis from the Emoji Picker for self-hosted deployments based on @M-ZubairAhmed's excellent draft documentation.

### DIFF
--- a/source/manage/hide-emojis.rst
+++ b/source/manage/hide-emojis.rst
@@ -1,0 +1,40 @@
+Hide emojis from the Emoji Picker
+=================================
+
+|all-plans| |self-hosted|
+
+.. |all-plans| image:: ../images/all-plans-badge.png
+  :scale: 30
+  :target: https://mattermost.com/pricing
+  :alt: Available in Mattermost Free and Starter subscription plans.
+
+.. |self-hosted| image:: ../images/self-hosted-badge.png
+  :scale: 30
+  :target: https://mattermost.com/deploy
+  :alt: Available for Mattermost Self-Hosted deployments.
+
+Emojis are small, digital images or icons used to communicate or express concepts such as emotions and physical gestures in messages. In addition to the default set of emojis that ship with Mattermost, users can supplement Mattermost standard emojis with by `uploading custom emojis <https://docs.mattermost.com/messaging/using-emoji.html#creating-custom-emojis>`__. 
+
+To hide a standard emoji in Mattermost, a System Admin must identify each emoji offending by name, and then rebuild their self-hosted instance of the Mattermost webapp to hide the emojis.
+
+.. note:: 
+
+    - The ability to remove default emojis from Cloud deployments is not supported.
+    - Removing a standard emoji from the Mattermost web app also removes it from all Mattermost clients, including the Desktop App and mobile apps. However, a hidden emoji doesn't stop users from using unicode characters to apply a hidden emoji visible to other Mattermost users.
+    - Customers who upgrade Mattermost `using the standard upgrade process <https://docs.mattermost.com/guides/deployment.html#upgrade-mattermost>`__ will see the offending emoji return following the upgrade. This is because the standard upgrade process retains only plugin data in ``mattermost/client/plugins``, and doesn't retain emoji data.
+
+To remove a standard emoji from the Mattermost Emoji Picker:
+
+1. Clone the `mattermost-webapp <https://github.com/mattermost/mattermost-webapp>`__ repository by running: ``git clone --depth=1 https://github.com/mattermost/mattermost-webapp``
+2. In Mattermost Channels, identify the names of the emojis to remove by hovering over over an emoji in the Emoji Picker, and noting its name in the format ``:name:``. For example: ``:smile:``.
+3. Create a ``TXT`` file, such as ``excludedemojis.txt`` that contains a list of the emojis to remove. Add each emoji name as a separate line.
+4. Navigate to the ``mattermost-webapp`` directory from step 1 by running: ``cd mattermost-webapp``.
+5. Install ``nvm`` by following `these instructions <https://github.com/nvm-sh/nvm#installing-and-updating>`__.
+6. Install the correct node version for the webapp by running ``nvm i``.
+7. Download the node packages in the webapp by running ``npm i``.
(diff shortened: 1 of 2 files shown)


enahum:
> Hide Emojis from the Emoji Picker

I really don't know if we should document this, there is no way (as of today) to "hide" emojis on mobile, thus this leading to create false expectations when users can still add this emojis in messages & reactions from the mobile app.

source/manage/hide-emojis.rst
```diff
@@ -0,0 +1,40 @@
...
+  :alt: Available for Mattermost Self-Hosted deployments.
+
+Emojis are small, digital images or icons used to communicate or express concepts such as emotions and physical gestures in messages. In addition to the default set of emojis that ship with Mattermost, users can supplement Mattermost standard emojis with by `uploading custom emojis <https://docs.mattermost.com/messaging/using-emoji.html#creating-custom-emojis>`__. 
+
+To hide a standard emoji in Mattermost, a System Admin must identify each emoji offending by name, and then rebuild their self-hosted instance of the Mattermost webapp to hide the emojis.
+
+.. note:: 
+
+    - The ability to remove default emojis from Cloud deployments is not supported.
+    - Removing a standard emoji from the Mattermost web app also removes it from all Mattermost clients, including the Desktop App and mobile apps. However, a hidden emoji doesn't stop users from using unicode characters to apply a hidden emoji visible to other Mattermost users.
```
enahum, line 23: how does this happen? have anyone done this successfully?

cwarnermm: @coltoneshaw - I recall you taking @M-ZubairAhmed's draft content for a spin. Did you observe how mobile apps would behave following this rebuild operation?

cwarnermm:
@esethna - Please see @enahum's concerns and reservations about documenting this functionality.

source/manage/hide-emojis.rst
coltoneshaw: It's not removed from the mobile app, just the desktop and web app. I _think_ it's also removed from the reactions on the mobile app.

enahum: mobile reactions are still present as well + the fact that all emojis are converted to unicode instead of using an image

source/manage/hide-emojis.rst
```diff
@@ -0,0 +1,40 @@
...
+2. In Mattermost Channels, identify the names of the emojis to remove by hovering over over an emoji in the Emoji Picker, and noting its name in the format ``:name:``. For example: ``:smile:``.
+3. Create a ``TXT`` file, such as ``excludedemojis.txt`` that contains a list of the emojis to remove. Add each emoji name as a separate line.
+4. Navigate to the ``mattermost-webapp`` directory from step 1 by running: ``cd mattermost-webapp``.
+5. Install ``nvm`` by following `these instructions <https://github.com/nvm-sh/nvm#installing-and-updating>`__.
+6. Install the correct node version for the webapp by running ``nvm i``.
+7. Download the node packages in the webapp by running ``npm i``.
+8. Use ``npm`` to invoke the script ``emoji_gen.mjs``. For example: ``npm run make-emojis -- --excluded-emoji-file .path-to-file/excludedemojis.txt``
+9. Build and package the modified mattermost-webapp by running ``make package``.
+10. Remove the existing client directory within the Mattermost install directory by running ``rm -r /opt/mattermost/client``.
+11. Extract the generated tar into the Mattermost install directory by running ``tar -xvf mattermost-webapp.tar.gz /opt/mattermost``.
```
coltoneshaw, line 38: My fault, the command would be `sudo tar -xvf mattermost-webapp.tar.gz -C /opt/mattermost`

source/manage/hide-emojis.rst
coltoneshaw: Confirmed what @enahum said above. When running this script it completely removes it from anything rendering our webapp, but does not remove it from the mobile app itself. 

You _can_ still insert this using the emoji keyboard on your desktop OS though. Example, on macOS i can copy 🤫 and paste it into Mattermost, which allows it to render.
Screen Shot 2022-02-02 at 4 42 22 PM

coltoneshaw: View of the same interaction from the mobile client. 

rn_image_picker_lib_temp_7b2ffbda-2557-4e72-a7ce-afca01874fde
rn_image_picker_lib_temp_0e57eeb0-d2b8-4c14-ba2a-93b39cbfae81

enahum:
I'm going to list a few things here to help understanding the extent of this:

1.  Mobile has it's own emoji.json file with all available built-in emojis
2. Mobile converts shortcodes `:emoji:` to unicode for both emojis in messages and reactions
3. Mobile treats emojis as images only for custom emojis
4. Mobile allows the user to pick an emoji from the OS keyboard (no control over this).
image

With those items listed what does it mean for someone who wants to remove an emoji from mobile
*. Edit the mobile app `emoji.json` file and remove the ones they want to exclude
*. Build their own mobile app
*. Deploy their own Push Proxy

this is a huge burden just to want to remove a few emojis or maybe even one, then no control over the OS built-in ones in the keyboard, which is the same for desktop (at least on macOS) if the OS Emoji picker is used
image

So in short, the reality is that one way or another those "hidden" emojis can be actually used.

coltoneshaw:
So, @enahum , it sounds like we almost need something built into the product like our text replace plug-in that just removes the offending emojis if they're ever used. Then also include a way to hide them from the auto fill within Mattermost.

In its current form our suggestions are hacky at best?

Is that a fair statement?

enahum:
> So, @enahum , it sounds like we almost need something built into the product like our text replace plug-in that just removes the offending emojis if they're ever used. Then also include a way to hide them from the auto fill within Mattermost.
> 
> In its current form our suggestions are hacky at best?
> 
> Is that a fair statement?

yes, I would say they are hacky at best and thus can be lead to false expectations and complaints. We should look into a built-in solution that would also disallow the use of certain unicode emojis so that our clients are able to dismiss the input

source/manage/hide-emojis.rst
M-ZubairAhmed: Hey @coltoneshaw @cwarnermm let's wait for this https://github.com/mattermost/mattermost-webapp/pull/9725 to get merged which makes it easy to add server side emoji file. Once that is merged I can update this pr to include those steps as well.

cwarnermm: Thanks, @M-ZubairAhmed!

PR #5403, opened 2022-02-04
Redirects Cleanup

cwarnermm:
The list of redirects for docs.mattermost.com were becoming difficult to manage and maintain, contained duplicate entires causing issues, and entries were no longer in a format the web server could interpret and return correctly. Any redirect missing the file extension HTML in the origin path was taking users to the home page rather than the expected destination page.

This PR:
- cleans up the redirects list
- removes duplicate entries
- adds the missing file extension for all origin URLs required by the web server
- organizes all redirects alphabetically based on the origin URL path
- introduces inline code comments to provide guidance for future contributors looking to add new redirects or manage existing redirects

Existing redirects have been randomly tested locally to ensure they continue to redirect as expected.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -61,770 +61,749 @@ source_suffix = [".rst", ".md"]
 
 # Redirects using: https://pypi.org/project/sphinx-reredirects/
 redirects = {
-    "integrations/jira": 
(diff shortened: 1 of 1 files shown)


justinegeffen (approved):
Awesome! Thank you!!

Issue #5362, opened 2022-01-11
Add install docs for Ubuntu 22.04


svelle:
Ubuntu 22.04 will release in a couple of months, beta release is imminent, we should get ahead of the curve and start working on install docs for this *before* the release so we got them ready when it comes out. Happy to assist where I an :)

cwarnermm:
Thanks, @svelle! Is the Support team open to providing the specific install details needed for this platform?

PR #5408, opened 2022-02-08
Ubuntu 22 Install docs.

coltoneshaw:
Summary
Updated docs to include Ubuntu 22.04 LTS. Additionally, fixed an outdated link in 2004 ubuntu postgres docs. 

Validated the docs locally on a VM for both postgres and mysql, no issues.

**Do not publish / merge until April 21, 2022**.

#### Ticket Link
Fixes https://github.com/mattermost/docs/issues/5362


source/install/install-ubuntu-2204-server.rst
```diff
@@ -0,0 +1,18 @@
+.. _install-ubuntu-2204-server:
+
+Installing Ubuntu Server 22.04 LTS
+----------------------------------
+
+Install the 64-bit version of Ubuntu Server on each machine that hosts one or more of the components.
+
+**To install Ubuntu Server 22.04:**
+
+1. To install Ubuntu Server 22.04, see the `Ubuntu Installation Guide. <https://help.ubuntu.com/22.04/installation-guide/amd64/index.html>`__
```
coltoneshaw, line 10: Docs aren't published by Ubuntuy yet, so need to get this link validated once they are.

source/install/install-ubuntu-2204-postgresql.rst
```diff
@@ -0,0 +1,109 @@
...
+
+**To install PostgreSQL on Ubuntu Server 22.04:**
+
+1. Log in to the server that will host the database and issue the following command:
+
+  ``sudo apt install postgresql postgresql-contrib``
+
+  When the installation is complete, the PostgreSQL server is running, and a Linux user account called *postgres* has been created.
+  
+By default on Ubuntu 22.04 this should be Postgres 14, however, Confirm this version is PostgreSQL version 10 or higher. You can do this by running ``psql -V``. If this isn't version 10 or higher you can reference the PostgreSQL documentation `here <https://www.postgresql.org/download/linux/ubuntu/>`__ to install an updated version of PostgreSQL.
```
coltoneshaw, line 18: We may be able to take out this line all together. Ubuntu 22.04 defaults to Postgres 14+, so there shouldn't be a need to validate the version anymore.

svelle: I think we should probably just take it out altogether.

coltoneshaw: I figure by the time we release this they'll have added the docs for it. What do you think?

nab-77 (approved):
All reads well. I do wonder, when thinking about support matrix, if we have an opportunity to streamline by cross-referencing content (e.g. db setup section without specific host info).

coltoneshaw:
> All reads well. I do wonder, when thinking about support matrix, if we have an opportunity to streamline by cross-referencing content (e.g. db setup section without specific host info).

I agree, it would clean up a ton of the pages and duplicated content right now. Most of the instructions _appear_ to be the same. In the past for RHEL docs we would simply point out when a command is different on a specific version and keep everything together. 

I don't know if that something we want to tackle in this PR or know for future?

cwarnermm:
I agree with you both that streamlining and simplifying our install docs requires us to start pulling back on the level of third-party software support we provide in the product documentation. Every time we include these details, it's information that must be maintained over time, otherwise we risk frustrating our users with inaccurate instructions.

@coltoneshaw - You seriously did a beautiful job of following our current install docs approach. Well done! 
@nab-77 - What is your recommendation on how to move forward with this PR specifically, given that we all agree that we need to scale back the third-party support content? 

My recommendation: Since we need to wait for the release of 22.04 to publish this update, we don't need an immediate decision. I'm inclined to publish this great content that Colton was kind enough to develop as-is, with an understanding that we'll revisit this content later this year with a streamlining goal in mind.

cwarnermm:
@coltoneshaw - Yes! Looks like 22.04 is now released, so we should finalize this PR.
@nab-77 - Given our intention to continue simplifying our installation instructions overall, how would you like to proceed with this PR specifically?

cwarnermm:
Thanks, @coltoneshaw! To merge this PR, we need to address the one Ubuntu installation docs link (that doesn't yet exist).

coltoneshaw:
@cwarnermm - You just mean adding the link? I'm not sure i'm following

cwarnermm:
Hi @coltoneshaw! I'm unable to find a link to the 22.04 installation guide. The URL referenced in these docs ``https://help.ubuntu.com/22.04/installation-guide/amd64/index.html`` doesn't resolve as-is (though the 20.04 version is available). Is there an alternate resource we could point admins to for installation guidance instead?

navjotjsingh:
The command `GRANT ALTER, CREATE, DELETE, DROP, INDEX, INSERT, SELECT, UPDATE, REFERENCES ON mattermost.* TO 'mmuser'@'%';` doesn't work for me.

I had to resort to using `ALL`. I guess some privileges are still missing from the above command used in Mattermost 7.0.2 and up.

This also needs to be updated for other guides as well.

source/install/install-ubuntu-2204-server.rst
coltoneshaw: Just reviewing this again, it's been moved per the text on https://help.ubuntu.com/

> As of the Ubuntu LTS release in 2020, the server documentation has moved to[ a different site](https://ubuntu.com/server/docs), and will automatically update when changes are made to the [discourse source code pages](https://discourse.ubuntu.com/t/introduction/11322).

So, it looks like https://ubuntu.com/server/docs/installation might be the best link. I can update it with this if we'd want. I know we are moving away from OS specific instructions a bit too, so i can close this PR down if needed.

justinegeffen:
@cwarnermm, could we merge this as-is or does it need more review?

cwarnermm:
There are 2 things outstanding for this PR:
- merge conflicts need to be resolved (@coltoneshaw)
- the Ubuntu v22.04 installation guide documentation needs to be published.

coltoneshaw:
@cwarnermm - I don't think we'll get the installation guide, they haven't posted it yet (a year later). I also now question if we even need to link to installing an OS. That feels like a really low level step, the sysadmin should know how to do... but can also google really easily. The docs we previously linked to were **long** and honestly, not super useful for the target audience of the link (people who don't know already how to install an OS). Take a look yourself and see if in 5 minutes you can find how to install Ubuntu 20.04 - https://help.ubuntu.com/20.04/installation-guide/amd64/install.en.pdf.  I know I couldn't.

Second, I haven't touched this since our past talks, docs like this will be deprecated. I'm happy to update anything left.

cwarnermm:
@coltoneshaw - I agree with you that our install docs content isn't meeting our target audiences where they are yet. 

I'm going to close this docs PR in favor of the in-progress platform-based installation approach, where a given platform page clearly notes the versions that Mattermost supports.

PR #5409, opened 2022-02-09
Updated docs links to legacy domains

cwarnermm:


cwarnermm:
@emdecr - Quick question before we approve and merge this docs PR with updated link URLs: Do any of the link updates for ``about.mattermost.com`` need to be changed in this PR? I'm thinking all updated URLs are good to go, but I wanted to double-check.

justinegeffen (approved):

emdecr:
@cwarnermm Looks good to me and thanks for checking. 
Also to clarify – my main worries for the `about.mm.com` links are for product because they have that future-proofing setup, but in this case, where we have more control over everything (and don't have to worry about users seeing different versions [like the detail about with how existing servers wouldn't get defaults updated]), I think we're good 👍

PR #5426, opened 2022-02-16
Correct socked based db creation

alerque:
Code labeled "create the database" was actually just a copy/paste of the
"create a user" blob. This is the actual createdb command including
setting the owner of the database to the right user.

Additionally the test code to connect to it does not need (and even
should not have) the username. The example uses peer authentication and
if it is going to work from the mattermost daemon it needs to work
without specifying the user based on the system user ID.


eosophob1a:
Tested this and looks good to me.

cwarnermm (approved):
Thank you, @alerque, for your great documentation correction PR!

PR #5427, opened 2022-02-17
Update self-managed-changelog.md

wiersgallak:
Adding a note about channel moderation replacing townsquare read-only setting which was deprecated


wiersgallak:
@amyblais I am rusty!  Thank you for fixing it!!

PR #5452, opened 2022-03-04
Fix http links pointing to mm.com uploads

emdecr:
Replacing http links with secure https replacements. URLs point to images on mattermost.com.

cc @cwarnermm 

TY!


cwarnermm (approved):
Thank you!

PR #5460, opened 2022-03-07
Clarify allowOtherServers behavior

icelander:
#### Summary

Clarifies that if `allowOtherServers` is true, the app will automatically connect to the `serverUrl`


cwarnermm (approved):
Thank you, @icelander!

PR #5499, opened 2022-03-28
Removed invite tip

cwarnermm:
The ability to invite members of a custom user group via the ``/invite`` slash command is not supported. Removed the tip featuring this use case.


BenCookie95 (approved):

laneycs (approved):
[👍🏽](https://emojipedia.org/thumbs-up-medium-skin-tone/) Passed PM review.

PR #5533, opened 2022-04-05
Adding links and clarifying support isn't free

it33:
Need to link to specific sections in troubleshooting guide


it33:
Also, support is only for paid subscriptions, "Premier Support" is an add-on to paid subscriptions, it's not the entitlement to support,

justinegeffen (approved):

PR #5542, opened 2022-04-05
fix mac link shortcut

coltoneshaw:
Showed `alt` as the mac option, doesn't exist on a standard mac keyboard.


cwarnermm (approved):
Thanks, @coltoneshaw!

PR #5547, opened 2022-04-07
Update channels-settings.rst

justinegeffen:
- Heading case standardization (it does differ from the case used in System Console but this will be aligned soon)
- Fixed some capitalization
- Fixed some formatting

Question: Should we use "System Message" or "system message"? And should @-mention mentions use backticks or italics?


source/channels/channels-settings.rst
```diff
@@ -128,11 +128,11 @@ Select **Theme Colors** to select from four standard themes designed by the Matt
 
 Select **Custom Theme** to customize your theme colors and share them with others by copying and pasting theme vectors into the input box. Observe a live preview as you customize theme colors, then select **Save** to confirm your changes. Discard your changes by selecting **Cancel**, or by exiting the settings modal and selecting **Yes, Discard**.
 
-In Enterprise Edition, if you belong to multiple teams, you can optionally select the checkbox "Apply new theme to all my teams" to have the theme show up across teams. Otherwise, the changes will only apply to the current team.
+In Enterprise editions, if you belong to multiple teams, you can optionally select the checkbox **Apply new theme to all my teams** to have the theme show up across teams. Otherwise, the changes will only apply to the current team.
```
cwarnermm, line 131:
```suggestion
In Enterprise edition, if you belong to multiple teams, you can optionally select the checkbox **Apply new theme to all my teams** to have the theme show up across teams. Otherwise, the changes will only apply to the current team.
```

cwarnermm:
Great update! I've left one comment. "system message" (lowercase) works well since it's not a formal name/noun. I agree that we need to be consistent when referencing @mentions; are they "@-mentions", "@mentions", or something else? I'm open to how we format them; however, once we align on the term we'll use, we may not need additional formatting such as italics or code.

cwarnermm (approved):

PR #5563, opened 2022-04-19
Update Important Upgrade Notes

amyblais:
Remove line about MariaDB support. The original text was related to the new migration system, but we don't officially support MariaDB anyway so the text is confusing.


cwarnermm (approved):
Thanks, @amyblais!

agnivade (approved):
Looks great. Thanks Amy

PR #5565, opened 2022-04-19
Update desktop-app-changelog.rst

amyblais:


source/install/desktop-app-changelog.rst
```diff
@@ -20,12 +20,99 @@ Desktop application changelog
...
+Linux
+^^^^^^
+
+- Updated the Linux closing behaviour to allow the app to close complely when pressing ``X``.
+- Changed the default setting for **Leave app running in notification area when application window is closed** on Linux to ``false`` by default.
+
+All Platforms
+^^^^^^^^^^^^^
+
+- Added support for Calls for the Desktop App.
```
streamer45, line 64: Support for Calls has been in Desktop since the plugin existed as it didn't need special changes on this side. What was added in v5.1 is the ability for the user to select which window to share when screensharing. Previously users could only share the whole screen (including any secondary desktop).

streamer45 (approved):
Looks good :+1:

source/install/desktop-app-changelog.rst
```diff
@@ -20,12 +20,99 @@ Desktop application changelog
...
+
+Breaking Changes / Upgrade Notes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- 
+
+Highlights
+~~~~~~~~~~~~~~~
+
+- Added a Desktop App auto-updater. The app now automatically checks for new updates on app start up.
```
esethna, line 50: Add link to docs from Carrie

source/install/desktop-app-changelog.rst
```diff
@@ -20,12 +20,99 @@ Desktop application changelog
...
+
+Improvements
+~~~~~~~~~~~~~~~
```
esethna, line 53: Mac: Mattermost can now be installed on the Mac App Store (link). Even if you’re already using Mattermost desktop on Mac, you can download and install it via the Mac App Store to access future automatic updates.

devinbinnie: We are still providing Mac builds on GitHub, so we should find a way to communicate that while those are fine to install, they do not support auto-updates.

esethna (approved):

devinbinnie (approved):

PR #5567, opened 2022-04-20
Search results cleanup - Part 1

cwarnermm:
Search results cleanup includes:
- removing duplicate entries (one for Cloud vs one for self-hosted where unnecessary)
- removing contents that exist only to be included in other source files
- updating include directives with an attribute that ignores :nosearch: arguments for hidden pages
- archiving legacy content that should no longer be returned in search results


cwarnermm:
For the Mattermost Handbook:

We're hiding content from search that meet the following criteria:
- any source file whose contents are only included as part of another docs page, and is not accessible from the LHS directly
- any source file whose contents are not officially supported (moving to the forum)
- when any source file is archived (moved to archive folder)

To hide pages from search results:
- Add ``:nosearch:`` to the first line of an RST source file to hide it from search results. Note that ``:nosearch:`` is not supported in MD files. 
- For source files included in other files, ensure the include directive specifies the attribute: ``:start-after: :nosearch:``. If this is omitted, the text "nosearch" displays on the docs page. 

For example:

```
.. include .. /directory/filename.rst
  :start-after: :nosearch:
```

justinegeffen (approved):

PR #5571, opened 2022-04-22
Update extended-support-release.rst

amyblais:


wiersgallak (approved):
Thanks!

source/upgrade/extended-support-release.rst
```diff
@@ -101,3 +101,12 @@ For a new upcoming ESR, we send out an email announcement on or close to release
 For a deprecated ESR, we send out an email announcement three months in advance. We also add reminders on our release announcements, changelogs, `important upgrade notes <https://docs.mattermost.com/upgrade/important-upgrade-notes.html>`_, and our `Forum site <https://forum.mattermost.com/>`_.
 
 To receive updates about Extended Support Releases, sign up for our mailing list `here <https://eepurl.com/dCKn2P>`_.
+
+Desktop and Mobile App Compatibility
+---------------------------------------------
+
+Earlier 4.x versions of Mattermost Desktop App and earlier v1.x versions of the Mobile App are backwards compatible and are supported with our supported Extended Support Releases. However, for an optimal user experience and for latest security fixes, we strongly recommend upgrading both your Mattermost Desktop and Mobile Apps to the latest version.
```
esethna, line 108: I'll let @nab-77 comment here but I think we should think about this from the customer perspective. Maybe these statements paraphrased from a customer recently will help in writing this section of the doc:

"We are planning to upgrade our server to the next ESR, what mobile and desktop versions are compatible?"
"If we upgrade mobile or desktop before we upgrade the server, will we have compatibility issues?"

amyblais: Made some tweaks.

esethna:
Will defer to @nab-77 on reviewing this, thanks!

source/upgrade/extended-support-release.rst
nab-77: It would be right for us to be explicit here, for example when talking to desktop v4.x and mobile v1.x and being backwards compatible, one would expect that we call a minimum version of both. Since we have a support matrix built against general plus extended, with extended as our "long pole" can I suggest we put a marker down?

Using the Mattermost Server 5.x ESR as a date, can we fix on the desktop and mobile client versions supported at that time (@amyblais  I look to you for exact versions). Then this is a fixed minimum version for both based on the older server version we support. I would also call out the server ESR versions explicitly. This will become the start of our support matrix.

Now that is not to say we will not support older/different configurations, but we need this flag in the sand.

amyblais: Since we don't normally backport fixes for previous mobile and desktop app releases, I assume the latest mobile and desktop app releases are the minimum supported version for each self-hosted ESR version.

For example, for v6.3 ESR (January 16, 2022 release), the minimum versions would be:
 - Desktop app v5.0 (October 13, 2021 release)
 - Mobile app v1.48.0 (November 16, 2021 release) (the next release was on February 16, 2022)

I'll add @enahum and @devinbinnie in case they have any other thoughts that I should take into consideration.

https://docs.mattermost.com/install/self-managed-changelog.html
https://docs.mattermost.com/install/desktop-app-changelog.html
https://docs.mattermost.com/deploy/mobile-app-changelog.html

devinbinnie: I would definitely recommend the latest possible Desktop App for all users using v6.0 or higher. The only major incompatibility we have is with Boards, where we have some issues with subpaths with v5.0+, but those should be patched in the newest versions of Boards, so it's more important to upgrade those plugins.

Otherwise, I think the above is fine.

devinbinnie (approved):

amyblais:
@nab-77 Kind reminder to review this, or we can also discuss this further next week as needed.

nab-77:
In the new table its not needed to state "Desktop app v5.0" as the col header describes this. Just go with the version in the cell (e.g. 5.0)

source/upgrade/extended-support-release.rst
```diff
@@ -101,3 +101,25 @@ For a new upcoming ESR, we send out an email announcement on or close to release
...
+
+Please review the `Desktop App changelog <https://docs.mattermost.com/install/desktop-app-changelog.html>`_ and the `Mobile App changelog <https://docs.mattermost.com/deploy/mobile-app-changelog.html>`_ notes for any self-hosted version requirements for features and functionalities, as well as notes on security fixes.
+
++-------------+------------------+-----------------------------------------+----------------------------------------+
+| ESR Version | Release Date     | Desktop App Minimum Supported Version   | Mobile App Minimum Supported Version   |
++=============+==================+=========================================+========================================+
+| 6.3         | January 16, 2022 | Desktop app v5.0                        | Mobile app v1.48.0                     |
```
nab-77, line 120: Just 5.0 and 1.48.0 in the cells.

source/upgrade/extended-support-release.rst
```diff
@@ -101,3 +101,25 @@ For a new upcoming ESR, we send out an email announcement on or close to release
...
++-------------+------------------+-----------------------------------------+----------------------------------------+
+| 5.37        | July 16, 2021    | Desktop app v4.7                        | Mobile app v1.45.0                     |
```
nab-77, line 122: Just 4.7 and 1.45.0 in the cells.

nab-77 (approved):
If we can just clean up the cell contents. Very happy we call this out.

PR #5592, opened 2022-04-27
v6.6.1 Docs

amyblais:
Will add the release date when ready.


source/install/self-managed-changelog.md
```diff
@@ -13,7 +13,13 @@ Latest Mattermost Releases:
 
 ## Release v6.6 - [Feature Release](https://docs.mattermost.com/administration/release-definitions.html#feature-release)
 
-**Release Day: 2022-04-16**
+- **v6.6.1, released **
+  - Mattermost v6.6.1 contains a medium severity level security fix. [Upgrading](https://docs.mattermost.com/upgrade/upgrading-mattermost-server.html) to this release is recommended. Details will be posted on our [security updates page](https://mattermost.com/security-updates/) 30 days after release as per the [Mattermost Responsible Disclosure Policy](https://mattermost.com/security-vulnerability-report/).
+  - Replaced an expired GPG key which is used to verify the enterprise binary.
+  - Fixed an issue with null values in the OAuthApps table's MattermostAppID column, which was introduced in 6.6.0 [MM-43500](https://mattermost.atlassian.net/browse/MM-43500).
```
cwarnermm, line 19:
```suggestion
  - Fixed an issue with null values in the OAuthApps table's MattermostAppID column, which was introduced in v6.6.0 [MM-43500](https://mattermost.atlassian.net/browse/MM-43500).
```

source/install/self-managed-changelog.md
```diff
@@ -13,7 +13,13 @@ Latest Mattermost Releases:
...
+  - Fixed an issue where the Workspace Optimization dashboard mentioned that the workspace had reached over 100 users, when it only had 16 users [MM-43215](https://mattermost.atlassian.net/browse/MM-43215).
```
cwarnermm, line 20: Would it be fair to say that this bug would be present if a deployment had 15 or 17 users as well? If so, consider changing the specific value of 16 to something along the lines of "when fewer than 100 users were registered." or similar.

cwarnermm (commented):
Thanks, @amyblais! I've provided feedback inline.

cwarnermm (approved):
Thanks!

PR #5625, opened 2022-05-02
MM-43524 Add note about not unzipping Slack exports

hmhealey:
Along with https://github.com/mattermost/mmetl/pull/19, this will hopefully help us not spend more time debugging issues with Slack exports because the folder structure is wrong due to being rezipped

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43524


cwarnermm (approved):
Thanks, @hmhealey!

PR #5651, opened 2022-05-10
Fix Broken GPG Key location

ghost:
#### Summary
Our GPG key in Keybase is expired and currently we host that key at our repository. 

#### Ticket Link
Jira: https://mattermost.atlassian.net/browse/DOPS-945


cwarnermm (approved):
Thanks, @pfltdv!

spirosoik (approved):

nab-77 (changes requested):
Bringing this into our infra is big +1 for me. Thank you @pfltdv.

My only question is if we rely on this key for more than just the signed Debian packages should the URL be linked to Debian (i.e. should this not be a more generic link)?

pfltdv:
@nab-77 Yes probably we should generate more generic link to publish it. And it is already mentioned at those pages and used at our repo scripts. 

* https://docs.mattermost.com/install/installing-mattermost-omnibus.html
* https://docs.mattermost.com/guides/deployment.html

@spirosoik @iyampaul Do you know any place that we can put/publish our gpg key?

spirosoik:
@nab-77 could you elaborate a bit more for "generic"? do you mean something like `keys.mattermost.com/*`?

nab-77:
i'm 0/5 on the URL @spirosoik but I am sure that since it's to be used across the packages we should remove the Debian reference from the link or....

we follow the format of https://{platform}.mattermost.com/pubkey.gpg

I guess the latter approach minimises any downstream impacts of changes in the what is live for Mattermost Omnibus, the new Debian Desktop packages (but does create the need to update keys over time in the various areas).

I think either works but a CenOS/RHEL/Rocky/Alma package using deb.packages.mattermost.com link would be interesting ;-)

pfltdv:
So instead of deploying our public key, how about creating a one dedicated keys page and publish our keys in that page in plain text format. So users may copy from it whenever they need. And we may refer to that page internally whenever needed?

cc @nab-77 @spirosoik @iyampaul @phoinixgrr

spirosoik:
> i'm 0/5 on the URL @spirosoik but I am sure that since it's to be used across the packages we should remove the Debian reference from the link or....
> 
> we follow the format of https://{platform}.mattermost.com/pubkey.gpg
> 
> I guess the latter approach minimises any downstream impacts of changes in what is live for Mattermost Omnibus, the new Debian Desktop packages (but does create the need to update keys over time in the various areas).
> 
> I think either works but a CentOS/RHEL/Rocky/Alma package using deb.packages.mattermost.com link would be interesting ;-)
> 
> Also worth noting that the OS vendors like RH, SUSE, all have their various keys published to a security page (plain txt versions to copy).

I am sure that they have URLs for these keys too. The reason is that I want to use an orchestration tool (ansible, simple bash) and I want to download the key from the URL of the provider (Mattermost) instead of copying it because the key can be updated right? 

We can of course put it in a single page but still for automation reasons we need it. cc @pfltdv

pfltdv:
We updated our keys at keybase. Shall we crate another topic to have common location and pipeline to store keys, common scripts etc.

PR #5695, opened 2022-06-03
Message formatting toolbar documentation updates

cwarnermm:
Documentation for: 
- https://github.com/mattermost/mattermost-webapp/pull/10133
- https://github.com/mattermost/mattermost-webapp/pull/10343

- Updated one of the most popular documentation pages on the product documentation site, the Format messages page, to include details on the new message formatting toolbar. 
- Also updated Enable post formatting
user preference documentation to note that the setting is hidden as of v7.0 (and applies the last known value).


source/channels/format-messages.rst
```diff
@@ -18,13 +18,131 @@ Format messages
...
+|                                                            | |code-icon|        | 
++------------------------------------------------------------+--------------------+
+| Add emojis or GIFs                                         | |emoji-icon|       | 
++------------------------------------------------------------+--------------------+
+
+Review how your message formatting will look when the message is sent by selecting the **Show/Hide Preview** |preview-icon| icon. Return to your draft message by selecting the icon again.
+
+You can hide the formatting options by selecting the **Show/Hide Formatting** |hide-formatting-icon| icon. Select the icon again to display the formatting options.
+
+.. tip::
```
cwarnermm, line 132: If users had Enable Post Formatting set to off prior to an upgrade to v7.0, they may never see the Markdown formatting as described in this tip. Consider removing to avoid confusion.

AshishDhama: @cwarnermm They will see formatting options irrespective of their `` Enable Post Formatting`` setting, it doesn’t affect the Advanced text editor at all.
@matthewbirtch Do we need to do something here

cwarnermm: According to the [product documentation](https://docs.mattermost.com/channels/channels-settings.html#enable-post-formatting), the Enable Post Formatting option controls whether post formatting is rendered. When On, posts will be rendered with Markdown formatting. When Off, the raw text will be shown.

Currently, on the Community Server, using the message formatting toolbar options, I see the Markdown syntax because I know I had this setting enabled previously. But if a user had this setting disabled, I'm wondering whether they'd see the syntax. If they don't see it, I don't necessarily want to provide this level of detail in the docs around syntax.

cwarnermm: @esethna has confirmed that the Enable Post Formatting option is being deprecated in favour of the new message formatting toolbar. I'll update this PR based on this information.

source/channels/format-messages.rst
```diff
@@ -18,13 +18,127 @@ Format messages
...
+  :width: 24px
+  :target: https://mattermost.com/deploy
+  :alt: Review your message text formatting using the Show/Hide preview icon in the message formatting toolbar.
+
+Use the messaging formatting toolbar
+------------------------------------
+
+From Mattermost v7.0, you can format your messages in Mattermost using the message formatting toolbar without having to specify any Markdown syntax.
+
+.. image:: ../images/message-formatting-toolbar.png
```
esethna, line 104: Add gif here instead : https://mattermost.atlassian.net/browse/MM-44774

esethna (approved):
Nice work, lets just add the GIF in

AshishDhama (approved):
Thanks @cwarnermm

justinegeffen (approved):

PR #5723, opened 2022-06-15
Update mm.com/download instances

emdecr:
#### Summary
Hi @cwarnermm~
 
We pushed up a recent change to the `/download` and `/deploy` URL setup on mm.com.
- The `/download` page (where we used to house app downloads) now lives at `/apps`
- And `/download` redirects to `/deploy`

Asana ticket with some context [here](https://app.asana.com/0/1201972580778743/1202199255867569/f).
As I made the change, I figured we would need to do some search and replace in docs, and didn't know if we communicated this switch (maybe we did and this PR is moot haha).

Also found a few instances where we were using old anchor links that no longer exist as well.

Drafting this PR for now to get a gut check from folks that I've made the right replacements.  
There's a lot of instances 😅 

Let me know if you have any questions. TY!

cc: @BenLloydPearson for visibility/feedback

### DIFF
--- a/source/configure/configuration-settings.rst
+++ b/source/configure/configuration-settings.rst
@@ -35,7 +35,7 @@ Configuration in database
 
 .. |cloud| image:: ../images/cloud-badge.png
   :scale: 30
-  :target: https://mattermost.com/apps
+  :target: https://mattermost.com/sign-up
   :alt: Available for Mattermost Cloud deployments.
 
 .. |self-hosted| image:: ../images/self-hosted-badge.png
--- a/source/configure/configuration-settings.rst
+++ b/source/configure/configuration-settings.rst
@@ -1737,7 +1737,7 @@ Mattermost Apps Download Page Link
 Configurable link to a download page for Mattermost Apps. When a link is present, an option to **Download Apps** will be added in the Main Menu so users can find the download page. Leave this field blank to hide the option from the Main Menu. Defaults to a page on mattermost.com where users can download the iOS, Android, and Desktop clients. If you're using an Enterprise App Store for your mobile apps, change this link to point to a customized download page where users can find the correct apps.
 
 +------------------------------------------------------------------------------------------------------------------------+
-| This feature's ``config.json`` setting is ``"AppDownloadLink": "https://mattermost.com/apps/"`` with string input. |
+| This feature's ``config.json`` setting is ``"AppDownloadLink": "https://mattermost.com/apps/"`` with string input.     |
 +------------------------------------------------------------------------------------------------------------------------+
 
 Android App Download Link
--- a/source/configure/configuration-settings.rst
+++ b/source/configure/configuration-settings.rst
@@ -5731,7 +5731,7 @@ This setting enables the Apps Bar and moves all Mattermost integration icons fro
 
 .. note::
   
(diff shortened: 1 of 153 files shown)


cwarnermm:
Thanks, @emdecr! This is great! I'll perform a similar search and replace review to catch any stragglers. To summarize the website URL changes:
- App downloads, including mobile and desktop, now live at ``/apps`` from ``/downloads``
- self-hosted deploys live at ``/deploy``
Correct?

A couple of things to note:
- Docs has a backlog task to change all Cloud badges URLs from ``/download`` to ``https://customers.mattermost.com/cloud/signup``. Any concerns about changing the Cloud badge updates in this PR from ``/apps`` to ``/cloud/signup``?
-  You can safely ignore any source files in ``/source/archive``. Source files in the ``archive`` folder are not converted into static HTML during builds, and are not returned in search results. I recommend removing the changes to archived files in this PR to simplify the PR itself by making changes only where needed.

emdecr:
@cwarnermm Yes, those two points are correct~

Re: your notes –

1. Hmm. `https://customers.mattermost.com/cloud/signup` has _just_ been redirected to `https://mattermost.com/sign-up`, so we'll probably want to switch it to that mm.com URL.
2. Ah, cool. Sounds good, I'll revert the changes made in archive files~

I'm about to jump into some tasks, but I'll try to circle back on this within the day. TY!

cwarnermm:
If https://mattermost.com/sign-up/ results in users getting a Cloud instance, then that's the perfect URL to use for the Cloud badges.

cwarnermm (approved):
Thank you so much, @emdecr, for this PR!

source/configure/configuration-settings.rst
```diff
@@ -1737,7 +1737,7 @@ Mattermost Apps Download Page Link
 Configurable link to a download page for Mattermost Apps. When a link is present, an option to **Download Apps** will be added in the Main Menu so users can find the download page. Leave this field blank to hide the option from the Main Menu. Defaults to a page on mattermost.com where users can download the iOS, Android, and Desktop clients. If you're using an Enterprise App Store for your mobile apps, change this link to point to a customized download page where users can find the correct apps.
 
 +------------------------------------------------------------------------------------------------------------------------+
-| This feature's ``config.json`` setting is ``"AppDownloadLink": "https://mattermost.com/download/"`` with string input. |
+| This feature's ``config.json`` setting is ``"AppDownloadLink": "https://mattermost.com/apps/"`` with string input. |
```
cwarnermm, line 1740: Making changes to table text often results in tables becoming malformed (the closing pipe character isn't aligned to the far right). When tables become malformed, Sphinx no longer displays them on the page.

nab-77:
@cwarnermm The change was reversed due primarily related to an in-product link going to mm.com/download for clients (there is also a Google metadata reason that supported reverting back).

cwarnermm:
Thanks, @nab-77! In summary:
- mm.com/apps - works; goes to clients page
- mm.com/download - works; goes to clients page
- in the future: mm.com/download will go to server page

emdecr:
@cwarnermm What you just outlined is how I understand it as well, fwiw.

Aaron just clarified that we're still good for using `/apps` because we want the page to start ranking in the meantime.
I think these changes are still good to go. Cloud Badges go to `/sign-up` and we're now making use of the `/apps` page.

If it looks good to you, I can remove my `Do Not Merge` tag and proceed with the review process? TY~!

PR #5733, opened 2022-06-17
Update search-for-messages.rst

wuwinson:
User reached out about not seeing any results when searching for URLs on postgres database, and I discovered [this forum post](https://forum.mattermost.com/t/solved-search-messages-with-url-keyword-returns-no-results-found/3068) during investigations. Confirmed with Eric that we won't fix this issue unless we do a backend search rewrite from [MM-1082](https://mattermost.atlassian.net/browse/MM-1082). Updating the docs to reflect this issue.


cwarnermm (approved):
Thanks, @wuwinson!

PR #5734, opened 2022-06-17
Sphinx search enhancements

neflyte:
Based on some feedback from @cwarnermm, here are a couple enhancements to the Sphinx search and a quick Makefile fix. :)

---

# Sphinx Search Enhancements

## Goals

- Re-enable search term highlighting on the search results page
- Make the length of search result summary text configurable
- Allow the search result summary to start from the beginning of the summary text

### Secondary Goal

- Fix the `Makefile` to use the correct build and doctree directories for the `livehtml` target

### Added

- Configurable options in `searchtools.js` to control:
	- Length of search result summary text
	- Using the beginning of the summary text as the basis for the search result summary
- Functionality to construct the search result summary starting from the beginning of the text rather than an excerpt

### Changed

- Length of search result summary text was increased from 240 characters (the Sphinx default) to 300 characters
- Search result summary text is taken from the start of the text versus an excerpt
- The `livehtml` target in the `Makefile` passes the correct build directory (`build/html`) and doctree directory (`build/doctrees`)

### `searchtools.js` Options

The following options are available to be configured in the `source/_static/searchtools.js` file:

```javascript
    /**
     * The length of text to use for the summary of each search result. If the summary text
     * is longer than this length, an ellipsis (...) will be appended.
     * The Sphinx default value is 240.
     */
    _summaryTextLength : 300,
    /**
     * Use the default Sphinx search summary generator instead of using the first x characters
     * of the summary text.
     */
    _useDefaultSearchSummary : false,
```

After making changes to any of the above options, re-run the `make html` command to make the changes take effect.


cwarnermm:
Hi @andrewbrown00! This docs PR re-introduces search result highlights within the search results view, and it changes how the results view displays page excerpts. 

In production today, you'll see that results include an excerpt that often looks like it starts mid-sentence. What Sphinx is actually doing is focusing the excerpt content around the first instance of a matching search term.

This PR displays excerpts containing up to 300 characters, starting from the top of the page. The result is a more readable excerpt with clearer context. However, by basing the excerpt from the the top of the page rather than the first instance of a search term, the user will likely see fewer highlighted terms in a search results view.

Looking forward to your feedback on this PR.

wiersgallak:
Thank you for helping us with this @neflyte! 

@cwarnermm Just to confirm behavior I wanted to share my results: 

When searching "log", I receive results with any docs pages which includes the search term. 

I will see the first 300 characters in the search excerpt and if the term I searched ("log") is present in the excerpt, I will see it highlighted. 
image

Once I click on a page that I think best answers my question, I do not see the search term highlighted on the page contents. 

Just as a point of reference, after a quick review of Gitlab's handbook and Slack's docs, this seems to be inline with what they are offering, however with some small behavior adjustments. I look forward to Andrew's input from the UX perspective.

cwarnermm:
You have described the behavior well, @wiersgallak. That's precisely how it's intended to work. On any given docs page, if you wanted to see the search highlighting, you could append ``?highlight=searchterm`` to the end of the page URL in the browser.

andrewbrown00 (approved):
Thanks all, changes look good and work as expected 🙂

Having the search term highlighted on the initial results page allows users to scan for their searched term quickly. Once the user selects a topic to drill down on, we remove the highlighting to reduce the page's noise since the content is no longer about the searched term, but the content itself.

@cwarnermm @wiersgallak

**Screenshots for reference**

| Results | Content |
|-----|-----|
| mattermost-docs-preview-pulls s3-website-us-east-1 amazonaws com_5734_search html_q=design check_keywords=yes area=default | mattermost-docs-preview-pulls s3-website-us-east-1 amazonaws com_5734_about_faq-design-decisions html |

justinegeffen (approved):

PR #5748, opened 2022-06-27
Fix the path for Desktop App logs for Linux users

devinbinnie:
#### Summary
There's a new directory that logs can be stored. I'm just adding it to the docs :)


cwarnermm (approved):
Thanks, @devinbinnie!

PR #5793, opened 2022-07-21
BUG: Env variables don't overwrite config on save

ashishbhate:
#### Summary

- The value of the environment variable is NOT written back to the config.json as that setting's value.

#### Ticket Link

- NA


noxer (approved):
Perfect, thank you!

cwarnermm (approved):
Thank you, @ashishbhate!

PR #5799, opened 2022-07-25
Export file names include Job ID

cwarnermm:
Documentation for: https://github.com/mattermost/mattermost-server/pull/20675
- Updated `mmctl export list` and `mmctl export job list` documentation to note that export files contain the Job ID in the file name.


noxer:
Technically it's not the Job ID *as* a file name, it just contains it. It looks like this: `jobID12345_export.zip`.

noxer (approved):
Perfect, thank you very much!

PR #5807, opened 2022-07-26
Added mmctl prefix to command examples

cwarnermm:
``mmctl`` prefixes were missing from most command examples in the product documentation which was resulting in confusion.


justinegeffen (approved):
Nice work!

PR #5813, opened 2022-07-28
Link to software requirements in Prepare To Upgrade

sadohert:
Adding a link and suggestion to double check OS and database versions.  Maybe this should be even broader (hardware specs, etc....)... that'll be for another PR.

I'm not married to the wording at all.  I considered putting it further down under "Upgrading from older versions", but I think if someone's coming here, they should be told to double check these things no matter what.


justinegeffen (approved):

cwarnermm (approved):
Thanks, @sadohert!

PR #5802, opened 2022-07-25
Docs listening: Update location of macOS logs

cwarnermm:
From Monterey/macOS 12, Mattermost logs have moved to ~Library/Containters/Mattermost.Desktop/Data/Library/Logs/Mattermost

PR #5822, opened 2022-08-05
Update troubleshooting.rst

anx-ag:
OSX Monterey Logpaths are different

### DIFF
--- a/source/install/troubleshooting.rst
+++ b/source/install/troubleshooting.rst
@@ -114,7 +114,7 @@ The desktop app log file can be found in the user directory:
 
 - **Windows:** ``%userprofile%\AppData\Roaming\Mattermost\logs``
 - **Linux:** ``~/.local/share/Mattermost/logs`` OR ``~/.config/Mattermost/logs``
-- **MacOS:** ``~/Library/Logs/Mattermost``
+- **MacOS:** ``~/Library/Logs/Mattermost`` OR ``~Library/Containters/Mattermost.Desktop/Data/Library/Logs/Mattermost`` (on Monterey and upwards)
 
 Mattermost Browser App logs
 ---------------------------
--- a/source/install/troubleshooting.rst
+++ b/source/install/troubleshooting.rst
@@ -114,7 +114,7 @@ The desktop app log file can be found in the user directory:
 
 - **Windows:** ``%userprofile%\AppData\Roaming\Mattermost\logs``
 - **Linux:** ``~/.local/share/Mattermost/logs`` OR ``~/.config/Mattermost/logs``
-- **MacOS:** ``~/Library/Logs/Mattermost`` OR ``~Library/Containters/Mattermost.Desktop/Data/Library/Logs/Mattermost`` (on Monterey and upwards)
+- **MacOS:** ``~/Library/Logs/Mattermost`` (DMG installation) OR ``~Library/Containers/Mattermost.Desktop/Data/Library/Logs/Mattermost`` (Appstore installation)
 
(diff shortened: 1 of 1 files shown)


coltoneshaw (approved):
Good catch! This is most likely due to the change to an App Store Application, which is containerized. If you download the `.dmg` it should be back in the old directory.

anx-ag:
Hey, you're right, I just tried it out on my Monterey VM and if I use the `.dmg`, the application uses the logs as outlined in the documentation, so the comment I proposed is wrong, since the log location is not dependent on the OSX version, but on the installation method. I think we should change it to f.ex. 

 - **MacOS:** ``~/Library/Logs/Mattermost`` (DMG install) OR ``~Library/Containters/Mattermost.Desktop/Data/Library/Logs/Mattermost`` (Appstore Install)

Not sure if I should open a new request here for that, but I think my PR should be modified again.

justinegeffen:
> Hey, you're right, I just tried it out on my Monterey VM and if I use the `.dmg`, the application uses the logs as outlined in the documentation, so the comment I proposed is wrong, since the log location is not dependent on the OSX version, but on the installation method. I think we should change it to f.ex.
> 
>     * **MacOS:** `~/Library/Logs/Mattermost` (DMG install) OR `~Library/Containters/Mattermost.Desktop/Data/Library/Logs/Mattermost` (Appstore Install)
> 
> 
> Not sure if I should open a new request here for that, but I think my PR should be modified again.

You can definitely update this PR. Just wanted to confirm that the directory is called "Containters" and not "Containers". :)

anx-ag:
Sorry again - typo, bad english and wrong information. A perfect first PR (updated now, btw).

cwarnermm:
Great minds think alike, @anx-ag! :) I recently staged a similar change for our upcoming v7.2 release: https://github.com/mattermost/docs/pull/5802/files

In my PR, I'm calling out a distinction based on the macOS version the user has. In your PR, you're calling out a distinction based on how the client was deployed. I feel that users will benefit from a combination of both your PR and my PR. What do you think?

anx-ag:
This was also my first assumption, but what @coltoneshaw said is correct: It does not depend on the OSX version, but only on the deployment model. I spun up a Monterey virtual machine and used the DMG installer from Github and was immediately put back to the original Logfile location. As soon as you install the app out of the appstore, the log file location changes to the Containers subdirectory.

coltoneshaw:
Changes look good to me!

cwarnermm:
Thank you @anx-ag and @coltoneshaw! @nab-77 - We've got a case where log paths change depending on the distribution method. Is this by design and intended?

coltoneshaw:
It's by design and how electron and Apple handle log paths. The desktop team recently submitted our desktop app to the macOS store for v5.1 and above. When this happens Apple by default containerizes the application, which also means a new log path is used. 

Electron logs in the log path available to the application, which will be different based on the installation method. 

So, based on how the user and their organization choose to distribute the Mattermost app Will dictate where the logs are stored. Unfortunately, when it comes to Apple this is just how it works and cannot be easily worked around.

cwarnermm:
Thanks, @coltoneshaw! This is great context!

cwarnermm (approved):
Thank you, @anx-ag!

anx-ag:
Ah, I didn't check it in real-life obviously - the text is so long now that the line gets wrapped :-/
Should we revert to (DMG install) and (Appstore Install) to save some characters (in the hope, that it will not wrap the text then)?

PR #5832, opened 2022-08-12
Small fixes identified while following

anx-ag:
I tried to follow the instructions today and recognized a few issues with this documentation.

If you do not move the "bulk-export-attachments" folder to a "data" subfolder, the import process is not able to find the attachments.
Also the filename "mattermost-bulk-export.zip" was wrong (should be import).

I also saw that this document uses content backlinks top, others don't. What's the default setting of that option? I mean the documentation looks a bit different anways depending on what site you are (anchor links on the left, anchor links on the right, clickable headings yes/no)...

Changing the "backlinks" option to "none" here fixes this problem and it's much more intuitive IMHO if you want to share the link to a section, but then again, you cannot go back to the TOC with just one click.

### DIFF
--- a/source/onboard/migrating-to-mattermost.rst
+++ b/source/onboard/migrating-to-mattermost.rst
@@ -184,17 +184,19 @@ Next, run this command to do the conversion. Replace ``<TEAM NAME>`` with the na
 
     ./mmetl transform slack --team <TEAM NAME> --file export-with-emails-and-attachments.zip --output mattermost_import.jsonl
 
-Next you have to create a zip file with the ``mattermost_import.jsonl`` file and the directory ``bulk-export-attachments`` that contains the attachments. On Linux and Mac you can use this command:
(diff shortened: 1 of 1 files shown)


source/onboard/migrating-to-mattermost.rst
```diff
@@ -184,17 +184,19 @@ Next, run this command to do the conversion. Replace ``<TEAM NAME>`` with the na
 
     ./mmetl transform slack --team <TEAM NAME> --file export-with-emails-and-attachments.zip --output mattermost_import.jsonl
 
-Next you have to create a zip file with the ``mattermost_import.jsonl`` file and the directory ``bulk-export-attachments`` that contains the attachments. On Linux and Mac you can use this command:
+Next you have to create a zip file with the ``mattermost_import.jsonl`` file and the directory ``bulk-export-attachments`` (which needs to be moved to a subdirectory ``data`` first) that contains the attachments. On Linux and Mac you can use this command:
```
cwarnermm, line 187:
```suggestion
Next you have to create a zip file with the ``mattermost_import.jsonl`` file and the directory ``bulk-export-attachments`` (which needs to be moved to a subdirectory ``data`` first) that contains the attachments. On Linux and macOS you can use this command:
```

cwarnermm (approved):
Thank you, @anx-ag!

PR #5852, opened 2022-08-22
Post formatting options can be disabled

cwarnermm:
Documentation for: https://github.com/mattermost/mattermost-webapp/pull/10881
- Clarified message formatting documentation with hide vs disable details
- corrected System Console paths for Environment > Developer config setting options

### DIFF
--- a/source/channels/format-messages.rst
+++ b/source/channels/format-messages.rst
@@ -21,37 +21,31 @@ Format messages
 .. |bold-icon| image:: ../images/format-bold_F0264.svg
   :height: 24px
   :width: 24px
-  :target: https://mattermost.com/deploy
   :alt: Bold message text using the Bold icon in the message formatting toolbar.
 
 .. |italics-icon| image:: ../images/format-italic_F0277.svg
   :height: 24px
   :width: 24px
-  :target: https://mattermost.com/deploy
   :alt: Italicize message text using the Italic icon in the message formatting toolbar.
 
 .. |strikeout-icon| image:: ../images/format-strikethrough-variant_F0281.svg
   :height: 24px
   :width: 24px
-  :target: https://mattermost.com/deploy
   :alt: Strike out message text using the Strikethrough icon in the message formatting toolbar.
 
 .. |headings-icon| image:: ../images/format-header_E81D.svg
   :height: 24px
   :width: 24px
-  :target: https://mattermost.com/deploy
   :alt: Format message text as a heading using the Heading icon in the message formatting toolbar. Headings 1 through 6 are supported.
 
 .. |links-icon| image:: ../images/link-variant_F0339.svg
   :height: 24px
   :width: 24px
-  :target: https://mattermost.com/deploy
   :alt: Add a message link using the Link icon in the message formatting toolbar.
 
 .. |attachments-icon| image:: ../images/paperclip_F03E2.svg
   :height: 24px
   :width: 24px
-  :target: https://mattermost.com/deploy
   :alt: Add a message attachment using the Upload files icon in the message formatting toolbar.
 
 .. |numbered-icon| image:: ../images/format-list-numbered_F027B.svg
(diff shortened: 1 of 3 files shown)


esethna (approved):

AshishDhama (approved):

source/channels/format-messages.rst
```diff
@@ -128,7 +122,11 @@ The message formatting toolbar offers the following formatting options:
 
 Review how your message formatting will look when the message is sent by selecting the **Show/Hide Preview** |preview-icon| icon. Return to your draft message by selecting the icon again.
 
-You can hide the formatting options by selecting the **Show/Hide Formatting** |hide-formatting-icon| icon. Select the icon again to show the formatting options.
+.. tip::
+
+  - Hide the formatting options by selecting the **Show/Hide Formatting** |hide-formatting-icon| icon. Select the icon again to show the formatting options. 
+  - Disable the formatting options by selecting the **Gear** |settings-icon| icon, going to **Settings > Advanced > Enable Post Formatting**, and setting this option to **Off**.
```
hmhealey, line 128: 0/5, but I don't think we need to document this here since it's not something we expect people to leave on for long periods of times

source/channels/format-messages.rst
```diff
@@ -128,7 +122,11 @@ The message formatting toolbar offers the following formatting options:
 
 Review how your message formatting will look when the message is sent by selecting the **Show/Hide Preview** |preview-icon| icon. Return to your draft message by selecting the icon again.
 
-You can hide the formatting options by selecting the **Show/Hide Formatting** |hide-formatting-icon| icon. Select the icon again to show the formatting options.
+.. tip::
+
+  - Hide the formatting options by selecting the **Show/Hide Formatting** |hide-formatting-icon| icon. Select the icon again to show the formatting options. 
+  - Disable the formatting options by selecting the **Gear** |settings-icon| icon, going to **Settings > Advanced > Enable Post Formatting**, and setting this option to **Off**. 
+  - System Admins can disable the formatting options for all users in the System Console by going to **Environment > Developer > Enable Client Performance Debugging**. See the `configuration settings <https://docs.mattermost.com/configure/environment-configuration-settings.html#enable-client-performance-debugging>`__ documentation for details.
```
hmhealey, line 129: This hasn't been changed yet, so it's not accurate. That setting is currently available all of the time

hmhealey (changes requested):

source/channels/format-messages.rst
cwarnermm: Docs has received customer feedback asking for details on how individual users can disable the new formatting options toolbar in their Mattermost clients. Is that what the Settings > Advanced > Enable Post Formatting offers? Or is to doing something else?

source/channels/format-messages.rst
cwarnermm: @hmhealey - My understanding of the configurability of this feature may be incorrect. I thought we had:
- a way to show/hide the formatting options within the UI element itself (the eye icon)
- a way for users to disable it via Settings so they never see it again
- a server-side kill switch in the System Console whose current label isn't as discoverable as we'd like

Which parts of my understanding are incorrect?

source/channels/format-messages.rst
hmhealey: No, it's for turning off Markdown rendering in posts. It was added originally to debug some very difficult to work around issues we used to have when Markdown was newer

cwarnermm: Thanks, @hmhealey! I'll update this docs PR accordingly.

source/channels/format-messages.rst
hmhealey: As above, I think we're getting some stuff confused here. For the post textbox:
- The eye button in the post textbox for turning on and off the formatting toolbar on it.
- I think there's a server-side setting for turning the formatting toolbar on and off entirely. Presumably this is going away eventually since we want everyone using it.

That's separate from the Enable Post Formatting option in the Advanced settings. We did also discuss moving it behind the Enable Client Performance Debugging setting since it's only intended for debugging, but that hasn't been done yet

cwarnermm: Thank you for the clarifications! The misunderstandings are entirely mine :)

hmhealey (approved):

laneycs (approved):

PR #5867, opened 2022-08-31
Implement mid-page link redirects

neflyte:
#### Summary
This PR introduces a solution to redirecting users from a subheading (a.k.a. HTML anchor, URL fragment) on one page to a different page or subheading. This kind of redirect is referred to as a  “_mid-page redirect_”.

For example, let’s say we moved a subheading named `how-do-push-notifications-work` from the `/mobile/mobile-faq.html` page to the `/deploy/mobile-faq.html` page.

Before this PR, the redirect at `/mobile/mobile-faq.html#how-do-push-notifications-work` would send the browser to the `/deploy/mobile-faq.html` page with whatever the last subheading redirect for that page was; it could be any subheading in the source page.

With this PR, the redirect at `/mobile/mobile-faq.html#how-do-push-notifications-work` will send the browser to the desired page and subheading.

The additional logic which enables mid-page redirects is implemented in JavaScript on the browser side. The logic is added to:

  - the redirect page - used in cases when the source page no longer exists
  - existing pages - used in cases when only a subheading has moved

The Sphinx `reredirects` extension is updated to:

  - Use Sphinx’s internal facility for writing redirect pages instead of writing pages independently
  - Use template files for redirects where a whole page has moved
  - Inject JavaScript logic into existing pages where subheadings have moved
  - No longer include unused implementation from the original `sphinx_reredirects` extension
  - Allow for redirect pages to be written without a file extension (e.g. `.html`) for environments where the HTTP server does not try appending a file extension on URIs that do not resolve to a resource; referred to as “_extensionless pages_”
  - Allow for redirects to function in environments where the base URL to the documentation set contains a path prefix (e.g. Mattermost preview environments)

A test suite has been added for the redirects extension. It can be run using the `test` Makefile target (i.e. `make test`).

**NOTE**: Please run `pipenv sync --dev` before running the test suite to ensure `pytest` has been installed.

---

New config (`conf.py`) options:
  - `mm_url_path_prefix` - defines a path prefix to prepend to redirect targets (default: "")
  - `redirect_write_extensionless_pages` - enables writing extensionless redirect pages (default: `False`)

New template files:
  - `simpleredirect.html` - used for a “simple” redirect of one whole page to another
  - `redirect.html` - used for redirecting multiple subheadings of one page to various different pages and subheadings

JavaScript logic:

The mid-page redirect logic depends on a mapping of URL fragment to redirect target that is injected into HTML pages as they are written. The mapping is generated by the `build_js_object()` function and looks like the following:

```javascript
const redirects = Object.freeze({“source_fragment”: “target_URL”, ...});
```

For redirects where a page with subheadings has moved, the logic looks like the following:
```html
<script>
  const DEFAULT_PAGE = "-";
  {{ redirects_object }}
  const url = new URL(window.location.href);
  let redirectKey = DEFAULT_PAGE;
  if (url.hash !== "" && url.hash !== "#") {
      redirectKey = url.hash.substring(1);
  }
  if (redirectKey in redirects) {
      let redirecturl = redirects[redirectKey];
      if (redirecturl.startsWith("/")) {
          redirecturl = window.location.origin + redirecturl;
      }
      window.location = redirecturl;
  } else {
      console.error("unknown redirect: " + redirectKey);
  }
</script>
```
The text `{{ redirects_object }}` is replaced with the generated redirect mapping.

For existing pages that have moved subheadings to other pages or subheadings, the logic looks like the following:
```html
<script>
  {{ fragment_redirects }}
  const url = new URL(window.location.href);
  if (url.hash !== "" && url.hash !== "#") {
    const fragment = url.hash.substring(1);
    if (fragment in redirects) {
        let redirecturl = redirects[fragment];
        if (redirecturl.startsWith("/")) {
            redirecturl = window.location.origin + redirecturl;
        }
        console.log("redirecting to " + redirecturl);
        window.location = redirecturl;
    } else {
        console.error("unknown fragment: " + fragment);
    }
  }
</script>
```
The text `{{ fragment_redirects }}` is replaced with the generated redirect mapping.


pfltdv:
Preview environment is ready http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/5868

neflyte:
I've made a change to allow redirects to work in a preview environment. The change requires an update to the command that builds the docs so we can specify the PR ID and add that to redirect target paths.

For example, in `preview-env-fork.yml` I modified the `Build` step to use the following command:
`make SPHINXOPTS="-j auto -D mm_url_path_prefix=/${{ github.event.number }}" html`

This makes the build prepend the PR ID path to redirect target URLs. A redirect target would look like `/1234/messaging/organizing-conversations.html` instead of `/messaging/organizing-conversations.html`.

@pfltdv Would this change be okay? I made a similar change to `preview-env.yml` for non-forked repos.

pfltdv:
> I've made a change to allow redirects to work in a preview environment. The change requires an update to the command that builds the docs so we can specify the PR ID and add that to redirect target paths.
> 
> For example, in `preview-env-fork.yml` I modified the `Build` step to use the following command: `make SPHINXOPTS="-j auto -D mm_url_path_prefix=/${{ github.event.number }}" html`
> 
> This makes the build prepend the PR ID path to redirect target URLs. A redirect target would look like `/1234/messaging/organizing-conversations.html` instead of `/messaging/organizing-conversations.html`.
> 
> @pfltdv Would this change be okay? I made a similar change to `preview-env.yml` for non-forked repos.

Thanks alot @neflyte for making the redirection work in the PR environment. Changes are definitely ok. Great work.

neflyte:
@pfltdv Is there anything else I need to do to get the pipeline changes I mentioned to work? The preview doesn't seem to be updated with the expected redirect pages.

pfltdv:
> @pfltdv Is there anything else I need to do to get the pipeline changes I mentioned to work? The preview doesn't seem to be updated with the expected redirect pages.

@neflyte For PR's from forked repositories we are using workflow run file from the master branch, since GitHub does not expose the secrets to the forked PRs. I will create a preview environment from cloned branch.

pfltdv:
@neflyte PR environment is created with same commits at http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/5874

pfltdv:
@neflyte @cwarnermm Pipeline changes are merged into master. Is it possible to merge master into this branch so changes will be active.

cwarnermm:
@pfltdv - Yes. @neflyte needs to resolve the file conflict in this PR, and then we can merge master into this branch.

cwarnermm:
@iyampaul - Adding you as a reviewer to share the fix we're applying to our mid-page redirects based on our recent troubleshooting conversation and your great idea for moving forward.

cwarnermm (approved):
Thanks, @neflyte! Really appreciate that you've created tests to validate redirect logic behaviour on demand. This provides us with confidence that redirects are working and provides better troubleshooting paths when needed.

source/templates/layout.html
```diff
@@ -237,6 +237,27 @@
...
       {%- endif %}
+
+      {# Fragment redirects #}
+      {%- if has_fragment_redirects %}
+      <script>
+        {{ fragment_redirects }}
+        const url = new URL(window.location.href);
+        if (url.hash !== "" && url.hash !== "#") {
+          const fragment = url.hash.substring(1);
+          if (fragment in redirects) {
```
iyampaul, line 248: Just to confirm: `redirects` is the path redirects we've defined in the config, right?

iyampaul: I think it'd be valuable to add some input validation here since we're taking that fragment/substring directly from the end-user.

source/templates/redirect.html
```diff
@@ -0,0 +1,30 @@
...
+<body>
+<script>
+    const DEFAULT_PAGE = "-";
+    // The line below is injected; all other lines are constant
+    {{ redirects_object }}
+    // The line above is injected; all other lines are constant
+    const url = new URL(window.location.href);
+    let redirectKey = DEFAULT_PAGE;
+    if (url.hash !== "" && url.hash !== "#") {
+        redirectKey = url.hash.substring(1);
```
iyampaul, line 16: We'll want to add the input validation here too.

source/templates/layout.html
neflyte: I've added fragment validation into both `layout.html` and `redirect.html`

source/templates/redirect.html
neflyte: I've added fragment validation to this code.

iyampaul (approved):

PR #5880, opened 2022-09-07
Update self-managed changelog for Insights

amyblais:
Add details for how to disable Insights https://community.mattermost.com/private-core/pl/ke3fjt4ifbfc7nbh4c8rrdn9wy.

Feel free to suggest additional details to add here. E.g. what can we describe about the DB performance issues?

### DIFF
--- a/source/install/self-managed-changelog.md
+++ b/source/install/self-managed-changelog.md
@@ -160,7 +160,7 @@ Mattermost v7.1.0 contains low severity level security fixes. [Upgrading](https:
 ### Highlights
 
 #### Insights (Beta) (Enterprise and Professional)
- - Added workplace insights consisting of usage and behavior data, which helps Enterprises further increase productivity of their employees through Mattermost functionality. This feature can be disabled using the ``MM_FeatureFlag_InsightsEnabled`` feature flag.
+ - Added workplace insights consisting of usage and behavior data, which helps Enterprises further increase productivity of their employees through Mattermost functionality. This feature can be disabled by setting the ``MM_FEATUREFLAGS_INSIGHTSENABLED`` environment variable to ``false``.
 
 ### Improvements
 
--- a/source/install/self-managed-changelog.md
+++ b/source/install/self-managed-changelog.md
@@ -160,7 +160,7 @@ Mattermost v7.1.0 contains low severity level security fixes. [Upgrading](https:
 ### Highlights
 
 #### Insights (Beta) (Enterprise and Professional)
- - Added workplace insights consisting of usage and behavior data, which helps Enterprises further increase productivity of their employees through Mattermost functionality.
+ - Added workplace insights consisting of usage and behavior data, which helps Enterprises further increase productivity of their employees through Mattermost functionality. This feature can be disabled using the ``MM_FeatureFlag_InsightsEnabled`` feature flag.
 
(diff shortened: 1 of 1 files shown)


source/install/self-managed-changelog.md
```diff
@@ -160,7 +160,7 @@ Mattermost v7.1.0 contains low severity level security fixes. [Upgrading](https:
 ### Highlights
 
 #### Insights (Beta) (Enterprise and Professional)
- - Added workplace insights consisting of usage and behavior data, which helps Enterprises further increase productivity of their employees through Mattermost functionality.
+ - Added workplace insights consisting of usage and behavior data, which helps Enterprises further increase productivity of their employees through Mattermost functionality. This feature can be disabled using the ``MM_FeatureFlag_InsightsEnabled`` feature flag.
```
agnivade, line 163: I am not sure this clarifies how _exactly_ can it be disabled via the feature flag. The environment variable needs to be set to false. 

@cwarnermm - do we want to document that here? Or is there somewhere specific in general where feature flags and how to disable them are documented?

amyblais: There are some details here https://developers.mattermost.com/contribute/server/feature-flags/#faq, but probably doesn't have all the details.

agnivade: Yes, that is there, I am just not sure if customers are aware of that since that's a developer oriented page. I'll be fine if we link to that page. Also, the feature flag should be `MM_FEATUREFLAGS_INSIGHTSENABLED`. Note the "s" at end of "flag".

cwarnermm: I don't believe there is any content available on docs.mm.com today that describes how to manage feature flags. The feature flag content on dev.mm.com that Amy linked to is likely the most comprehensive information we have published across our information assets.

The docs don't have feature flag content because feature flags are in place to roll out new functionality, and are subsequently removed once functionality is verified via a Cloud release. Instead of documenting feature flags, the product documentation includes all configuration settings that disable product functionality.

I'm 0/5 on where in the product documentation we should add details on disabling Insights, but here are a couple of  options if the changelogs don't feel like the right place:
- On the Insights page itself: https://docs.mattermost.com/welcome/insights.html
- In our configuration settings documentation for system admins: https://docs.mattermost.com/configure/configuration-settings.html

I'm new to disabling functionality via feature flags. What does this process look like in terms of actionable steps?
- Does a self-hosted system admin update the ``config.json`` file with an environment variable?
- What would Cloud system admins need to do to disable Insights?
- Do admins need to perform extra steps for Mattermost clients including desktop and mobile (v1/v2)?
- Is there a System Console UI admins can use to disable feature flags? Or is that read-only?
- Once Insights is disabled, can admins re-enable at a later date?

agnivade: Understood. @amyblais - let's just avoid the word feature flag and say "This feature can be disabled by setting the `MM_FEATUREFLAGS_INSIGHTSENABLED` environment variable to `false`." It clearly says what needs to be done, and avoids any further rabbit holes.

amyblais: I've applied the suggested edit 👍

mkraft (approved):

cwarnermm (approved):

PR #5883, opened 2022-09-08
Fix the section of user authentication for workspace migration

spirosoik:
#### Summary
When the authentication method is email then we are not migrating the passwords and  we require all users to create new passwords at first login. This prevents the need of moving the credentials in bulk from one workspace to another and the risk that involves.

### DIFF
--- a/source/manage/cloud-data-export.rst
+++ b/source/manage/cloud-data-export.rst
@@ -140,6 +140,8 @@ Then extract the export file to use it by running the following mmctl command:
 
 Once your migration is complete and you’ve imported your data into your self-hosted instance we recommend that you take a few days to validate your data and ensure everything is working as expected before taking down your Cloud instance.
 
+.. note:: If you are using email/password authentication, your users require to reset their passwords.
+
 If you encounter any issues or problems, please contact our Support team via https://customers.mattermost.com/cloud/contact-us, or through the `Mattermost Help Center <https://support.mattermost.com/>`_.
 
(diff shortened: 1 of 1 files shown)


jasonblais (approved):

source/manage/cloud-data-export.rst
```diff
@@ -153,7 +155,7 @@ Before you begin your migration, ensure you’re on Mattermost version 6.5 (or l
 
 **User Authentication**
 
-If you’re using a login method other than email, you’ll need to make the required changes to your infrastructure so that you can continue your existing authentication method in the cloud. Examples of these methods are LDAP and GitLab SSO.
+If you’re using an email login method, your users need to reset their passwords after the migration has been completed. For any other authentocation method, you’ll need to make the required changes to your infrastructure so that you can continue your existing authentication method in the cloud. Examples of these methods are LDAP and GitLab SSO.
```
nab-77, line 158:
```suggestion
If you are using the email login method, users will need to reset their passwords after the migration has been completed. Other authentication methods, such as LDAP and GitLab SSO, required changes to your infrastructure allowing the specific authentication method to function in Mattermost cloud.
```

nab-77 (approved):
Small edits but on the whole a much needed update. Thanks @spirosoik.

cwarnermm (approved):
Thanks, @spirosoik! I've provided editorial feedback inline. Approving for expediency.

PR #5892, opened 2022-09-15

anx-ag:
5.37.10 has been released.


JtheBAB (commented):
Would it not be better to just say 

"upgrade to the latest 5.37 version" 

?

anx-ag:
Not sure how many 5.37 releases will be shipped, if 5.37.10 is the last one, it should be OK I guess.

amyblais (approved):

PR #5919, opened 2022-10-06
Update docs regarding Illegal mix of collations

mukul-kr:
#### Summary

Update Documentation for the SQL collation error as suggested.
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/20636

### DIFF
--- a/source/upgrade/prepare-to-upgrade-mattermost.rst
+++ b/source/upgrade/prepare-to-upgrade-mattermost.rst
@@ -20,6 +20,8 @@ Mattermost v7.1 introduces schema changes in the form of a new column and its in
 
 You can run the following SQL queries before the upgrade that obtains a lock on ``Reactions`` table. Users' reactions posted during this time won't be reflected in the database until the migrations are complete. This is fully backwards-compatible.
 
+If your connection collation and table collations are different can give error `Illegal mix of collations`. The solution is to set the same collation for both connection and the table. There are different collations at different levels - connection, database, table, column. And due to various reasons, DB administrators can choose to set different collation levels for different objects.
+
 .. tabs:: 
 
   .. tab:: MySQL


agnivade (approved):
Nothing apart from Carrie's suggestions. Thanks!

mukul-kr:
Hey, @cwarnermm should I rebase or merge with a commit?

cwarnermm:
Thank you for applying the suggested changes, @mukul-kr! I will rebase and merge your PR into our code base.

cwarnermm (approved):

cwarnermm:
Thank YOU, @mukul-kr! It's contributions like yours that make Mattermost product documentation become best in class and delivers great value.

adsche:
This change somewhat disconnects the paragraph about the "Reactions" table from the code to update that same table. It seems this has nothing to do with Reactions, as this only really breaks due to focalboard or playbook upgrades from my experience?

Additionally, should this maybe come with code on how to check that the the collations are correctly setup for the upgrade - since this is the "prepare for upgrade" page? I still don't know what was wrong in the database, my collations are all set to the same value, yet most updates of mattermost result in broken collations. I checked with `SHOW VARIABLES LIKE 'collation%';` in MySQL to confirm connection and server values.

Additionally, it would be great to provide (on the Troubleshooting page?) code to fix the broken tables after the update. Here's what I did in MySQL:

`SELECT CONCAT("ALTER TABLE ", TABLE_SCHEMA, '.', TABLE_NAME," COLLATE utf8mb4_0900_ai_ci;") AS Command FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA="mattermost" AND TABLE_TYPE="BASE TABLE";`

This generates commands that I had to copy to another file (e.g., `collation.sql`) and then execute via `mysql`. I am sure there are better ways to do it, maybe you could document one. (NB This takes forever on a <50 user/ <6mio posts instance, especially on the 'Posts' table.)

PR #5939, opened 2022-10-13

amyblais:
Minor update to reflect that React 17 changes are coming in v7.6 instead of v7.5.


source/upgrade/important-upgrade-notes.rst
```diff
@@ -9,7 +9,7 @@ Important Upgrade Notes
...
+   - Plugins with a webapp component may need to be updated to work with upcoming Mattermost v7.6 and the updated ``React v17``. This is to avoid plugins crashing with an error about ``findDOMNode`` being called on an unmounted component. While our `starter template <https://github.com/mattermost/mattermost-plugin-starter-template>`_ depended on an external version of ``React``, it did not do the same for ``ReactDOM``. Plugins need to update their ``webpack.config.js`` directives to externalize ``ReactDOM``. For reference, see https://github.com/mattermost/mattermost-plugin-playbooks/pull/1489. Server-side only plugins are unaffected. This change can be done for existing plugins any time prior to upgrading to Mattermost v7.6 and is backwards compatible with older versions of Mattermost.
```
cwarnermm, line 12:
```suggestion
   - Plugins with a webapp component may need to be updated to work with the upcoming Mattermost v7.6 release and the updated ``React v17`` dependency. This is to avoid plugins crashing with an error about ``findDOMNode`` being called on an unmounted component. While our `starter template <https://github.com/mattermost/mattermost-plugin-starter-template>`_ depended on an external version of ``React``, it did not do the same for ``ReactDOM``. Plugins need to update their ``webpack.config.js`` directives to externalize ``ReactDOM``. For reference, see https://github.com/mattermost/mattermost-plugin-playbooks/pull/1489. Server-side only plugins are unaffected. This change can be done for existing plugins any time prior to upgrading to Mattermost v7.6 and is backwards compatible with older versions of Mattermost.
```

cwarnermm (commented):
Provided feedback inline.

cwarnermm (approved):

hanzei:
cc @mickmister @catalintomai

Issue #5955, opened 2022-10-18
Help Wanted: Clarify whether inactive/deactivated users count towards the total number of active users reported via System Console


cwarnermm:
**Problem space**
A common question system admins ask on the Community Server is whether inactive and/or deactivated users count towards the active user count reported in the System Console. This common question indicates that the product documentation is missing this information.

Answer: No. The Active User Count field reported via **System Console > Site Stats** is calculated using logic documented here: https://github.com/coltoneshaw/Mattermost-SQL-Queries#active-users'

**Target audience**: System Administrators

**Goal**
Clarify in product documentation that inactive and deactivated users aren't counted towards the total number of active users.

**Delivery expectations**
- 1 PR with all proposed changes
- Update the Reporting configuration settings admin documentation page via https://docs.mattermost.com/configure/reporting-configuration-settings.html#site-statistics
- Comment in the PR to request a generated preview of your work. The Docs team will add the appropriate labels to your PR.

**Tech needed**
- RST format
- Sphinx

**Affected repository**: mattermost/docs

**Where to get help**
Join the Mattermost Docs team on the Mattermost Community Server in the [~Documentation Working Group channel](https://community.mattermost.com/core/channels/dwg-documentation-working-group) to discuss this ticket, or feel free to comment in this docs ticket.

PR #5965, opened 2022-10-19
Clarifying active user count behaviour (#5955)

hmmmmmmm:
#### Summary
Adding requested clarification that the count of active users does not include those users which are inactive or deactivated.

#### Ticket Link
https://github.com/mattermost/docs/issues/5955


cwarnermm (approved):
Thank you, @hmmmmmmm!

PR #5968, opened 2022-10-20
Specify expected timeframe for NP application

anx-ag:
Inspired by https://forum.mattermost.com/t/license-for-non-profits-how-long-for-a-response/14182/16


cwarnermm:
@jordanafung - This PR adds a timeframe in which non-profit licensing decision will be made. Does the timeframe proposed in this PR make sense and align with the updates you're looking to make to the non-profit program in general?

PR #6007, opened 2022-10-28
MM-47632 Update steps to customize web app for products

hmhealey:
We added some stuff to that folder that isn't replaced by rebuilding the web app, so it needs to be copied over. I feel like plugins might also have something that needs to get copied over as well, but that would only apply on upgrade.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47632


cwarnermm:
@neflyte - Are you open to confirming whether similar updates are needed specifically from a plugin perspective?

cwarnermm (approved):
Thanks, @hmhealey!

PR #6014, opened 2022-11-01
Change order of tabs for desktop deloyment

esethna:
I updated the default order of the tabs on this page to match our install base (ie most installed is Windows, then mac I believe) cc// @nab-77


nab-77 (approved):
A very reasonable change IMO.

PR #6020, opened 2022-11-03
Added Desktop App config file location

cwarnermm:


source/install/desktop-app-install.rst
```diff
@@ -202,6 +200,16 @@ The following additional documentation resources are also available for the Matt
...
 ----------------------------------------------
 
+Where is configuation stored locally?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The location of the Mattermost desktop app configuration file depends on the platform where you're running Mattermost:
+
+- Windows: ``Users\<username>\AppData\Roaming\Mattermost``
+- macOS DMG: ``/Users/<username>/Library/Application Support/Mattermost``
+- macOS TestFlight: ``/Users/<username>/Library/Containers/Mattermost/Data/Library/Application Support/Mattermost``
```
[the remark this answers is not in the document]
devinbinnie: > @devinbinnie - Is there a separate path for macOS via the App Store? Or would that be the same path as TestFlight?

App Store and TestFlight should be the same.

> Also - Do users refer to the "DMG version" using other terminology? Will "DMG" be understood widely enough?

Not sure exactly, I call it that because of the file name. Another way could be to call it the 'standalone installer'. That might be better. cc @nab-77 if you have an opinion.

devinbinnie (approved):

source/install/desktop-app-install.rst
```diff
@@ -202,6 +200,16 @@ The following additional documentation resources are also available for the Matt
...
 ----------------------------------------------
 
+Where is configuation stored locally?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The location of the Mattermost desktop app configuration file depends on the platform where you're running Mattermost (and, in the case of macOS, how you've chosen to install the app):
+
+- Windows: ``Users\<username>\AppData\Roaming\Mattermost``
+- macOS standalone download: ``/Users/<username>/Library/Application Support/Mattermost``
+- macOS App Store & TestFlight: ``/Users/<username>/Library/Containers/Mattermost/Data/Library/Application Support/Mattermost``
```
nab-77, line 210: I think we could/should drop "& TestFlight". I say this as its an Apple delivery mechanism, which is basically the second of the two path variants.

source/install/desktop-app-install.rst
```diff
@@ -202,6 +200,16 @@ The following additional documentation resources are also available for the Matt
 Troubleshooting your Desktop App installation
 ----------------------------------------------
 
+Where is configuation stored locally?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The location of the Mattermost desktop app configuration file depends on the platform where you're running Mattermost (and, in the case of macOS, how you've chosen to install the app):
+
+- Windows: ``Users\<username>\AppData\Roaming\Mattermost``
+- macOS standalone download: ``/Users/<username>/Library/Application Support/Mattermost``
```
nab-77, line 209: I would go for "macOS installer" (which by definition covers both download and package type, which in this case is DMG).

cwarnermm:
Thanks, @nab-77! I'll update these changes directly in the ``v7.5-documentation`` docs branch.

PR #6033, opened 2022-11-08
Update groups-filter-sort.rst

justinegeffen:
Updated filtering section to remove "Select" constraint.

### DIFF
--- a/source/boards/groups-filter-sort.rst
+++ b/source/boards/groups-filter-sort.rst
@@ -34,7 +34,12 @@ To ungroup cards on table view, select the **Group by** option at the top of the
 Filters
 -------
 
-You can filter cards on your board using all the text properties.
+You can filter cards on your board if they utilize any of the following property types:
+- Select
+- Text
+- Email
+- Phone
+- URL
 
 .. note:: 
   
--- a/source/boards/groups-filter-sort.rst
+++ b/source/boards/groups-filter-sort.rst
@@ -38,7 +38,7 @@ You can filter cards on your board using all the text properties.
 
 .. note:: 
   
-   You must have at least one **Select** property on your board for filtering to work.
+   You must have at least one of the property types listed above on your board for filtering to work. Filtering by other property types such as **Person** or **Date** is not yet supported.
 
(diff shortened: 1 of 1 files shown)


source/boards/groups-filter-sort.rst
```diff
@@ -34,11 +34,11 @@ To ungroup cards on table view, select the **Group by** option at the top of the
 Filters
 -------
 
-You can filter cards on your board if they utilize the **Select** property.
+You can filter cards on your board using all the text properties.
```
justinegeffen, line 37: @wuwinson, is this accurate and should I also mention Filters now support all text properties! You’re no longer limited to only using filters with the Select property type. You can filter your board more granularly to see the cards most relevant to you. Boards now also supports filtering by the following property types:
- Text
- Email
- Phone
- URL

wuwinson, line 37:
```suggestion
You can filter cards on your board if they utilize any of the following property types:
- Select
- Text
- Email
- Phone
- URL
```

source/boards/groups-filter-sort.rst
```diff
@@ -34,11 +34,11 @@ To ungroup cards on table view, select the **Group by** option at the top of the
...
 
 .. note:: 
   
-   You must have at least one **Select** property on your board for filtering to work. Filtering by other property types such as **Person** or **Date** is not yet supported.
+   You must have at least one **Select** property on your board for filtering to work.
```
wuwinson, line 41:
```suggestion
   You must have at least one of the property types listed above on your board for filtering to work. Filtering by other property types such as **Person** or **Date** is not yet supported.
```

source/boards/groups-filter-sort.rst
wuwinson: Hey Justine, users can still filter by the *Select* property as well, so many some tweaks here.

wuwinson (approved):

cwarnermm (approved):
Good updates! Really like @wuwinson's clarifications as well. Approving for expediency.

PR #6042, opened 2022-11-11
Update hardware requirements

amyblais:
Based on this thread, the supported Ubuntu version is now 20.04 https://community.mattermost.com/private-core/pl/48an6do9rbyh3kggbh7iqntmpy. I'll also add a note to the changelog.

### DIFF
--- a/source/install/software-hardware-requirements.rst
+++ b/source/install/software-hardware-requirements.rst
@@ -85,7 +85,7 @@ Server software
 Mattermost server operating system
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-- Ubuntu 18.04, Debian Buster, CentOS 6+, CentOS 7+, RedHat Enterprise Linux 7+, Oracle Linux 6+, Oracle Linux 7+.
+- Ubuntu 20.04, Debian Buster, CentOS 6+, CentOS 7+, RedHat Enterprise Linux 7+, Oracle Linux 6+, Oracle Linux 7+.
 - Using the Mattermost `Docker deployment <https://github.com/mattermost/docker>`__ on a Docker-compatible operating system (Linux-based OS) is still recommended.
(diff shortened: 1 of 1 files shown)


source/install/software-hardware-requirements.rst
```diff
@@ -85,7 +85,7 @@ Server software
 Mattermost server operating system
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-- Ubuntu 18.04, Debian Buster, CentOS 6+, CentOS 7+, RedHat Enterprise Linux 7+, Oracle Linux 6+, Oracle Linux 7+.
+- Ubuntu 20.04, Debian Buster, CentOS 6+, CentOS 7+, RedHat Enterprise Linux 7+, Oracle Linux 6+, Oracle Linux 7+.
```
cwarnermm, line 88: Do we want to specify that it's Ubuntu 20.04 LTS? CC @nab-77

nab-77: @cwarnermm spot on re this being 20.04 LTS.

I commented in the channel above but want to check @agnivade that we decided that we no longer support 18.04 as of 7.5?

agnivade: I'll leave it you :) I have outlined a technical solution as a way forward. If we want to support 18.04 for 7.5, we'll need to implement the fix and backport it.

mvitale1989: For the best user experience, we should perhaps minimize the surprise factor and communicate the deprecation timeline properly, before dropping support altogether.

This means to still support 18.04 by backporting the fix as mentioned by @agnivade, but also write down both `Ubuntu 20.04 LTS` and `Ubuntu 18.04 (Deprecated, supported until April 2023)` in the supported OS list. Wdyt?

PR #6044, opened 2022-11-14
Update mark-messages-unread.rst

ayrotideysarkar:


cwarnermm (commented):
Thanks, @ayrotideysarkar. In cases where the current mobile app v2 beta release doesn't offer the same functionality available in web/desktop, we'll simply leave that topic as-is without mobile details.

With this in mind, we won't be moving forward with this PR.

ayrotideysarkar:
@cwarnermm okay... I think I have done all the channels which needed to add the mobile documentation but if there is something left... please let me know then I can work with that channel

PR #6057, opened 2022-11-17
Update groups-filter-sort.rst

justinegeffen:
Updated content in line with changes to filtering.


wuwinson (approved):
Thanks Justine! LGTM!

PR #6064, opened 2022-11-22
Use go install instead of go get

hanzei:
#### Summary
The installation of binaries using `go get` is not longer supported. The command errors with:
```sh
$ go get -u github.com/mattermost/mmctl
go: go.mod file not found in current directory or any parent directory.
	'go get' is no longer supported outside a module.
	To build and install a command, use 'go install' with a version,
	like 'go install example.com/cmd@latest'
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
```

`go install` should be used instead.

#### Ticket Link
None


justinegeffen (approved):

noxer (approved):
Good catch. We updated the build scripts and forgot about the docs. Thank you!

PR #6101, opened 2022-12-03
Fix rm with -rf to avoid "rm: missing operand"

solancer:
#### Summary
This is a fix for the "rm: missing operand" error

Screenshot 2022-12-03 at 8 28 23 PM

### DIFF
--- a/source/upgrade/upgrading-mattermost-server.rst
+++ b/source/upgrade/upgrading-mattermost-server.rst
@@ -72,7 +72,7 @@ Upgrade Mattermost Server
 
    .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


solancer:
> 

/check-cla

coltoneshaw (commented):
We actually don't need this change. What's happening here is the `find` command found nothing, and is passing nothing to `xargs` for the `rm` command so it fails. The `-f` flag will suppress this warning and make it sound like it completes, when it actually fails in the background. 

I'm wondering if adjusting the end command to `sudo xargs  --no-run-if-empty rm -r` from the [man page](https://man7.org/linux/man-pages/man1/xargs.1.html) would be a better idea.

In theory, though, based on the steps above its output you should or shouldn't run this command.

cwarnermm:
Thanks, @coltoneshaw, for your input on this docs PR. @solancer - Based on Colton's feedback, we won't be moving forward with your PR. Looking forward to working with you again in the future.

PR #6128, opened 2022-12-19
Add clarification on ICE host override setting for Calls

streamer45:
#### Summary

Adding a note explaining what could happen when using a hostname instead of an IP address as the `ICE host override` setting for Calls. This clarification resulted to be key in resolving an issue with a community user.


source/configure/plugins-configuration-settings.rst
```diff
@@ -321,6 +321,9 @@ ICE host override
 
 An optional override to the host that gets advertised to clients when connecting to calls. Depending on the network infrastructure (e.g. instance behind a NAT device) it may be necessary to set this field to the client facing external IP in order to let clients connect successfully. When empty or unset, the RTC service will attempt to automatically find the instance's public IP through STUN.
 
+.. note::
+  This field also supports a hostname (e.g. domain name) although ultimately an IP address is passed to clients. This means that a DNS resolution happens on the Mattermost instance which in certain cases could result in the resolved IP to be different from the one the clients would see, causing connectivity to fail. When in doubt, prefer using an IP address directly or make sure the resolution on the host side reflects the one on the client.
```
cwarnermm, line 325:
```suggestion
  This field also supports a hostname (e.g. domain name) although ultimately an IP address is passed to clients. This means that a DNS resolution happens on the Mattermost instance which, in certain cases, could result in the resolved IP to be different from the one the clients would see, causing connectivity to fail. When in doubt, we recommend using an IP address directly or making sure the resolution on the host side reflects the one on the client.
```

cwarnermm (commented):
Provided editorial feedback inline.

cpoile (approved):

Issue #5951, opened 2022-10-18
Help Wanted: Update format of configuration settings


cwarnermm:
**Problem space**
~ 50% of Mattermost configuration settings are now published in a new format. 

The following pages of configuration settings need to be updated to the new format:

- https://docs.mattermost.com/configure/site-configuration-settings.html (WIP)
- https://docs.mattermost.com/configure/authentication-configuration-settings.html
- https://docs.mattermost.com/configure/plugins-configuration-settings.html
- https://docs.mattermost.com/configure/integrations-configuration-settings.html
- https://docs.mattermost.com/configure/compliance-configuration-settings.html
- https://docs.mattermost.com/configure/experimental-configuration-settings.html

**Target audiences**: System administrators, Support teams, integrators, developers

**Goal**: All Mattermost configuration settings are documented consistently using the same approach and format.

**Delivery expectations**
- 1 PR for each section of configuration updates; multiple contributors can work on this ticket at the same time.
- Use the [Environment configuration settings documentation](https://docs.mattermost.com/configure/environment-configuration-settings.html) and source code as a guideline when converting settings to the new table-based format.
- Plan & deployment badges and legacy edition support details will be incorporated by the Mattermost Docs team
- Comment in the PR to request a generated preview of your work. The Docs team will add the appropriate labels to your PR.

**Tech needed**
- System Console access; note that config settings can differ between a self-hosted and Cloud-based deployment
- a copy of the ``config.json`` file is helpful for looking up the equivalent configuration setting, but this is something that the Docs team can assist with
- RST format
- Sphinx

**Affected repository**: mattermost/docs

**Where to get help**
Join the Mattermost Docs team on the Mattermost Community Server in the [~Documentation Working Group channel](https://community.mattermost.com/core/channels/dwg-documentation-working-group) to discuss this ticket, or feel free to comment in this docs ticket.

cwarnermm:
Thank you, @matthew-src! Really appreciate your help with this important work!

You've explained the work well. The tables are reStructuredText formatted. The description of the config setting is written in plain English for all audiences. 

Some existing configuration settings in the docs may be missing enough detail, such as a decent description, default values, possible values, and/or the full path to the config value in the ``config.json`` file (which also forms the environment variable). I can help fill in these missing details.

Really looking forward to your PR, @matthew-src.

matthew-src:
[...]

PR #6131, opened 2022-12-21
[GH-5951] Update Authentication > OAuth 2.0 > Office 365

matthew-src:
#### Summary

This PR updates the Office 365 section of the OAuth 2.0 settings to the table-based format. This completes the OAuth 2.0 section.

#### Ticket Link

This PR partly addresses https://github.com/mattermost/docs/issues/5951.

PR #6132, opened 2022-12-23
[GH-5951] Update Authentication > OpenID

matthew-src:
#### Summary

This PR updates the OpenID section of the Authentication configuration settings.

#### Ticket Link


cwarnermm:
Hi @matthew-src! Please see [my comments](https://github.com/mattermost/docs/pull/6131#issuecomment-1379368898) for details on how to address the merge conflicts in this PR. And please don't hesitate to reach out if I can be of assistance. Happy to help with this!

matthew-src:
I think that the conflicts are resolved, and that these metadata descriptions should be reviewed:

- Office 365 Open ID Directory tenant ID
- Office 365 Open ID Client ID (references Google account; this error is in the current documentation)
- Office 365 Open ID Client secret (also references Google account)

matthew-src:
Thank you, @cwarnermm. I updated those settings such that they would conform to the description in the table, as I did in PR https://github.com/mattermost/docs/pull/6131.

cwarnermm:
You're doing a great job adding contextual links to the configuration setting documentation, and those additions are very helpful to system admins looking for more information about a given config.

Reviewing the preview, I want to call out a few minor things about the readability of the tables:

image
- The content in brackets is currently forcing the width of the column to expand, and the second column to become less readable. 
- While the table columns auto-adjust, you can force the table width by adding a hard return earlier in the sentence. You can even break up sentences mid-link as long as the text wrapped to the next line left-aligns correctly. In this case it's bullets, so it'll be a hanging indent to preserve.
- The word "settings" is currently a link, but "settings" applies to both contexts in a given line. I recommend updating the link syntax to remove "settings" from the link itself so that it simply displays as text. This change will also help links that contain brackets.

image
- There's considerable whitespace in the left column which is forcing the right column to be less readable. This is likely due to extra spaces inserted into the left column. You can force that column's width as noted above.
- Really like the additional links to help guide the user's experience. What do you think of moving all "see also" links to a subsequent row. See ``configure/environment-configuration-settings.html#managed-resource-paths`` as an example. This would introduce more consistency across the config settings pages and draw more attention to the default value details.

cwarnermm (commented):
Thanks, @matthew-src! I've provided feedback inline.

matthew-src:
Hi @cwarnermm, I have done some work based on your feedback. I realized that the additional whitespace was caused in part by the long link addresses in the tables, although the addresses themselves are invisible when the HTML is rendered. My solution is to move the links to another row, and to shorten both the visible link text and the address. I implemented this change for the GitLab section in the latest commit. Please let me know what you think.

source/configure/authentication-configuration-settings.rst
```diff
@@ -2176,21 +2176,39 @@ Select OpenID Connect service provider
 
 *Available in legacy Enterprise Edition E20*
 
-Choose whether OpenID Connect can be used for account creation and login. Options include:
++-------------------------------------------------------------------------------+---------------------------------------------------------------+
+| Use this setting to enable OpenID Connect, with these options:                | - System Config path: **Authentication > OpenID Connect**     | 
+|                                                                               | - ``config.json`` setting: N/A                                |
+| - **Do not allow login via an OpenID provider**                               | - Environment variable: N/A                                   |
+| - **GitLab** `(settings) <#gitlab-openid-settings>`__                         |                                                               |
```
cwarnermm, line 2183: I think we need a verb in front of "settings" as a clear call to action, as "see settings".

cwarnermm:
Thanks, @matthew-src! I've provided feedback inline to update the link text.

matthew-src:
Hi @cwarnermm, I would resolve this conflict by deleting the section where the links were changed, as this will be replaced entirely by the new table. Is that acceptable? Also, should the links in the table be full URLs instead of anchors?

cwarnermm:
Hi @matthew-src. I agree with your approach to resolve the merge conflict on this PR. And all internal links between pages of the product docs site should be relative and not absolute because relative links are more robust/less prone to break when content is reorganized.

You'll see absolute URLs creep in because Sphinx supports a number of relative link types, each with their own required syntax, that contributors find overly complex and cumbersome to support.

source/configure/authentication-configuration-settings.rst
```diff
@@ -2379,43 +2438,67 @@ This value is the ID of the application's AAD directory.
...
+OpenID Connect (other) settings
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. include:: ../_static/badges/ent-pro-cloud-selfhosted.rst
+  :start-after: :nosearch:
+
+Enable OpenID Connect authentication with other service providers
+'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+.. include:: ../_static/badges/ent-pro-cloud-selfhosted.rst
```
cwarnermm, line 2500: this duplicate line should be removed

source/configure/authentication-configuration-settings.rst
```diff
@@ -2379,43 +2438,67 @@ This value is the ID of the application's AAD directory.
...
+  :start-after: :nosearch:
```
cwarnermm, line 2501: this duplicate line should be removed

cwarnermm (commented):
Thanks, @matthew-src. I've flagged 2 lines that should be removed as duplicates.

matthew-src:
@cwarnermm Got it!

cwarnermm:
My mistake! The content I thought was missing isn't missing at all; I was looking at section headers. Approving.

cwarnermm (approved):
Thanks, @matthew-src!

PR #6133, opened 2022-12-28
Fix broken docs build due to pipenv update

neflyte:
#### Summary

The docs build process has broken due to an update of `pipenv`. Since `pipenv`'s version has not been locked down, there was a risk of a build issue like this suddenly arising.

This PR fixes the following issues:
- `pipenv` is locked down to v2022.12.19 for the build pipeline
- `Pipfile.lock` is re-generated with new hashes to resolve install errors with latest `pipenv`
- `Makefile` was updated to specifically use Python 3.9 so it matches what the pipeline installs
- `Makefile` was updated to fail if there is a validation issue with `Pipfile` or `Pipfile.lock`
- `pytest` was updated to v7.2.0 to resolve CVE-2022-42969

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 # Minimal makefile for Sphinx documentation
 #
-
+.PHONY: Makefile help python-deps linkcheck livehtml python-deps test compass-icons
+#
 # You can set these variables from the command line, and also
 # from the environment for the last three.
 SOURCEDIR       = source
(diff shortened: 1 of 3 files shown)


cwarnermm (approved):
Thank you so much, @neflyte! Great catch, troubleshooting, and finding a resolution to this build issue. Also appreciate the security fixes!

PR #6137, opened 2023-01-06
Added note about moving cards to other boards

justinegeffen:
Content added to address user feedback: "couldn't find the info about moving a card to another board ? Maybe the feature doesn't exist yet ?"


wuwinson (approved):
Yep, this is not possible yet. LGTM!

PR #6148, opened 2023-01-11
Playbooks: Multiple runs

justinegeffen:
- Added content for multiple runs in a channel

### DIFF
--- a/source/playbooks/work-with-playbooks.rst
+++ b/source/playbooks/work-with-playbooks.rst
@@ -76,11 +76,12 @@ Create fewer channels
 
 From Mattermost v7.7, you can choose to start each run in a new channel or re-use the same channel.
 
-Here are some scenarios where you might want to start each run in the same channel:
+Here are some scenarios why you might want to start each run in the same channel:
 
 - Short, frequently-used processes benefit from being in the same channel - it keeps the process streamlined.
 - Teams with multiple independent workflows, such as release teams, benefit from having them in one place.
 - Cutting down on the number of new channels created makes it easier to find run channels again.
+- The run name isn't linked to the the channel's name so you can tell multiple workflows apart.
 
 When you're configuring your playbook:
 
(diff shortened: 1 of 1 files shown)


source/playbooks/work-with-playbooks.rst
```diff
@@ -71,18 +71,29 @@ Make checklists
...
+- The run name isn't linked to the the channel's name so you can tell multiple workflows apart.
+
+When you're configuring your playbook:
 
 - You can link it to an existing channel so that each run starts in that channel.
 - You can choose that each time the playbook is run, it creates a new channel.
 
 To access this setting, open the playbooks tab. Select the playbook you want to edit, then select the **Outline** tab. Select **Actions** in the left-hand menu and make your selection under the **When a run starts** heading.
 
+When you start a run, your selection is the default but can be changed for each run.
```
trilopin, line 93: would be this a good place to introduce "move run to a different channel" feature?

Something like, "Additionally, it is also possible to move a started run to another channel if that is needed"

source/playbooks/work-with-playbooks.rst
```diff
@@ -71,18 +71,29 @@ Make checklists
 
 5. Choose **Save**.
 
-Runs and channel behavior
--------------------------
+Create fewer channels
+---------------------
 
-When setting up your playbook:
+From Mattermost v7.7, you can choose to start each run in a new channel or re-use the same channel.
```
trilopin, line 77:
```suggestion
From Mattermost v7.7, you can choose to start each run in a new channel or re-use an already existing channel.
```

not sure about the exact wording, but the concept of "same channel" seems a bit awkward because when I start reading the sentence, in my mind I'm not in the context of any channel. 

Additionally, the existing-channel / new-channel are the exact words that we use in the product. Not sure if there are the best, but I'd like to use similar expressions both product and docs.

trilopin (approved):
Good job!

I left some **optional** comments

source/playbooks/work-with-playbooks.rst
justinegeffen: Good idea, thanks! :)

source/playbooks/work-with-playbooks.rst
```diff
@@ -71,18 +71,29 @@ Make checklists
...
+- The run name isn't linked to the the channel's name so you can tell multiple workflows apart.
+
+When you're configuring your playbook:
 
 - You can link it to an existing channel so that each run starts in that channel.
 - You can choose that each time the playbook is run, it creates a new channel.
 
 To access this setting, open the playbooks tab. Select the playbook you want to edit, then select the **Outline** tab. Select **Actions** in the left-hand menu and make your selection under the **When a run starts** heading.
 
+When you start a run, your selection is the default but can be changed for each run.
```
justinegeffen, line 93:
```suggestion
When you start a run, your selection is the default but can be changed for each run. Additionally, it's also possible to move a started run to another channel, so you're not locked into whichever option you select.
```

cwarnermm (approved):
This is a really great update! Well done! Love the clear focus on the user's outcome.

PR #6181, opened 2023-01-24
Update mobile-app-changelog.md

amyblais:
Add known issue.

### DIFF
--- a/source/deploy/mobile-app-changelog.md
+++ b/source/deploy/mobile-app-changelog.md
@@ -30,6 +30,7 @@ Mattermost mobile v2.0 is a major update to the iOS and Android apps bringing su
 Users now gain a more reliable and feature-rich application, improving their experience for collaborating with their teams on the go.
 
 ### Known Issues
+ - Landscape mode doesn't work on Android and some iOS devices.
  - Users are unable to adjust the font size via the OS font size setting.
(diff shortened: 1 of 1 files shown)


source/deploy/mobile-app-changelog.md
```diff
@@ -30,6 +30,7 @@ Mattermost mobile v2.0 is a major update to the iOS and Android apps bringing su
 Users now gain a more reliable and feature-rich application, improving their experience for collaborating with their teams on the go.
 
 ### Known Issues
+ - Landscape mode doesn't work on Android and iPhone devices.
```
enahum, line 33: can we remove iPhones? that was a conscious decision and is exactly like that on v1.

In fact Android phones are also locked in v1, but I guess newer Android devices (the ones that fold) can be used as tablets.. it looks to me that the ones that are "tablets" are the problem

source/deploy/mobile-app-changelog.md
```diff
@@ -30,6 +30,7 @@ Mattermost mobile v2.0 is a major update to the iOS and Android apps bringing su
 Users now gain a more reliable and feature-rich application, improving their experience for collaborating with their teams on the go.
 
 ### Known Issues
+ - Landscape mode doesn't work on newer Android devices.
```
enahum, line 33:
```suggestion
 - Landscape mode doesn't work on Android tablets.
```

example:
image

enahum (approved):

PR #6191, opened 2023-01-27
Update true-up review documentation.

ConorMacpherson:
#### Summary

Recently, some changes have been added to update the process for true-up review.
This change updates the instructions for sending Mattermost a true-up review report.

#### Ticket Link

[MM-50030](https://mattermost.atlassian.net/browse/MM-50030)

[MM-50030]: https://mattermost.atlassian.net/browse/MM-50030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


justinegeffen:
Awesome! Thanks so much for this, @ConorMacpherson! I was under the impression that this was for v7.8. Could you confirm that it's live now?

justinegeffen:
Yep! Once it's merged it's live. I'll add a milestone and label to this. :)

johndavidlugtu:
Hey @ConorMacpherson  -- good initiative here.

We would want to maintain the documentation for the old process involving having to take screenshots, since the feature you've built will only be available for customers that have version 7.9 or newer.

We'd want to have something like, "For 7.9 and older..." and "For 7.9 and newer" sections

ConorMacpherson:
Thanks @johndavidlugtu! Makes sense, I'll add the old section back and update these changes to make sure they mention 7.9 tomorrow morning :)

source/about/self-hosted-subscriptions.rst
```diff
@@ -92,18 +92,31 @@ We'll send you an email notice around the end of the quarter reminding you to se
...
+
    Not sure where to find the true-up review section? Please reach out to your account executive, Customer Success Manager, orders@mattermost.com, or support@mattermost.com for help.
-   
+
 If you have more total active users than you purchased in your annual subscription, your Customer Success Manager will provide you with a true-up quote for the new users added.
 
-The additional invoice will be pro-rated based on the number of months left in your subscription term, including the months for the calendar quarter for the time you send the report. We will not provide downward adjustments. Mattermost will invoice based on Mattermost’s `current list prices <https://mattermost.com/pricing/>`__.
+The additional invoice will be pro-rated based on the number of months left in your subscription term, including the months for the calendar quarter for the time you pull the report. We will not provide downward adjustments. Mattermost will invoice based on Mattermost’s `current list prices <https://mattermost.com/pricing/>`__.
+
+Quarterly true-up reports in version 7.9.
```
johndavidlugtu: .. or newer

ConorMacpherson: Thanks John :)

source/about/self-hosted-subscriptions.rst
```diff
@@ -99,13 +99,25 @@ To send this report, take a screenshot of **System Console > Site Statistics** a
 - We don't need your server address, so if it appears on your screenshot, you can redact it from the image.
 
 .. tip:: 
-   
-   Not sure where to take the screenshot? Please reach out to your account executive, Customer Success Manager, orders@mattermost.com, or support@mattermost.com for help.
-   
+
+   Not sure where to find the true-up review section? Please reach out to your account executive, Customer Success Manager, orders@mattermost.com, or support@mattermost.com for help.
```
johndavidlugtu, line 103: "True-up review section" might not be something they're immediately familiar with. Recommend using "Site Statistics" which is in the System Console and searchable

source/about/self-hosted-subscriptions.rst
```diff
@@ -99,13 +99,25 @@ To send this report, take a screenshot of **System Console > Site Statistics** a
...
 
+Quarterly true-up reports in version 7.9 or newer
+-------------------------------------------------
+
+Quarterly true-up reports have changed in Mattermost version 7.9.
+The reporting period is still the same, but we've introduced a process that is much easier than taking screenshots.
+
+To send the report, a System Admin must navigate to the **System Console** and open either the **Site Statistics** or **Team Statistics** pages.
+
+If the current date is within the True-Up reporting period, a panel will be visible at the top of the page with a button to share your systems statistics directly to Mattermost.
```
johndavidlugtu, line 117: How do you think we can make "True-up reporting period" clear here? We have the illustration, but it's currently in the "prior to version 7.9", which is also needed there.
Screenshot 2023-01-31 at 6 40 02 PM

ConorMacpherson: I suppose two options:

1. Make a reference to the image, saying something like:
  - `If the current date is within the True-Up reporting period outlined in the previous section...`
2. Re-use the image in the new section

Thoughts?

johndavidlugtu: Ya re-using might be easier here

justinegeffen:
Documentation for https://github.com/mattermost/mattermost-webapp/pull/11844

source/about/self-hosted-subscriptions.rst
```diff
@@ -82,8 +83,8 @@ This process adds additional users to your existing subscription and is not a ne
 
 If you don't have a self-hosted license in the Customer Portal and want to add users, `contact us <https://mattermost.com/contact-us/>`__.
 
-Quarterly true-up report
-------------------------
+Quarterly true-up report prior to v7.9
+--------------------------------------
 
 When you buy an annual Mattermost subscription, you agree to provide quarterly reports of the actual number of registered users within your system. A registered user is a user who has an account in a workspace and does not show as **Inactive** in **System Console > User Management > Users**.
```
justinegeffen, line 89: We don't refer to "registered users" anywhere else - is this meant to be "active users"? :)

ConorMacpherson: I think "active users" makes sense here. Just an FYI, I found a few other instances that could use a change (in another pr) if we want to avoid "registered users"

https://github.com/mattermost/docs/search?q=%22registered+user%22%2C%22registered+users%22

johndavidlugtu: Heads up I think the source of that "registered users" language comes from our legal terms: https://mattermost.com/software-services-license-agreement/ (Section 3.3b)

"Activated users" in my opinion is the one that best describes these users, but I believe since our terms still reflect "Registered Active Users", "active users" is the next best thing

justinegeffen (approved):

PR #6230, opened 2023-02-15
Added mobile data retention details

cwarnermm:


justinegeffen (approved):

source/channels/channels-settings.rst
```diff
@@ -206,6 +206,14 @@ Select which language Mattermost displays in the user interface. Options include
 - 中文 (繁體) - Traditional Chinese
 - 日本語 - Japanese
 
+Advanced settings
+~~~~~~~~~~~~~~~~~
+
+Using the Mattermost mobile app, you can delete local Mattermost files from your device. Access **Settings** by tapping on your profile picture.
```
larkox, line 212: Not sure if clear all the different clicks that have to be made. You have to click the profile picture and then click on Settings. But maybe it is clear as it is phrased right now.

larkox, line 212: Maybe makes sense to clarify that the only stuff deleted will be the stuff regarding the actual server, not regarding every single server registered on the app.

larkox (commented):
Some comments.

source/channels/channels-settings.rst
```diff
@@ -206,6 +206,14 @@ Select which language Mattermost displays in the user interface. Options include
...
+
+  1. Tap **Advanced Settings**.
+  2. Tap **Delete local files**.
```
cwarnermm, line 215:
```suggestion
  2. Tap **Delete local files**. Only data specific to the current server is removed from your device. You'll need to repeat this process for each server you've configured on the mobile app.
```

source/channels/channels-settings.rst
cwarnermm: This is a standard approached we're using to walk mobile users to their end goals. We'll likely revisit this approach in a future release.

cwarnermm: @larkox - I've drafted an update to clarify what data is/isn't deleted with this action. Anything you'd like to see changed?

larkox (approved):

PR #6251, opened 2023-03-03
Remove notification banner

emdecr:
#### Summary
This PR removes the notification banner since the GitLab webinar has passed. I've commented the element out – for when we decide we need to turn it back on.

Thanks!

### DIFF
--- a/source/_static/myscript.js
+++ b/source/_static/myscript.js
@@ -170,20 +170,24 @@ $(document).ready(function () {
 		}
 	});
 	// Notification Banner
-	if (localStorage.getItem("docsFeedback") === null) {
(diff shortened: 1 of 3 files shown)


emdecr:
Ooo. Give me like 5 mins...there's some styling I need to fix hah!

cwarnermm (approved):
Thanks, @emdecr! This PR also serves as a great blueprint when we want to disable in the future.

emdecr:
Np @cwarnermm ! I also realized there's a few things we can tweak re: how we were capturing whether a user closed the banner, but I'll work that into the next time we change it.

PR #6254, opened 2023-03-06
[GH-5951] Update Plugins > Calls settings

matthew-src:
#### Summary

This PR updates the Calls section of the Plugins settings page to the new table-based format.

#### Ticket Link

#### Notes

Please take notice of the ICE Host override setting. There is an unusual amount of text that I attempted to consolidate into a **Note** section. ~~Also, the table for the Maximum call recording duration setting is not rendering in my preview environment. It seems there is a problem with the markup somewhere, but I cannot find it.~~ [I fixed the problem just after making the PR.]


cwarnermm:
Hi @matthew-src! One Calls config setting isn't displaying table data:
image

matthew-src:
> Hi @matthew-src! One Calls config setting isn't displaying table data: image

This has been fixed in the latest commit.

matthew-src:
> @streamer45 - Do Calls config settings have equivalent `config.json` values, or are all Calls config settings available only in the System Console UI?

As far as I can tell, it does not have any `config.json` values. I think this is confirmed by the RST metadata, which also does not list any `config.json` values or environment variables.

streamer45:
> > @streamer45 - Do Calls config settings have equivalent `config.json` values, or are all Calls config settings available only in the System Console UI?
> 
> As far as I can tell, it does not have any `config.json` values. I think this is confirmed by the RST metadata, which also does not list any `config.json` values or environment variables.

Calls is a plugin so it has it's own section in `config.json` but it's a dynamic object meaning that settings are only persisted once they get saved. See example below:

```json
    "PluginSettings": {
        "Enable": true,
        "EnableUploads": true,
        "AllowInsecureDownloadURL": false,
        "EnableHealthCheck": true,
        "Directory": "./plugins",
        "ClientDirectory": "./client/plugins",
        "Plugins": {
            "com.mattermost.calls": {
                "allowenablecalls": false,
                "allowscreensharing": true,
                "defaultenabled": true,
                "enablerecordings": true,
                "icehostoverride": "0.0.0.0",
                "iceserversconfigs": "",
                "jobserviceurl": "http://localhost:4545",
                "maxcallparticipants": 0,
                "maxrecordingduration": 15,
                "rtcdserviceurl": "http://rtcd-local:8045",
                "serversideturn": false,
                "turncredentialsexpirationminutes": 1440,
                "turnstaticauthsecret": "secret",
                "udpserveraddress": "",
                "udpserverport": 8443
            },
```

So they can be changed through config file although the recommendation would be to do it through UI unless users know what they are doing.

cwarnermm (approved):
Great job all around, @matthew-src! Thank you! I like how you addressed the note for the ICE Host Override setting, and I can confirm that the one issue raised during first review is resolved.

cwarnermm:
Thanks, @streamer45 for the context around the Calls config settings in the ``config.json`` file. We'll keep the docs as-is and only list the System Console UI paths for simplicity.

@svelle - Given that the Calls config settings _can_ be managed via the ``config.json`` once they're dynamically added to the file, would a Support Knowledge Base article help guide admins who are looking to do this (since they won't find this documented in the product docs)?

svelle:
> (since they won't find this documented in the product docs)

Why isn't this documented in the product docs, or rather, why shouldn't this be? 
The KB, in my opinion, is not the right place to put configs unless they're meant to fix something that is broken or commonly set up in a wrong way.

PR #6257, opened 2023-03-08
Update notification banner

emdecr:
#### Summary
This PR updates the notification banner for the MM/MST LP.
As discussed, the targeted merge for this PR is tomorrow, Mar 9th @ 9am EST.

cc: @cwarnermm @azigler @nat-gunner for visibility - adding everyone as a reviewer for more coverage

developers.mm.com PR coming shortly.

Let me know if you have any Qs. TY!


cwarnermm:
Do we want to open banner links within the same browser window or a new browser window?

azigler (approved):
I think a new tab is best, just because it's to a whole other domain and we should assume they were going to `docs` for a reason and we should leave the tab alone. But I don't feel very strongly! 1/5

emdecr:
Awesome, thanks y'all – let's go with the new tab 👍

PR #6258, opened 2023-03-09
[GH-5951] Update Plugins > Channel Export and Demo Settings

matthew-src:
#### Summary

This PR updates the Plugins > Channel Export and Demo settings to the table-based format.

#### Ticket Link

This PR partly addresses https://github.com/mattermost/docs/issues/5951.

PR #6259, opened 2023-03-09
[GH-5951] Update Plugins > GIF commands

matthew-src:
#### Summary

This PR updates the Plugins > GIF commands settings to the table-based format.

#### Ticket Link


cwarnermm:
Thanks, @matthew-src! 

As noted via https://github.com/mattermost/docs/pull/6258#issuecomment-1464083213, in this PR, can you please the ``N/A`` from both the table-based content as well as the search metadata preceding each docs section? The ``N/A`` tells reads that these equivalents don't exist when they do which is misleading. It's better to have a content gap in the docs than than to publish incorrect details.

cwarnermm (approved):
Thanks, @matthew-src!

PR #6264, opened 2023-03-10
Remove CTA button in nav

emdecr:
#### Summary

This PR removes the Get Started CTA button in the navigation. We just changed the link to `/download/` on the mm.com site to pivot towards SH. But for docs if we're on the installation pages, we're creating a circular journey – and for all the other pages the docs are mostly targeted towards current users - so download may not be necessary (?). There's also the possibility to use the button for another CTA if one works better on docs.

Wanted to push this up to open the discussion here – tagging @aaron-l-thompson to chime in with any other details.

TY~

### DIFF
--- a/source/templates/header.html
+++ b/source/templates/header.html
@@ -410,11 +410,8 @@
 							</figure>
 						</div>
 					</li>
-					<li class="site-nav__toplink site-nav__button">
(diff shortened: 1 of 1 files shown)


cwarnermm:
This change makes a lot of sense, @emdecr! No concerns from Docs and happy to merge when Marketing is ready.

cwarnermm (approved):

PR #6265, opened 2023-03-10
Fix wrong query

anx-ag:
related to pr-6263


cwarnermm (approved):
Thanks, @anx-ag! Next steps include a technical review of the change by Engineering.

harshilsharma63:
This is incorrect and not needed anymore as we're moving that SQL to an in-product migration. Thanks for the PR anyways!

PR #6270, opened 2023-03-14
Removed manual SQL data check as the fix is now in-code

harshilsharma63:
#### Summary
Removed a SQL instruction to manually check and fix data before upgrading. The fix is now in-code so nothing needs to be done manually.

Related PR with fix added in-code to 7.8 - https://github.com/mattermost/focalboard/pull/4639


cwarnermm (approved):
Thanks so much, @harshilsharma63!

PR #6272, opened 2023-03-14
Update send-messages.rst

justinegeffen:
Docs for https://github.com/mattermost/mattermost-webapp/pull/11069

### DIFF
--- a/source/channels/send-messages.rst
+++ b/source/channels/send-messages.rst
@@ -65,6 +65,11 @@ Select **Edit** to edit your own messages. Editing a message won't trigger new @
 
 Select **Delete** to delete your own messages. Select **Delete** again to confirm.
 
+From Mattermost v7.9, you can view the edit history of an edited message and restore a previous version of an edited message.
+
+- **To view edit history:** Hover over the word *Edited* next to your edited message. In the right-hand sidebar, you'll see all previous versions of the message.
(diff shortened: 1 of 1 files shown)


cwarnermm:
@esethna - Is this feature available in mobile v2.x as well? Or only via web/desktop for now?

cwarnermm (approved):
Thanks, @justinegeffen. Approved assuming no mobile updates are needed.

esethna:
Webapp only

source/channels/send-messages.rst
```diff
@@ -65,6 +65,11 @@ Select **Edit** to edit your own messages. Editing a message won't trigger new @
 
 Select **Delete** to delete your own messages. Select **Delete** again to confirm.
 
+From Mattermost v7.9, you can view the edit history of an edited message and restore a previous version of an edited message.
```
esethna, line 68: > From Mattermost v7.9

Do we need to state this?

justinegeffen: We usually do if the change doesn't apply to previous versions and the customer needs to upgrade to use it.

cwarnermm: @esethna One common piece of feedback we receive from docs site visitors is that a given feature isn't available to them (because they're on an older version of Mattermost).

One clear way forward is to set expectations up front about the minimum release needed to access a feature. Another, longer term approach could be versioned documentation.

PR #6292, opened 2023-03-19
Add team edition download link

anx-ag:
To match the Upgrade Documentation where we already added this. Also fix spelling at the bottom.


cwarnermm (approved):
Thanks, @anx-ag!

PR #6293, opened 2023-03-20
Fix wrong filename and format

anx-ag:
This command does not work, the processed import file always gets an ID generated and prepended to the file and needs to be specified for the import process command. Additionally, JSONL Files cannot be uploaded directly, only ZIP files are supported.

The preview didn't work for me here, so I'm not 100% sure I got it right as far as formatting is concerned, sorry.

### DIFF
--- a/source/onboard/run-bulk-loading-command.rst
+++ b/source/onboard/run-bulk-loading-command.rst
@@ -10,11 +10,11 @@ Before running the bulk loading command, you must first create a `JSONL <https:/
 
   .. tab:: Use mmctl
 
-    1. After you create the file, upload the file to the database by running the `mmctl import upload </manage/mmctl-command-line-tool.html#mmctl-import-upload>`__ command. For example: ``mmctl import upload data.jsonl``.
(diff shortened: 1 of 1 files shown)


source/onboard/run-bulk-loading-command.rst
```diff
@@ -10,11 +10,11 @@ Before running the bulk loading command, you must first create a `JSONL <https:/
 
   .. tab:: Use mmctl
 
-    1. After you create the file, upload the file to the database by running the `mmctl import upload </manage/mmctl-command-line-tool.html#mmctl-import-upload>`__ command. For example: ``mmctl import upload data.jsonl``.
+    1. After you created the JSONL file, you need to zip it (`zip -r data.zip data.jsonl`) and upload the ZIP file to the database by running the `mmctl import upload </manage/mmctl-command-line-tool.html#mmctl-import-upload>`__ command. For example: ``mmctl import upload data.zip``.
```
cwarnermm, line 13:
```suggestion
    1. After you create the JSONL file, you need to zip it by running the ``zip -r data.zip data.jsonl`` command, and upload the ZIP file to the database by running the ``mmctl import upload </manage/mmctl-command-line-tool.html#mmctl-import-upload>``__ command. For example: ``mmctl import upload data.zip``.
```

PR #6294, opened 2023-03-20
Remove auth_data from example user object

anx-ag:
When trying to import the example user object, the import process fails with the error message:

Data: map[error:Error during job execution. — BulkImport: User AuthData and Password are mutually exclusive. 

The example is not valid, so I suggest to remove the "auth_data" option here.


cwarnermm (approved):
Thanks, @anx-ag!

PR #6297, opened 2023-03-20
Update cloud-subscriptions.rst

justinegeffen:
Cloud deletion steps.

### DIFF
--- a/source/about/cloud-subscriptions.rst
+++ b/source/about/cloud-subscriptions.rst
@@ -176,6 +176,6 @@ These offerings are deployed in the following ways:
 Can I cancel and delete my Cloud workspace?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Yes. If you have a Mattermost Free or a  monthly Mattermost Enterprise plan, you can delete your workspace via **System Console > Billing & Account > Subscription**.
+Yes. If you have a Mattermost Free or a monthly Mattermost Professional plan, you can delete your workspace via **System Console > Billing & Account > Subscription**.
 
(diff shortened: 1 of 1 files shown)


source/about/cloud-subscriptions.rst
```diff
@@ -177,4 +176,6 @@ These offerings are deployed in the following ways:
 Can I cancel and delete my Cloud workspace?
```
justinegeffen, line 176: @johndavidlugtu, was there a decision around whether the customer portal account deletion is offered? And what messaging do end-users see when they try to access the deleted workspace?

johndavidlugtu: Yep, the decision was to not offer it at this point, as we've theorized that the only a very few subset of workspace deletion requests genuinely want their portal account deletion. Since it was offered as a question from support, we suspected users just marked it as "Yes"

source/about/cloud-subscriptions.rst
```diff
@@ -177,4 +176,6 @@ These offerings are deployed in the following ways:
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Yes. Please contact our Mattermost `Support team <mailto:support@mattermost.com>`_ with the subject **Mattermost Cloud Workspace Deletion Request** for assistance.
+Yes. If you have a Mattermost Free or a  monthly Mattermost Enterprise plan, you can delete your workspace via **System Console > Billing & Account > Subscription**.
```
johndavidlugtu, line 179: monthly *Professional plan

johndavidlugtu (changes requested):

source/about/cloud-subscriptions.rst
justinegeffen, line 179:
```suggestion
Yes. If you have a Mattermost Free or a monthly Mattermost Professional plan, you can delete your workspace via **System Console > Billing & Account > Subscription**.
```

PR #6304, opened 2023-03-23
Remove duplicated paragraph

agarciamontoro:
#### Summary
I was reading this page and noticed that the first two paragraphs are almost identical. I removed the first one here, since the second reads a bit more natural to me, but I'm 0/5 on which to maintain.

#### Ticket Link
--


cwarnermm (approved):
Thanks, @agarciamontoro!

PR #6317, opened 2023-04-06
[MM-51797] Add bullet point on calls transmission security

streamer45:
#### Summary

PR adds a bullet point to our transmission security section to address the security of calls.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51797


streamer45:
@jupenur Would be great if you could give this a quick review to make sure everything is sound from a security perspective. Let me know :)

source/about/security.rst
```diff
@@ -35,6 +35,12 @@ Transmission security
 ~~~~~~~~~~~~~~~~~~~~~~
 
 - Mattermost supports TLS encryption using AES-256 with 2048-bit RSA on all data transmissions between Mattermost client applications and the Mattermost server across both LAN and internet.
+- Connections to calls are secured with a combination of:
```
jupenur, line 38: 0/5 is it clear enough that this is an exception to the above?

jupenur (approved):
Apologies for the delay, I was on PTO. Just one small comment, otherwise LGTM :)

source/about/security.rst
cwarnermm: @streamer45 ^

source/about/security.rst
```diff
@@ -35,6 +35,12 @@ Transmission security
 ~~~~~~~~~~~~~~~~~~~~~~
 
 - Mattermost supports TLS encryption using AES-256 with 2048-bit RSA on all data transmissions between Mattermost client applications and the Mattermost server across both LAN and internet.
```
cwarnermm, line 37:
```suggestion
- Mattermost supports TLS encryption using AES-256 with 2048-bit RSA between Mattermost client applications and the Mattermost server across both LAN and internet.
```

PR #6329, opened 2023-04-13
Document upgrade best practices

agnivade:
#### Summary

Documenting our guidelines on upgrades


source/upgrade/prepare-to-upgrade-mattermost.rst
```diff
@@ -10,6 +10,13 @@ In most cases, you can `upgrade Mattermost Server </upgrade/upgrading-mattermost
   :backlinks: top
   :local:
 
+Upgrade Best Practices
+----------------------
+
+Mattermost will aim to have non-locking, backwards-compatible migrations in general. But the backwards compatibility guarantee extends to only the last ESR version. For example, if there are three ESR versions ESR1, ESR2, and ESR3, upgrading from ESR1 to ESR2, and then ESR2 to ESR3 will ensure backwards compatibility, but not from ESR1 to ESR3 directly.
```
nab-77, line 16:
```suggestion
Mattermost will aim to have non-locking, backwards-compatible migrations in general. This backwards compatibility guarantee extends to only the last ESR version. For example, if there are three ESR versions ESR1, ESR2, and ESR3, upgrading from ESR1 to ESR2, and then ESR2 to ESR3 will ensure backwards compatibility, but not from ESR1 to ESR3 directly.
```

source/upgrade/prepare-to-upgrade-mattermost.rst
```diff
@@ -10,6 +10,13 @@ In most cases, you can `upgrade Mattermost Server </upgrade/upgrading-mattermost
...
+
+Our recommendation is to upgrade often, so as to not fall behind on upgrades. And if you fall behind, upgrade to the closest ESR version first, and from there to the next ESR. Do not attempt to directly upgrade to the latest version as it might break backwards compatiblity of the older nodes in the cluster during the upgrade.
```
nab-77, line 18:
```suggestion
Our recommendation is the case of delayed upgrades that you should, at a minimum, upgrade to the closest ESR version first, and from there to the next ESR. Do not attempt to directly upgrade to the latest version as it might break backwards compatibility of the older nodes in the cluster during the upgrade.
```

nab-77 (approved):
Few minor edits

source/upgrade/prepare-to-upgrade-mattermost.rst
agnivade: @nab-77 - Did you mean "Our recommendation is, _in_ the case of delayed upgrades that you should, ..."?

cwarnermm: @agnivade @nab-77 - I've proposed an update to this text that simplifies the message. Let me know if the original meaning is intact.

agnivade: ping @nab-77

PR #6335, opened 2023-04-16
App Bar link + naming

anx-ag:
I recognized that the link to the system console settings for the appbar was not working and it seems as if the correct name would be "App bar" instead of "Apps bar"


anx-ag:
Wrong: </configure/configuration-settings.html#enable-app-bar> 
Right: </configure/experimental-configuration-settings.html#enable-app-bar>

You can also reject my PR and just fix this link if the wording is still under discussion, should probably have created two PRs and not merge both topics into one, sorry.

justinegeffen:
Thanks again for this PR, @anx-ag! We've restarted a discussion around the naming convention, [here](https://community.mattermost.com/core/pl/ky1msapy9jn68cmajk7b716q8a). I'm closing this specific PR but changes will be made in the future. :)

PR #6343, opened 2023-04-18
[MM-51839] Link to ice host override setting documentation

streamer45:
#### Summary

The addition I originally wanted to make was already there (second note in https://docs.mattermost.com/configure/plugins-configuration-settings.html#ice-host-override), just not where I was looking for. I still think the current place is the right one so simply adding some minor linking from the deployment page to facilitate discovery.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51839


cpoile (approved):

cwarnermm (approved):
Great update! Thanks, @streamer45.

PR #6361, opened 2023-04-27
Trace logging doesn't apply to SAML

matthew-src:
This change addresses an issue for 7.11 from the docs task board.


cwarnermm (approved):
Thanks, @matthew-src!

PR #6374, opened 2023-05-02
Removed Mattermost Boards and Playbooks from plugins config page

matthew-src:
#### Summary

This PR removes the Mattermost Boards and Playbooks content from the plugins config page ahead of the monorepo release.


cwarnermm:
@wuwinson @itao - Now that the Boards and Playbooks verticals are merged into the mono-repo, is it possible to run Boards or Playbooks as plugins anymore? 

Looking for confirmation that there's no longer a need to have plugin-specific configuration settings documented for either Boards or Playbooks, and that these settings can safely be removed.

wuwinson:
> @wuwinson @itao - Now that the Boards and Playbooks verticals are merged into the mono-repo, is it possible to run Boards or Playbooks as plugins anymore?
> 
> Looking for confirmation that there's no longer a need to have plugin-specific configuration settings documented for either Boards or Playbooks, and that these settings can safely be removed.

It should not be possible to run the plugins anymore after v7.11. cc @sbishel @lieut-data just to be 100% sure.

lieut-data:
@wuwinson, that's correct!

cwarnermm (approved):
Thanks, @matthew-src!

PR #6375, opened 2023-05-02
Other MFA apps can be used instead of Google Authenticator

matthew-src:
#### Summary

This PR revises MFA documentation to remove exclusive references to Google Authenticator in order to inform users that other apps can be used to implement MFA.


cwarnermm (approved):
Thanks, @matthew-src!

PR #6381, opened 2023-05-05
Capture ids from in-product linkouts

emdecr:
#### Summary
This PR adds functionality to capture IDs from in-product linkouts, which we can then reference in analytics if they take certain actions down the funnel. The same mechanism will be in place on mm.com.

Still a work in progress – cc: @cwarnermm for visibility and if there's any questions first-blush. TY!


emdecr:
Closing because we're going to use Google Tag Manager vs hard-coding in the codebase for now. Will create new PR, if anything. TY!

PR #6390, opened 2023-05-09
Update common-support-packet.rst

grubbins:
It was really unclear that "Commercial Support" is not a section in the System Console, but a separate menu item... screenshot would also be great!

### DIFF
--- a/source/manage/common-support-packet.rst
+++ b/source/manage/common-support-packet.rst
@@ -8,7 +8,7 @@ Use the System Console to generate a Mattermost Support Packet that includes con
 
 To generate a Support Packet:
 
-1. Go to System Console, then select **Commercial Support**.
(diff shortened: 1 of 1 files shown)


source/manage/common-support-packet.rst
```diff
@@ -8,7 +8,7 @@ Use the System Console to generate a Mattermost Support Packet that includes con
 
 To generate a Support Packet:
 
-1. Go to System Console, then select **Commercial Support**.
+1. Go to System Console, then go to the menu (from the menu icon at upper left), and select **Commercial Support**.
```
cwarnermm, line 11:
```suggestion
1. Go to System Console and select **Commercial Support** from the System Console menu. 
```

cwarnermm (approved):

PR #6391, opened 2023-05-09
Update install-tar.rst

nab-77:
Small amendment to the config.json step due to release asset variance.


cwarnermm (approved):
Thanks, @nab-77!

PR #6393, opened 2023-05-10
Update index.html

matthew-src:
Changed /sign-up link to /download based on docs board task


cwarnermm:
@emdecr - We did a sweep of the docs repository looking for any outstanding ``/sign-up`` links that should be updated to ``/download``. This PR changes the one instance we found. Can you please review?

emdecr (approved):
Thanks for the sweep y'all! Did a search as well for the `sign-up` landing page on mm.com and there doesn't seem to be any instances 👍 

cc: @cwarnermm

PR #6419, opened 2023-05-18
Remove legacy references in Channels

matthew-src:
#### Summary

This PR removes legacy references in the Channels docs directory.


cwarnermm (approved):
Thanks, @matthew-src!

PR #6438, opened 2023-05-29
Streamlined install steps for quick scanning

cwarnermm:
- Reduced mini-ToC depth to 1
- Removed unnecessary details
- introduced tab-based content to clarify user paths

Note: Updated Ubuntu, tarball, Docker, and Omnibus only; remaining deploy methods will be updated in a future PR.


nab-77 (approved):
Great change ahead of the ToC re-org (with some great ideas to carry into this new stucture).

PR #6456, opened 2023-06-12
Update PostgreSQL DB config docs

cwarnermm:
- Clarified version-specific details on this page where needed.
- Made this page easier to find via the navigation pane & Deployment landing page.
- Updated formatting for CLI commands to make them easier to copy/paste/scan.

@emdecr - Can I get your help to format the ``code`` on this page, via tabs, to match the ``code`` format we're using for our installation pages, please? (see example below) I suspect extra CSS is needed due to this content being tab-based.

image


emdecr:
@cwarnermm Sure thing! Will add to the queue and circle-back if I think it's a bigger lift than I'm thinking. Pressing due date on this, or do we have some buffer? TY!

nab-77 (changes requested):
Only one edit. Wondering if we can format as we have done with recent install doc (i.e. true left-align). Feels easier to read.

cwarnermm:
@nab-77 - Yes, we will be able to do exactly that once we removed the tabbed approach on this page.

PR #6457, opened 2023-06-12
fix: docker compose command

matinzd:
#### Summary

Fix docker compose command

#### Ticket Link

N/A


coltoneshaw (approved):
These changes would work on all machines, if I understand correctly. `docker compose` is the new version of `docker-compose`. The only hurdle is if someone hasn't updated their docker package recently they might not be able to run this. But they can do a quick google search.

matinzd:
> These changes would work on all machines, if I understand correctly. `docker compose` is the new version of `docker-compose`. The only hurdle is if someone hasn't updated their docker package recently they might not be able to run this. But they can do a quick google search.

Yeah, but it's better to keep the document updated to the latest version of docker compose plugin.

PR #6470, opened 2023-06-14
Updated audit logging docs to include multi-line JSON & env var support

cwarnermm:
Documentation for:
- https://github.com/mattermost/mattermost/pull/23324
- https://github.com/mattermost/mattermost/pull/23610

### DIFF
--- a/source/comply/audit-log.rst
+++ b/source/comply/audit-log.rst
@@ -105,53 +105,6 @@ Examples of values for the ``AdvancedLoggingJSON`` setting are:
             }
         }
 
-3. Environment variable:
-
-  .. code-block:: shell
-
-    export MM_LOGSETTINGS_ADVANCEDLOGGINGJSON=$(jq -n -c '{
-      "file1": {
-        "Type": "file",
-        "Format": "json",
-        "Levels": [
-          {"ID": 5, "Name": "debug", "Stacktrace": false},
-          {"ID": 4, "Name": "info", "Stacktrace": false},
-          {"ID": 3, "Name": "warn", "Stacktrace": false},
-          {"ID": 2, "Name": "error", "Stacktrace": true},
-          {"ID": 1, "Name": "fatal", "Stacktrace": true},
-          {"ID": 0, "Name": "panic", "Stacktrace": true}
-        ],
-        "Options": {
-          "Compress": true,
-          "Filename": "mattermost_logr.log",
-          "MaxAgeDays": 0,
-          "MaxBackups": 10,
-          "MaxSizeMB": 10 
-        },
-        "MaxQueueSize": 1000
-      }
-    }')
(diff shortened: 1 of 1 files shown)


source/comply/audit-log.rst
```diff
@@ -24,11 +24,17 @@ Configure audit logging
 
 Configuring Mattermost to enable audit logging requires editing the ``config.json`` file directly. Audit logging can’t be managed using the System Console.
 
-In the ``config.json`` file, go to the ``ExperimentalAuditSettings`` section. Within the ``AdvancedLoggingConfig`` setting, you can specify an absolute or relative filespec to another configuration file or a JSON string. The process of configuring audit logging includes specifying destination targets, event names to include, and the verbosity of the audit log output.
+In the ``config.json`` file, go to the ``ExperimentalAuditSettings`` section. Within the ``AdvancedLoggingConfig`` setting, you can specify an absolute or relative filespec to another configuration file, embed multi-line JSON in the ``config.json`` file, or an environment variable. The process of configuring audit logging includes specifying destination targets, event names to include, and the verbosity of the audit log output.
```
wiggin77, line 27: `AdvancedLoggingConfig` is now deprecated and will be removed in a future release.  The new field is `AdvancedLoggingJSON` and only that field allows multi-line JSON.

AdvancedLoggingConfig (deprecated):  escaped single line JSON string  OR  filespec
AdvancedLoggingJSON:  multi-line JSON  OR  filespec   (allow supports the escaped single line JSON string but I don't think that should be documented)

cwarnermm: I completely missed that important nuance. Thank you, @wiggin77!

source/comply/audit-log.rst
```diff
@@ -68,18 +77,80 @@ The example JSON configuration specifies two log targets: one outputs to the con
...
+                },
+                "MaxQueueSize": 1000
+            }
+        }
+
+3. Environment variable:
+
+  .. code-block:: shell
+
+    export MM_LOGSETTINGS_ADVANCEDLOGGINGJSON=$(jq -n -c '{
```
wiggin77, line 112: `MM_LOGSETTINGS_ADVANCEDLOGGINGJSON` doesn't apply to audit logging.  That is the config for general logging.  Not sure if having it here is helpful or confusing.  It may be better suited to an equivalent page documenting general logging settings.

wiggin77 (approved):
LGTM :+1:   I'll leave it to you to decide if the general logging example (MM_LOGSETTINGS_ADVANCEDLOGGINGJSON) should remain here.

source/comply/audit-log.rst
cwarnermm: Thanks, @wiggin77 -- I agree that leaving this section in will introduce confusion, so best to remove. Would you be open to drafting an outline of a general logging docs page where this content would eventually live?

PR #6487, opened 2023-06-20
Fix download links

bbodenmiller:
#### Summary

- Fix download links 
- Put them in consistent location for each release

### DIFF
--- a/source/install/desktop-app-changelog.rst
+++ b/source/install/desktop-app-changelog.rst
@@ -18,7 +18,7 @@ Release v5.4
 
 **Release Day: June 19, 2023**
 
-**Download Binaries:** `Mattermost Desktop on GitHub <https://github.com/mattermost/desktop/releases/latest>`_
+**Download Binaries:** `Mattermost Desktop on GitHub <https://github.com/mattermost/desktop/releases/tag/v5.4.0>`_
 
 Compatibility
 ~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


source/install/desktop-app-changelog.rst
```diff
@@ -18,7 +18,7 @@ Release v5.4
 
 **Release Day: June 19, 2023**
 
-**Download Binaries:** `Mattermost Desktop on GitHub <https://github.com/mattermost/desktop/releases/latest>`_
```
amyblais, line 21: I don't have context on why we want to remove the "latest" link. Other changes in the PR look good to me.

bbodenmiller: @amyblais might be okay for latest release but seems like historically folks have forgot to update this link when next release comes out so less error prone to point at release directly.

amyblais: Makes sense, I think we can leave the "latest" link here, and I'll add to the release checklist to update the link when a new release is out.

amyblais: If the submitter is ok with it, we can leave the "latest" link here, and I'll add to the release checklist to update the link when a new release is out.

amyblais: If the submitter is ok with it, we can leave the "latest" link here, and I'll add to the release checklist to update the link when a new release is out.

source/install/desktop-app-changelog.rst
```diff
@@ -18,7 +18,7 @@ Release v5.4
...
+**Download Binaries:** `Mattermost Desktop on GitHub <https://github.com/mattermost/desktop/releases/tag/v5.4.0>`_
```
bbodenmiller, line 21:
```suggestion
**Download Binaries:** `Mattermost Desktop on GitHub <https://github.com/mattermost/desktop/releases/latest>`_
```

source/install/desktop-app-changelog.rst
bbodenmiller: @amyblais sounds good updated

amyblais (approved):

amyblais:
Thank you!

PR #6498, opened 2023-06-26
Transitioning Boards to plugin

cwarnermm:
This PR is for pulling out Boards content from docs.mm.com

Content being migrated to GitHub via https://github.com/mattermost/focalboard/pull/4862


esethna (changes requested):
@cwarnermm thanks for this! Hard to track all the changes so I just took a look in the preview env. Here's some high level thoughts:

1. I think the LHS section title for "Mattermost boards plugin" could be "Focalboard". The intro can describe that the focalboard plugin for Mattermost enables managing projects etc... with a link to the focalboard repo.

2. Then we can refer to all instances of the "Boards plugin" as the "Focalboard plugin" 

3. The plugin settings doc should have a subsection for enabling the plugin (since it's disabled by default in v8.0)

cwarnermm:
Hi @esethna! This PR should be ready for your final review. Thanks!

esethna (approved):
Thank you Carrie!

PR #6514, opened 2023-07-12
CLD-5793 - mmctl release changes

mvitale1989:
#### Summary

Update mmctl related docs, due to the many changes introduced with release v8.0.0.

A few notes:
* Previously, the build docs were in the [mmctl repository](https://github.com/mattermost/mmctl). But after the monorepo move, I can't find any such docs within the monorepo. However the mmctl documentation page already contains such instructions, and this PR updates it
* The `go install` command in the PR currently doesn't work. Jesse is taking a look ([discussion](https://community.mattermost.com/private-core/pl/cdhikbmjhbdumd9shez5h5sy5w)), but if it turns there's a dealbreaker we can just remove it. I'll keep the `Do not merge` label until we understand this
* homebrew is a community effort, we don't support it, so i added a note for that ([discussion](https://mattermost.atlassian.net/browse/CLD-5793?focusedCommentId=152974))
* The mmctl unit tests actually require the `enterprise` repo to be checked out at the root directory of the monorepo. Do we want to keep mentioning it in these docs? Or shall we remove it, and keep the unit tests internal?

I think in general there should be a bigger effort around the documentation, regarding the monorepo move. There may be a lot of obsolete command that don't work anymore.

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-5793


cwarnermm (approved):
Thank you so much for this great docs PR, @mvitale1989! I really appreciate your help keeping this content current and valuable for our customers.

isacikgoz (approved):

PR #6555, opened 2023-07-28
Fix URL to ad-ldap config

vish9812:


cwarnermm (approved):
Thank you, @vish9812!

Issue #6499, opened 2023-06-26
Add to docs the requirement to use SSL for the copy link feature when using non https or localhost desktop app


amyblais:
The [ticket](https://mattermost.atlassian.net/browse/MM-45024) was closed since there is no solution or workaround for this specially for the desktop app. Following are the technical reasons. Pasting below comments from the developers:

"It is true that navigator.clipboard is not available for non-secure sites or for localhost, the only and widely used workaround is to use document.execCommand, (which is a deprecated api by several browsers due to the potential security risks it has), and it is already used in our code base as a fallback for the original navigator.clipboard unavailability.
The fact is that Electron uses Chromium as the running app browser and Chromium has already deprecated the document.execCommand, so clipboard is not available because the lack of certificate and document.execCommand is deprecated meaning there is no way to make it work.

TL;DR: for non https or localhost desktop app, the copy link button won’t work and there is nothing technically possible to do about this. Only option is that the customer uses a ssl certificate, even a self-signed certificate would work."

Is there anywhere in the documentation that we could add the requirement of SSL for the copy link feature?

cwarnermm:
Addressed via https://github.com/mattermost/docs/pull/6563

PR #6563, opened 2023-07-28
SSL needed for copy link feature for non-secure deployments

cwarnermm:
Documentation for: https://github.com/mattermost/docs/issues/6499


cwarnermm:
@pvev @marianunez - Is there better guidance we could offer our admins to address the issue raised regarding shared links beyond or in addition to what's been proposed in this docs PR?

source/onboard/ssl-client-certificate.rst
```diff
@@ -11,7 +11,8 @@ Before you begin, follow the `official guides to install Mattermost </guides/adm
 For the purposes of this guide, the Mattermost server domain name is ``example.mattermost.com``, and the user account is ``mmuser`` with email ``mmuser@mattermost.com`` and password ``mmuser-password``.
 
 .. note::
-  Generating the client certificates in this section is optional if you have already generated them before.
+  - Generating the client certificates in this section is optional if you have already generated them before.
+  - An SSL certificate (or a self-signed certificate) is required in cases where your Mattermost Server **Site URL** is configured as ``HTTP`` or ``localhost``.
```
pvev, line 15: thanks @cwarnermm for enhancing the documentation around this issue 👍 . The `share-links.rst` seems like a good place to mention the potential problems with the copy link function, but I wonder if is there another place in the documentation that guides on how to invite members since that is also a very important place where the "copy link" function won't work without ssl certificate and can be annoying for users.
For the `ssl-client-certificate.rst`, maybe we should add like a strong suggestion on why it is really important to use it in a production environment (Security and Privacy, Compliance and Browser limitations) and mentioning that some product functionalities could be degraded because the lack of it, like in this case, the copy to clipboard functionality.

cwarnermm: @pvev - I've incorporated your feedback. Thank you again! Are there any other areas of product functionality that would be impacted that we'll want to call out via docs? Or would sharing links via messages and inviting people via URL be the only two?

source/onboard/ssl-client-certificate.rst
```diff
@@ -10,8 +10,9 @@ Before you begin, follow the `official guides to install Mattermost </guides/adm
 
 For the purposes of this guide, the Mattermost server domain name is ``example.mattermost.com``, and the user account is ``mmuser`` with email ``mmuser@mattermost.com`` and password ``mmuser-password``.
 
-.. note::
-  Generating the client certificates in this section is optional if you have already generated them before.
+.. important::
+  - Generating the client certificates in this section is optional if you have already generated them before.
+  - If your production Mattermost Server instance **Site URL** is ``HTTP`` or ``localhost``, we strongly recommend configuring an SSL certificate (or a self-signed certificate) for security, privacy, compliance, as well as to avoid browser limitations that can prevent Mattermost product features from working that copy data using the user's local clipboard, including `sharing links in messages </channels/share-links.html>`__ and `inviting new members to teams </welcome/about-teams.html#send-a-team-invite-link>`__.
```
pvev, line 15: I would just remove this part from this sentence:
`If your production Mattermost Server instance **Site URL** is ``HTTP`` or ``localhost``,`

pvev (approved):
LGTM @cwarnermm . Thanks! 🚢

PR #6579, opened 2023-08-09
Made MS Teams Sync plugin version more generic

cwarnermm:
Reduced maintenance impact by removing version specifics.


jespino (approved):
LGTM

PR #6588, opened 2023-08-17
Clarify use of read replicas

grubbins:
The docs suggest that reads are evenly distributed between the master DB and the read replica DBs. Based on both the server code and the observed behaviour, I do not believe that is the case - I believe that (if any read replicas are available) read queries are distributed among the read replicas only. This is a useful distinction to keep in mind for site admins.


agnivade (commented):
The original text is technically correct. While it's true that 99% of times, read requests are sent to the read replicas, there are cases where they aren't. If you want to change the text to reflect this, we can do it. Otherwise we can leave it as is.

grubbins:
Yes - I think there's a big difference between "Mattermost distributes read requests across all databases", and "Mattermost distributes *almost all* read requests across the read replicas only". Would indeed like to see an update...

Are you asking me to propose different wording?

cwarnermm:
@grubbins - We'd welcome your docs contribution with updated, clarified wording if you're open to it.

grubbins:
OK - please see what you think of my latest wording.

agnivade (approved):

source/scale/high-availability-cluster.rst
```diff
@@ -214,11 +214,14 @@ Migrating to NAS or S3 from local storage is beyond the scope of this document.
 Database configuration
 ^^^^^^^^^^^^^^^^^^^^^^
 
-Use the read replica feature to scale the database. The Mattermost server can be set up to use one master database and multiple read replica databases. Mattermost distributes read requests across all databases, and sends write requests to the master database, and those changes are then sent to update the read replicas.
+Use the read replica feature to scale the database. The Mattermost server can be set up to use one master database and onr or more read replica databases.
```
cwarnermm, line 217:
```suggestion
Use the read replica feature to scale the database. The Mattermost server can be set up to use one master database and one or more read replica databases.
```

cwarnermm (approved):
Thank you so much, @grubbins!

grubbins:
Oops - I think the markdown requires a linebreak before the bullet points. The html currently up on the docs page has them as inline asterisks.

Could you insert the linebreak @cwarnermm ? Or do you need me to open a new PR with it?

cwarnermm:
Fixed via https://github.com/mattermost/docs/pull/6627

PR #6604, opened 2023-08-21
[CLD-6153] Migration to GitHub Actions 

phoinixgrr:
#### Summary

Migrating to GitHub Actions. Review by commit is recommended. 

- [[chore] Rename build workflow to ci workflow](https://github.com/mattermost/docs/commit/ff25adf34ff8afda77caaee9106bc64f4b80f582)
- [[chore] Removing unused files since moving to ghactions](https://github.com/mattermost/docs/commit/5bc7d0c38b2b4becd292aaf6e72c04d30ae138f2)
- [[feat] Migrating to GitHub Actions](https://github.com/mattermost/docs/commit/b8ab819ed460f029caf7426d6800f84f7c92c084)

  Introducing cd.yml and ci_master.yml worfklows, to tackle safe deployment to AWS S3 service.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+        make python-deps
+        make html
+    - name: ci/Persist docs artifacts
+      uses: actions/upload-artifact@65d862660abb392b8c4a3d1195a2108db131dd05 # v3.1.2
+      with:
+        name: docs
+        path: build/html
+    - name: ci/Persist doc logs artifacts
+      uses: actions/upload-artifact@65d862660abb392b8c4a3d1195a2108db131dd05 # v3.1.2
+      with:
```
toninis, line 26: Add retention here for the CI builds . Do we actually need those uploaded ? Aren't they visible on the console ?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,29 @@
+name: CD
```
toninis, line 1: ✅ This is great . Should we though migrate it to our internal repo ?

toninis (approved):
🚀 Great one . Just some minor comments about logs and location of the workflow

.github/workflows/ci.yml
phoinixgrr: There is retention. It is the default(90d), and this was the way that this pipeline used to run from ever. 
Sure we can change this. Kindly define a proposed retention . 


-- 
The way `workflow_run:` works, is that we completely seperate the build process and the deploy process, to be able to safely run with inheriting secrets to the CD pipeline. 

In order to do that, we need a way to connect the calling pipeline (CI_MASTER) with the actual pipeline that deploys the build to Amazon S3. This is performed via upload-artifacts.  This is how we safely share data between these pipelines. 

The privileged pipeline containing AWS Secrets, does not build anything. for security reasons.

.github/workflows/cd.yml
phoinixgrr: @toninis  Sure . I'l raise a seperate ticket to tackle this.

.github/workflows/ci_master.yml
```diff
@@ -1,39 +1,30 @@
-name: Build
-
+name: CI Master
```
mvitale1989, line 1: This file seems to be identical to `ci.yml`, save for the workflow trigger.

Would it make sense to remove this duplication and move the trigger over to `ci.yml`, rather than having a dedicated file for master branch's CI?

mvitale1989 (commented):
Awesome work, thanks Akis! Left a small comment on whether it's possible to deduplicate the `ci*.yml` files

.github/workflows/ci_master.yml
phoinixgrr: We can do that. 
But, How will be able to distinguish running the CD process, only on master branch ? 

The `workflow_run:` triggers on workflow complete.

mvitale1989: There seems to be a way to control which branches' workflows the `workflow_run` trigger should react to, e.g. by adding the `branches: [master]` key/value to the trigger: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#limiting-your-workflow-to-run-based-on-branches

phoinixgrr: @mvitale1989  Pushed https://github.com/mattermost/docs/pull/6604/commits/61ca6e84af197fc5015ab2deb02904bc3a365f2d to address this.

.github/workflows/ci.yml
phoinixgrr: @toninis  Any objections merging this as is?

mvitale1989 (approved):

PR #6608, opened 2023-08-24
Clarify Show Email Address option

esarafianou:
#### Summary
Clarifies the Show Email address option highlighting all the System Roles that have access to the email address even if the option is set to false. 

#### Ticket Link
* https://mattermost.atlassian.net/browse/MM-52959 (internal)


cwarnermm (approved):
Thanks, @esarafianou!

PR #6618, opened 2023-08-25
Added setcap FAQ for single server deploys

cwarnermm:
@nab-77 - Does this also apply to deploys on Ubuntu?

@emdecr - I'd like to apply the blue-text formatting to in-line code on the RHEL deploy page to match the tarball page. Can you modify this docs PR to demonstrate how this is achieved please? I'll then apply it to other deploy pages as needed. Thanks!


source/install/install-rhel-8.rst
```diff
@@ -98,4 +98,37 @@ If you wish to remove the Mattermost Server for any reason, you must stop the Ma
...
 	Depending on your configuration, there are several important folders in ``/opt/mattermost`` to backup. These are ``config``, ``logs``, ``plugins``, ``client/plugins``, and ``data``. We strongly recommend you back up these locations before running the ``rm`` command.
 
-You may also remove the Mattermost systemd unit file and the user/group created for running the application.
\ No newline at end of file
+You may also remove the Mattermost systemd unit file and the user/group created for running the application.
+
+Frequently asked questions
+--------------------------
+
+Can I run Mattermost without a proxy?
```
nab-77, line 106: I would place this after the FAQs about starting at boot and any mentions of nginx.

source/install/install-tar.rst
```diff
@@ -88,6 +88,21 @@ You may also remove the Mattermost systemd unit file and the user/group created
 Frequently asked questions
 --------------------------
 
+Can I run Mattermost without a proxy?
```
nab-77, line 91: I would place this after the FAQs about starting at boot and any mentions of nginx.

source/install/install-rhel-8.rst
```diff
@@ -98,4 +98,37 @@ If you wish to remove the Mattermost Server for any reason, you must stop the Ma
...
+Yes. However, like NGINX, in place, Mattermost binds to 443 instead of 8065. The Mattermost binary requires the correct permissions to do that binding. You must activate the ``CAP_NET_BIND_SERVICE`` capability to allow the new Mattermost binary to bind to ports lower than 1024 by running the following command:
+
+.. code-block:: none
+  :class: mm-code-block 
+
+    sudo setcap cap_net_bind_service=+ep ./mattermost/bin/mattermost
+
+.. note::
+  
+  - We don't recommend deploying Mattermost on a single server unless you have fewer than 200 users.
```
nab-77, line 118: Right now we can't really make this claim and since MM Omnibus is not limited to 200 users I would remove this item.

Maybe a note that having a proxy in front on Mattermost is highly recommended and if setup/maintenance is a worry then use MM Omnibus.

source/install/install-tar.rst
```diff
@@ -88,6 +88,21 @@ You may also remove the Mattermost systemd unit file and the user/group created
...
+Yes. However, like NGINX, in place, Mattermost binds to 443 instead of 8065. The Mattermost binary requires the correct permissions to do that binding. You must activate the ``CAP_NET_BIND_SERVICE`` capability to allow the new Mattermost binary to bind to ports lower than 1024 by running the following command:
+
+.. code-block:: none
+  :class: mm-code-block 
+
+    sudo setcap cap_net_bind_service=+ep ./mattermost/bin/mattermost
+
+.. note::
+  
+  - We don't recommend deploying Mattermost on a single server unless you have fewer than 200 users.
```
nab-77, line 103: Right now we can't really make this claim and since MM Omnibus is not limited to 200 users I would remove this item.

Maybe a note that having a proxy in front on Mattermost is highly recommended and if setup/maintenance is a worry then use MM Omnibus.

nab-77 (changes requested):
@cwarnermm this FAQ item is relevant for all OSs we document (but we should sanity check the cmd is correct).

source/_static/mytheme.css
```diff
@@ -2759,7 +2759,9 @@ mark {
 #install-mattermost-server-using-the-tarball p code,
 #install-mattermost-server-using-the-tarball h3 code,
 #install-mattermost-server-on-ubuntu p code,
-#install-mattermost-server-on-ubuntu h3 code {
+#install-mattermost-server-on-ubuntu h3 code,
+#install-mattermost-on-rhel p code,
+#install-mattermost-on-rhel h3 code {
```
emdecr, line 2764: @cwarnermm Re: inline code - every page has a unique ID on the content container. 
Screen Shot 2023-09-01 at 10 35 18 AM

What you'll want to do is add that id to the list of selectors we have for the inline code on install pages. Specifically, for paragraph and h3 elements on the page (as that's where we're using them thus far).
```css
#unique-id-here p code,
#unique-id-here h3 code 
```

We can probably scale this via some sort of logic esp if we keep adding this styling to more pages, but we can do that in another pass.

Lemme know, if anything - thanks!

cwarnermm: Thank you so much, @emdecr!

source/install/install-ubuntu.rst
coltoneshaw, line 1: @cwarnermm - Why do we differ our wording from tar / ubuntu ? Realistically, we've always said if using Mattermost as the proxy we don't suggest going over 200 users. That statement remains true for any OS.

I'd alter the wording slightly and match it on all three to say:

> When using Mattermost as the proxy we do not support over 200 concurrent users. If you've exceeded 200 concurrent users, you will need a proxy, like nginx,  infront of Mattermost to better handle the traffic.

I don't think it's a technical limitation, though, more of an architectural one that starts to expose cracks at scale.

coltoneshaw:
> @coltoneshaw - Are you or a member of your team open to spot-checking the setcap command documented via the tarball & Ubuntu deployment pages, please? (they're copied over from the RedHat deploy page and may need platform-specific tweaks).

The actual commands appear to be correct and from what I can gather, it's OS agnostic for it.

source/install/install-rhel-8.rst
```diff
@@ -102,4 +102,37 @@ If you wish to remove the Mattermost Server for any reason, you must stop the Ma
...
+Yes. Mattermost binds to 443 instead of 8065. The Mattermost binary requires the correct permissions to do that binding. You must activate the ``CAP_NET_BIND_SERVICE`` capability to allow the new Mattermost binary to bind to ports lower than 1024 by running the following command:
+
+.. code-block:: none
+  :class: mm-code-block 
+
+    sudo setcap cap_net_bind_service=+ep ./mattermost/bin/mattermost
+
+.. note::
+  
+  - We recommend the `Mattermost Omnibus install method </install/installing-mattermost-omnibus.html>`__ over the tarball if you are running the Mattermost Server and database a single system as this greatly reduces setup and ongoing maintenance.
```
coltoneshaw, line 137: @cwarnermm - I feel like because this is RHEL docs, we should remove this line as it's not relevant. @nab-77

source/install/install-tar.rst
```diff
@@ -107,6 +107,17 @@ Why does Mattermost fail to start at system boot?
...
+  - We recommend the `Mattermost Omnibus install method </install/installing-mattermost-omnibus.html>`__ over the tarball if you are running the Mattermost Server and database a single system as this greatly reduces setup and ongoing maintenance.
```
coltoneshaw, line 123: We should add a `**Ubuntu Only` to this because it's generic tar docs. I wouldn't want to mislead here.

Side note, the omnibus page has no info on what OSs it supports

cwarnermm: @nab-77 - The Omnibus docs are missing the prerequisites section currently. Are you open to advising on these details?

source/install/install-ubuntu.rst
cwarnermm: Good callout, @coltoneshaw! @nab-77 - Thoughts?

cwarnermm:
@nab-77 - Looking forward to your input on this docs PR.

source/install/install-rhel-8.rst
nab-77: @cwarnermm yeah. The guidance is about single server but if the admin is on this RHEL page its because they use RHEL :-)

source/install/install-tar.rst
nab-77: @cwarnermm I am not following. What is missing on the Omnibus install page?

nab-77 (approved):
Format is good. Agree to the Omnibus alternative note box.

PR #6633, opened 2023-08-30
Update software-hardware-requirements.rst

svelle:
This PR makes the section "MySQL 8 Support" directly linkable. 
Right now the way the page is formatted the header for that section looks like all the other headings but isn't linkable, however in support it can be very useful to be able to link directly to that section as the collation issues come up here and there.


cwarnermm (approved):
Thanks, @svelle!

PR #6637, opened 2023-09-06
Remote Marketplace disabled

cwarnermm:
Documentation for: https://github.com/mattermost/mattermost/pull/24311


esethna:
Adding @calebroseland to review. I think this is still accurate in v9.0 but we may be reenabling the remote marketplace in v9.0.1 to allow for remote updates to supported plugins

esethna (approved):

PR #6639, opened 2023-09-07
Update keyboard-shortcuts-for-channels.rst

johndavidlugtu:
Update `Ctrl + K` to mention the functionality of making a text into a hyperlink

@cwarnermm thoughts on this? Currently, `Ctrl + K` only works as "Find Channels if no link is created".

That said, there's also a thought to entertain: Should `Ctrl + K` be used as a keyboard shortcut to "Find Channels"?

### DIFF
--- a/source/channels/keyboard-shortcuts-for-channels.rst
+++ b/source/channels/keyboard-shortcuts-for-channels.rst
@@ -31,7 +31,7 @@ The following keyboard shortcuts are supported in all `supported browsers </inst
 +-------------------------------------+------------------------------+----------------------------------------------------------------------------------+
 | :kbd:`Alt` + select channel         | :kbd:`⌥` + select channel    | Mark the last post in the channel as unread.                                     |
 +-------------------------------------+------------------------------+----------------------------------------------------------------------------------+
-| :kbd:`Ctrl` :kbd:`K`                | :kbd:`⌘` :kbd:`K`            | Open the **Find Channels** dialog.                                               |
+| :kbd:`Ctrl` :kbd:`K`                | :kbd:`⌘` :kbd:`K`            | Open the **Find Channels** dialog. If text is highlighted, a hyperlink is created|
 +-------------------------------------+------------------------------+----------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


cwarnermm:
@johndavidlugtu - One keyboard shortcut that performs two actions is likely something we want to avoid. CC @matthewbirtch

matthewbirtch:
My understanding is the following:

- if text is selected in the message input field, CTRL-K makes a hyperlink
- in any other case, CTRL-K opens the find channels modal

This is the jira ticket for this change: https://mattermost.atlassian.net/browse/MM-48061

johndavidlugtu:
> One keyboard shortcut that performs two actions is likely something we want to avoid

Correct -- but in this case, the keyboard shortcut depends on the context (whether there's highlighted text or not). @cwarnermm 

@matthewbirtch that's right

cwarnermm:
From a UX perspective, is a context-aware shortcut shared among 2 actions something we want to move forward with, @matthewbirtch? Happy to support via the docs.

matthewbirtch:
> From a UX perspective, is a context-aware shortcut shared among 2 actions something we want to move forward with, @matthewbirtch? Happy to support via the docs.

It _is_ already in the product. And, this is actually something familiar to the developer crowd since github does this very same thing.

Command-K is a standard keyboard shortcut for creating links across all major text editors so we wanted to offer that familiarity while still supporting the find channels shortcut in all other cases - this was the solution we came up with. cc @esethna.

cwarnermm:
Thanks all! @johndavidlugtu - The code base has moved past this docs PR. It'll be easier to close this PR and create a new docs PR for this update. I've added this to the v9.1 release updates effort. Closing.

johndavidlugtu:
Just thought to drop something related to this conversation:

https://support.atlassian.com/jira-software-cloud/docs/what-is-the-command-palette/

Jira decided to also rollout a context-aware `Command + K` shortcut

cwarnermm:
Closed in favour of https://github.com/mattermost/docs/pull/6696

PR #6650, opened 2023-09-19
fix: Add missing slash for proper redirection on database preparation

toninis:
#### Summary

Add missing slash to fix redirection to database preparation


cwarnermm (approved):
Thank you, @toninis!!

PR #6664, opened 2023-09-27
Initial redo of config settings for postgres

coltoneshaw:
#### Summary
Attempting to pull some of the config settings for Postgres that exist in a gist right now - https://gist.github.com/coltoneshaw/48c67b0b2af0018849211596b2c46e36


source/scale/high-availability-cluster.rst
```diff
@@ -289,40 +289,69 @@ The database can be configured for high availability and transparent failover us
...
+  max_worker_processes = 12
+  max_parallel_workers_per_gather = 4
+  max_parallel_workers = 12
+  max_parallel_maintenance_workers = 4
 
-11. **maintenance_work_mem**: 1GB (reduce this to 512MB if your server has less than 32GB of RAM)
+**Config for Postgres Replica node**
 
-12. **autovacuum_vacuum_cost_limit**: 500
+Copy all the above settings to the read replica, and modify or add only the below.
```
coltoneshaw, line 338: @cwarnermm - I don't love this specifically, I can replica the config down below as well but that also felt noisy

coltoneshaw:
@cwarnermm / @nab-77 - I have docs on how to set up replication for postgres on self hosted also here - https://gist.github.com/coltoneshaw/48c67b0b2af0018849211596b2c46e36

Wondering if you'd like me to bring these into the docs as well. I know customers have opinions on how to do replication, but at least _how_ to do it in a way that we find works best might be useful. let me know.

source/scale/high-availability-cluster.rst
```diff
@@ -289,40 +289,69 @@ The database can be configured for high availability and transparent failover us
...
-1. **max_connections**: If you are using read-replicas, set reader connections to 1024, and set writer connections to 256. If you are using a single instance, then only setting reader connections to 1024 should be sufficient. If the instance is lower capacity than r5.xlarge, then set it to a lower number. Also tune the ``MaxOpenConns`` setting under the ``SqlSettings`` of the Mattermost app accordingly.
+**Config for Postgres Primary or Writer node**
 
-2. **random_page_cost**: Set it to 1.1, unless the DB is using spinning disks.
+.. code-block:: bash
 
-3. **work_mem**: Set it to 16MB for readers, and 32MB for writers. If it's a single instance, 16MB should be sufficient. If the instance is of a lower capacity than r5.xlarge, then set it to a lower number.
+  # If the instance is lower capacity than r5.xlarge, then set it to a lower number. 
+  # Also tune the "MaxOpenConns" setting under the "SqlSettings" of the Mattermost app accordingly. 
+  # Note that "MaxOpenConns" on Mattermost is per connection string.
```
agnivade, line 300: "per connection" may be confusing. Perhaps better to say "per database"?

source/scale/high-availability-cluster.rst
```diff
@@ -289,40 +289,69 @@ The database can be configured for high availability and transparent failover us
...
-12. **autovacuum_vacuum_cost_limit**: 500
+Copy all the above settings to the read replica, and modify or add only the below. 
 
-.. note::
+.. code-block:: bash
   
-   If you are using pgbouncer, or any similar connection pooling proxy, in front of your DB, then apply the keepalive settings to the proxy instead, and revert the keepalive settings for the DB back to defaults.
+  # If the instance is lower capacity than r5.xlarge, then set it to a lower number. 
+  # Also tune the "MaxOpenConns" setting under the "SqlSettings" of the Mattermost app accordingly. 
+  # Note that "MaxOpenConns" on Mattermost is per connection string.
```
agnivade, line 344: same as above

agnivade (commented):
Just a small note. Looks good otherwise.

source/scale/high-availability-cluster.rst
coltoneshaw: What i'm trying to say is "per connection string" but you're right it might be unclear. The setting it's truly per database if the customer is using RDS and chooses their endpoint, which is one string for N number of databases. 

@cwarnermm - any suggestions on making it more clear? I'm probably over thinking it in the name of clarity

TLDR; Taking the below config for example. There are 4 databases configured (`primary`, `replica-1`, `replica-2`, and `search-replica`). That means MaxOpenConns can do 300 * 4

```sql
      "SqlSettings": {
        "DataSource": "primary",
        "DataSourceReplicas": [
          "replica-1",
          "replica-2"
        ],
        "DataSourceSearchReplicas": [
          "search-replica"
        ],
        "MaxOpenConns": 300,
    },
```

But if we modify it to be the below, you have three strings with `replicas` representing `replica-1` and `replica-2`, it's just been abstracted away from Mattermost. So the max connections Mattermost will open here is 300 * 3. 

```sql
      "SqlSettings": {
        "DataSource": "primary",
        "DataSourceReplicas": [
          "replicas"
        ],
        "DataSourceSearchReplicas": [
          "search-replica"
        ],
        "MaxOpenConns": 300,
    },
```

agnivade: Maybe "per datasource endpoint"?

cwarnermm:
> @cwarnermm / @nab-77 - I have docs on how to set up replication for postgres on self hosted also here - https://gist.github.com/coltoneshaw/48c67b0b2af0018849211596b2c46e36
> 
> Wondering if you'd like me to bring these into the docs as well. I know customers have opinions on how to do replication, but at least _how_ to do it in a way that we find works best might be useful. let me know.

Our recommended way of doing replications is excellent guidance to include in the product documentation.

source/scale/high-availability-cluster.rst
cwarnermm: "Per database endpoint" is precise and doesn't assume the number of databases in play. Out of curiosity, why is "per connection string" unclear?

source/scale/high-availability-cluster.rst
cwarnermm: Agreed, @coltoneshaw. Any time you find yourself duplicating content, it's worth pausing to consider whether that duplication is truly in the best interest of the user consuming the information (since it'll add a maintenance burden to ensure the duplication remains synchronized).

I like the approach you're taking. Is there an opportunity to split the two tasks up for better clarity? A task that copies, followed by a task that clarifies what to modify/add to?

source/scale/high-availability-cluster.rst
agnivade: Per connection is confusing. Per connection string is technically fine, but the word generally used is dsn or datasource name or database endpoint. 0/5 on what to use.

cwarnermm (approved):

agnivade (approved):

PR #6665, opened 2023-09-27
Clarified OS required for Omnibus

coltoneshaw:
#### Summary
This page gave the incorrect impression that any linux OS could run omnibus. Clarified that


cwarnermm (approved):
Thank you, @coltoneshaw!

PR #6671, opened 2023-09-29
Replica lag settings and guide

coltoneshaw:
#### Summary
Updated the replica lag setting to be more explicit on how to configure this properly. Also, added more details to the examples.


source/configure/database-configuration-settings.rst
```diff
@@ -388,27 +388,111 @@ Replica lag settings
...
+3. Navigate to your Grafana instance monitoring Mattermost and open the `Mattermost Performance Monitoring v2 <https://grafana.com/grafana/dashboards/15582-mattermost-performance-monitoring-v2/>`__ dashboard.
+4. The ``QueryTimeLag`` chart is already setup for you utilizing the existing ``Replica Lag`` chart. If using ``QueryAbsoluteLag`` metric clone the ``Replica Lag`` chart and edit the query to use the below absolute lag metrics and modify the title to be ``Replica Lag Absolute``.
+
+    .. code-block:: none
+
+      mattermost_db_replica_lag_abs{instance=~"$server"}
+
+    .. image::  ../../source/images/database-configuration-settings-replica-lag-grafana-1.jpg
+      :alt: A screenshot showing how to clone a chart within Grafana
+
```
coltoneshaw, line 490: I don't know how you'd suggest adding spaces here between the images. The images may not be needed, but I felt it added some clarity.

source/configure/database-configuration-settings.rst
```diff
@@ -376,7 +376,7 @@ Replica lag settings
 | String array input consists of:                        |                                                                                  |
 |                                                        |                                                                                  |
 | - ``DataSource``: The database credentials to connect  |                                                                                  |
-|   to the replica instance.                             |                                                                                  |
+|   to the database instance.                            |                                                                                  |
```
coltoneshaw, line 379: This was very intentionally changed. As I dug into each query we provided, all the guidance said that they can be run on any node for AWS Aurora and MySQL Group replication or the primary for Postgres replication slots

coltoneshaw:
The only large missing item would be the replication lag settings for Postgres using WAL Archive streaming. I need to do some digging into that one

source/configure/database-configuration-settings.rst
coltoneshaw: Aurora - https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora_global_db_instance_status.html

> You can run this function from any instance in an Aurora PostgreSQL DB cluster or an Aurora PostgreSQL global database. The function returns details about lag for all replica instances.

MySQL Group Replication - https://dev.mysql.com/doc/refman/8.0/en/group-replication-replication-group-member-stats.html

> The information is shared between all the server instances that are members of the replication group, so information on all the group members can be queried from any member.

streamer45 (approved):
Great improvements! /cc @agnivade as the subject expert, in case you'd like to give it a quick pass.

agnivade:
Looks good. 

> The only large missing item would be the replication lag settings for Postgres using WAL Archive streaming. I need to do some digging into that one

The replication slots method is the preferred method AFAIU right? If so, I'd rather nudge customers to use that rather than us documenting for every method.

coltoneshaw:
@agnivade - Thanks for the review!

> The replication slots method is the preferred method AFAIU right? If so, I'd rather nudge customers to use that rather than us documenting for every method.

It's not necessarily preferred, it's just the one natively built into Postgres, so it's easier to manage / setup in my opinion. It's also what I've built my replication docs upon that we share with customers.

We have customers using both slots and wal streaming, so I think it would be best to have both since there is only two here. The MySQL ones will eventually be removed.

source/configure/database-configuration-settings.rst
```diff
@@ -388,27 +388,111 @@ Replica lag settings
...
-|                                                                                                                                           |
-| For MySQL Group Replication, the absolute lag can be measured from the number of pending transactions in the applier queue:               |
-|                                                                                                                                           |
-| .. code-block:: sql                                                                                                                       |
-|                                                                                                                                           |
-|   select member_id, count_transactions_remote_in_applier_queue FROM performance_schema.replication_group_member_stats where member_id=<>  |
+| - The ``QueryAbsoluteLag`` and ``QueryTimeLag`` queries must return a single row.                                                         |
+| - To properly monitor this you must setup `performance monitoring </scale/performance-monitoring.html>`__ for Mattermost.    |
```
cwarnermm, line 394: FYI: There's a pipe misaligned on this table that will prevent the table from displaying.

source/configure/database-configuration-settings.rst
```diff
@@ -388,27 +388,111 @@ Replica lag settings
...
+    .. tab:: AWS Aurora
+
+      Add the configuration highlighted below to your ``SqlSettings.ReplicaLagSettings`` array. You only need to add this once because replication statistics for AWS Aurora nodes are visible across all server instances that are members of the cluster. Be sure to change the ``DataSource`` to point to a single node in the group. 
+
+      For more information on Aurora replication stats, see the `AWS Aurora documentaion <https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora_global_db_instance_status.html>`__.
+
+          **Example:**
+
+          .. code-block:: json
+            :emphasize-lines: 4,5,6,7,8
```
cwarnermm, line 410: TIL! I hadn't seen this option before! Thank you!

cwarnermm (approved):

cwarnermm:
@coltoneshaw - Added a couple of non-blocking comments, and noted one formatting issue that will prevent a table from displaying as expected. Great PR! Thank you! Is there any outstanding work remaining on this PR from your perspective?

source/configure/database-configuration-settings.rst
coltoneshaw: Right?! seemed pretty cool and easy to add.

coltoneshaw:
@cwarnermm - I just fixed the pipe issue. I think it's good to merge in it's current state and as we uncover more ways to track replica lag we can add them here.

PR #6673, opened 2023-10-01
Fix whois site URL in faq-illicit-use.rst

varghesejose2020:
Fix whois site URL in faq-illicit-use.rst


cwarnermm (approved):
Thank you, @varghesejose2020!

Please note: Files within the ``/archive`` directory shouldn't be modified since that directory contains outdated, legacy content we no longer build into the docs site. Thanks!

PR #6685, opened 2023-10-11
[MM-54496] Document authentication options to Calls services

streamer45:
#### Summary

Porting a couple of important notes on authentication options for Calls related services since some customers were struggling to find them as they were a little buried in the GitHub repositories.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54496


source/configure/plugins-configuration-settings.rst
```diff
@@ -432,14 +432,19 @@ RTCD service URL
...
++---------------------------------------------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------+
+| **Note**:                                                                                                                                                                                                                                                                 |
+|                                                                                                                                                                                                                                                                           |
+| - The client will self-register the first time it connects to the service and store the authentication key in the database. If no client ID is explicitly provided, the diagnostic ID of the Mattermost installation will be used.                                        |
```
cpoile, line 445: We should maybe add a bit about what to do when they get an error (client id already registered or not found) -- I've gotten those from time to time when restarting.

cpoile (approved):
One minor suggestion, thanks!

source/configure/plugins-configuration-settings.rst
streamer45: I am thinking that would fit more into a troubleshooting section though.

cwarnermm (approved):

source/configure/plugins-configuration-settings.rst
streamer45: Created https://mattermost.atlassian.net/browse/MM-54926 as follow up.

PR #6758, opened 2023-11-02
Fix config name for "Enable diagnostics and error reporting"

hmhealey:
#### Summary
Just fixing a typo I noticed


source/configure/logging-configuration-settings.rst
```diff
@@ -229,12 +229,12 @@ Enable diagnostics and error reporting
...
 
-+-----------------------------------------------+---------------------------------------------------------------------------+
-| Configure Mattermost to allow any combination | - System Config path: **Environment > Logging**                           |
-| of console, local file, syslog, and TCP       | - ``config.json setting``: ``".LogSettings.AdvancedLoggingJSON": "",``    |
-| socket targets, and send log records to       | - Environment variable: ``MM_LOGSETTINGS_ADVANCEDLOGGINGJSON``            |
-| multiple targets.                             |                                                                           |
-+-----------------------------------------------+---------------------------------------------------------------------------+
++-----------------------------------------------+-------------------------------------------------------------------------+
+| Configure Mattermost to allow any combination | - System Config path: **Environment > Logging**                         |
+| of console, local file, syslog, and TCP       | - ``config.json setting``: ``".LogSettings.EnableDiagnostics": "",``    |
```
cwarnermm, line 234: Great catch, @hmhealey! Can you update the search metadata directly above the heading for this config setting too, please, with the same updates? Thanks!

hmhealey: Will do. I also just noticed that the description of this setting is also totally wrong as well

cwarnermm: You're right! Wow, that's some bad copypasta! Really appreciate your help with these updates!

hmhealey: That should be fixed now. I wasn't too fond of saying it was sent to "Mattermost, Inc", but I copied that from the System Console description. I couldn't think of a better way to clarify that it's being sent to us and not just being stored in the MM server

cwarnermm (approved):
Thanks so much, @hmhealey!

PR #6763, opened 2023-11-06
Clarified SSO availability

cwarnermm:


victorehernandez (commented):
@cwarnermm This looks good in that the options _not available_ in the Free Edition are no longer cited.  I wonder, though - as per https://docs.mattermost.com/onboard/sso-gitlab.html, GitLab SSO with the OAuth 2.0 standard is available in the Free Edition.  I'm pondering whether we should note that in this particular part of the documentation.  I'd hate to add complexity, but just want to make sure our docs are solid and consistent.  Thoughts?

cwarnermm:
Thanks, @victorehernandez - You make an excellent point about the need to clarify our GitLab SSO support down to the standard, since support differs between free and paid editions. My latest commit applies your feedback.

victorehernandez (approved):
@cwarnermm This looks good - consistent with the availability across editions and with the documentation in https://docs.mattermost.com/onboard/sso-gitlab.html.

Thank you!

PR #6790, opened 2023-11-15
Clarify export create command details

cwarnermm:


hanzei:
@mgdelacroix Do we also need to update the code in https://github.com/mattermost/mattermost/blob/afb48219c0d925f12db19e3ccc81830de2a54702/server/cmd/mmctl/commands/export.go#L102-L103? What is the Single Source of Through for the `mmctl` docs?

cwarnermm:
You raise an excellent point, @hanzei. The mmctl repo on GitHub is the single source of truth that informs the mmctl product documentation. However, the mmctl GitHub repo isn't expected to be a carbon copy of the docs content, but rather a starting point or a bare minimum of detail. The product docs provide additional context where necessary. Or in this case, clarify functionality as needed.

@mgdelacroix and I have chatted in the past about consolidating these two sources of information into one to avoid situations where the two information sources differ.

I'm 0/5 on the value of updating the mmctl repo with this update. Thoughts?

hanzei:
Thanks for sharing that extra context @cwarnermm :+1: 

Given that this PR modifies the documentation of the individual parameters, we should update the help text of the command as well.

On a higher level, I wonder much the difference between the docs here and the docs in `mmctl` really is.

cwarnermm:
Agreed that the in-product help text for the mmctl command should also be updated. I'll create a PR for that update.

Your question on the delta between the two repos is a good one, and one that @mgdelacroix and I have discussed at length in the past. The docs repo includes more context and detail than the mmctl repo overall. Knowing precisely what the differences are would go a long way towards understand what's possible in terms of consolidation. 

We'd save considerable manual effort keeping these 2 sources sync'd by working from a single source of truth.

mgdelacroix:
Agree on updating the `mmctl` repo for this change. Generally, I think the rule of thumb from my PoV is to make both sources similar in arguments, flags and descriptions, and then think if the examples and long descriptions make sense in the context of a user exploring the `--help` of the command while running it. Long explanations, links to other sections and things like that may not be that useful, but explaining a caveat, setting expectations on particular behaviors and (of course) referencing the extended docs are.

In any case, 100% agree @cwarnermm, is an ongoing discussion and a non-solved problem :sweat_smile: As per merging both sources of truth, it's a shame that we didn't include the `docs` repo in the mono-repo when we merged, that would have helped to having both synced, although is a bigger discussion for sure

PR #6867, opened 2024-01-12
Fix version support message

ShrootBuck:
Wording could probably be improved


cwarnermm:
@nab-77 - Do we have an upper limit on the Ubuntu release version Mattermost supports?

ShrootBuck:
> @nab-77 - Do we have an upper limit on the Ubuntu release version Mattermost supports?

It's for the Omnibus package. I tried on 23.10 and there was a message saying that 22.04 LTS was the limit.

nab-77 (approved):
@ShrootBuck thanks for raising this.

We do indeed have version checking in the Omnibus installer. We do this as part of our testing to ensure with LTS being our preferred version (for obvious reasons). Will discuss with team on regarding speed of tracking Ubuntu release.

In the meantime are you able to deploy using 22.04.3 LTS (Aug 2023 release)?

PR #6875, opened 2024-01-16
Fix plan header depths

bbodenmiller:
#### Summary

Move plans under main plan header


cwarnermm (approved):
Thanks, @bbodenmiller!

PR #6885, opened 2024-01-25
Update performance-monitoring.rst

hanzei:
#### Summary
Clarify that `/debug/pprof/profile` needs to be called with a `seconds` query parameter.

#### Ticket Link
None


agarciamontoro (approved):
I'm not opposed to the change, better explicit than implicit, but it's not mandatory, [it defaults to 30 seconds](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/net/http/pprof/pprof.go;l=135-138).

PR #6923, opened 2024-02-16
Update configuation-in-a-database.rst

jlandells:
SELECT statement to validate the migration was incorrect—the Active column is boolean

#### Summary

Towards the bottom of this page, there's a SQL SELECT statement that is used to validate that the config has been successfully migrated to the database.  The WHERE clause was checking that `Active=1`, but the Active column is of type boolean, so this generated a SQL error.  Corrected the example to `true`.


cwarnermm (approved):
Thanks, @jlandells!!

PR #6947, opened 2024-02-29
remove nginx proxy_cache

doc-sheet:
#### Summary
Current proxy_cache settings are kinda harmful and dangerous.

When everything is ok cache is bypassed, but during downtime nginx configured to serve cached responses and there is a huge problem.

Default `proxy_cache_key $scheme$proxy_host$request_uri;` does not separate users from each other.
Users might get cached pages with another user's session.

There are 2 ways of fix
1. add some variable to proxy_cache_key (like `$cookie_MMAUTHTOKEN`)
2. remove cache completely

I prefer 2nd one.


maxwellpower (approved):
Pending any specific reason we recommend using caching, I approve removing this recommendation from our default config. I am not entirely sure it is beneficial. Perhaps for static assets on larger installs.

I would also agree with OP that any risk of presenting another users session from cache during an outage is not a preferred outcome.

fmartingr (approved):
I agree with the change if we can't find any particular reason to have the cache enabled. It would only make sense for the static files, avatars and attachments, but surely not for all requests.

cwarnermm:
@nab-77 - Looping in Product Management on this docs PR for final call. How do you recommend we proceed?

nab-77 (approved):
I can indeed see how a cache from Nginx can be problematic (certainly in the cases that MM is offline).

source/install/config-proxy-nginx.rst
```diff
@@ -10,7 +10,7 @@ Configure NGINX as a proxy for Mattermost server
 NGINX is configured using a file in the ``/etc/nginx/sites-available`` directory. You need to create the file and then enable it. When creating the file, you need the IP address of your Mattermost server and the fully qualified domain name (FQDN) of your Mattermost website.
 
 1. Log in to the server that hosts NGINX and open a terminal window.
-2. Create a configuration file for Mattermost by running 
+2. Create a configuration file for Mattermost by running
```
cwarnermm, line 13:
```suggestion
2. Create a configuration file for Mattermost by running the following command:
```

source/install/config-ssl-http2-nginx.rst
```diff
@@ -16,29 +16,27 @@ Although you can configure HTTP/2 without SSL, both Firefox and Chrome browsers
...
 
   .. note::
-   
+
    - On Ubuntu this file is located at ``/etc/nginx/sites-available/``. If you don't have this file, run ``sudo touch /etc/nginx/sites-available/mattermost``.
    - On CentOS/RHEL this file is located at ``/etc/nginx/conf.d/``. If you don't have this file, run ``sudo touch /etc/nginx/conf.d/mattermost``.
-   - Note that the IP address included in the examples in this documentation may not match your network configuration. 
-   - If you're running NGINX on the same machine as Mattermost, and NGINX resolves ``localhost`` to more than one IP address (IPv4 or IPv6), we recommend using ``127.0.0.1`` instead of ``localhost``. 
-   
+   - Note that the IP address included in the examples in this documentation may not match your network configuration.
```
cwarnermm, line 30:
```suggestion
   - The IP address included in the examples in this documentation may not match your network configuration.
```

cwarnermm:
@doc-sheet - Thank you for your documentation PR. Very much appreciated!

PR #6991, opened 2024-03-19
Fix nav scripting, classes and styles

emdecr:
#### Summary
This PR addresses the nav bug where an `active` class is added to the nav upon page load. I've changed the classes used to activate the nav. My guess is that the `active` class is used for some other functionality on the site/page and was being applied even if the user had not interacted with the nav. 

Seen on: https://docs.mattermost.com/deploy/mobile-app-changelog.html


cwarnermm (approved):

cwarnermm:
Confirmed that the mobile changelog page issue is resolved with this docs PR.

PR #6994, opened 2024-03-21
fix: add missing "mmctl" keyword in a command example

vish9812:


cwarnermm (approved):
Great catch, @vish9812! Thank you!

PR #7063, opened 2024-04-12
Update SqlSettings.MaxIdleConns default setting

hanzei:
#### Summary
Per https://github.com/mattermost/mattermost/blob/98712737e600e8b11e7d72ea90faa266227aa855/server/public/model/config.go#L1236-L1238.

It seems like the default setting was always at 20, and the docs were always incorrect.

#### Ticket Link
None

### DIFF
--- a/source/configure/database-configuration-settings.rst
+++ b/source/configure/database-configuration-settings.rst
@@ -153,7 +153,7 @@ Data source
   :systemconsole: Environment > Database
   :configjson: SqlSettings.MaxIdleConns
   :environment: MM_SQLSETTINGS_MAXIDLECONNS
-  :description: The maximum number of idle connections held open to the database. Default is **10**.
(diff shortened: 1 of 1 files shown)


agnivade (approved):
Long term, this would be a pain to maintain. Because it is completely manual and there's no sync with the code. I'd rather just have a link the the default config.json generated for any config setting.

cwarnermm:
@hanzei - Please update the default value in the table to match the search metadata. Thanks!

cwarnermm:
@agnivade "Because it is completely manual and there's no sync with the code. I'd rather just have a link the the default config.json generated for any config setting." -- I agree. It would be ideal to find a way to generate config docs directly from code rather than maintain the docs manually. Lowers the potential for errors, and reduces manual effort each release.

hanzei:
@cwarnermm Good catch; I also updated the table. PTAL.

cwarnermm (approved):

PR #7096, opened 2024-04-23
Correctly state that Off-Topic can get covered to a private channel

hanzei:
#### Ticket Link
None


svelle (approved):
LGTM

source/collaborate/convert-public-channels.rst
```diff
@@ -11,7 +11,7 @@ Convert public channels to private channels
 You must be a system admin or team admin to convert public channels to private channels. When a channel is converted from public to private, its history and membership are preserved. Membership in a private channel remains as invitation only. Publicly-shared files remain accessible to anyone with the link.
 
 .. note::
-    Default channels such as ``Town Square`` and ``Off-Topic`` can't be converted to private channels.
+    Default channels such as ``Town Square`` can't be converted to private channels.
```
cwarnermm, line 14: @hanzei - If both Town Square and Off-Topic are default channels, and default channels can't be converted to private channels, why is Off-Topic being removed here? There's additional docs impact to this change, so I'm looking to clarify my product understanding.

svelle: @cwarnermm both get created by default but Town-Square can not be deleted as it is the fallback channel where any user will be redirected to in case they navigate to a missing channel or leave one. Off-Topic can be deleted and doesn't have the same use as Town-Square.

cwarnermm:
Thanks, @svelle, really appreciate the explanation! I'll review the docs for additional impact and commit any necessary changes to this PR before approving/merging. Really great catch, @hanzei.

source/collaborate/convert-public-channels.rst
hanzei, line 14: Maybe this is even clearer; what do you think, @cwarnermm?
```suggestion
    The default channel ``Town Square`` can't be converted to private channels.
```

cwarnermm: TIL! :)

svelle (approved):

cwarnermm (approved):

PR #7101, opened 2024-04-24
For formatting of Advanced logging example

hanzei:
#### Summary
**Before**
Screenshot from 2024-04-24 10-51-32

**After**
Screenshot from 2024-04-24 11-00-35

#### Ticket Link
None


source/manage/logging.rst
```diff
@@ -185,7 +185,7 @@ Define advanced log output
                     "MaxBackups": 10,
                     "MaxSizeMB": 500
```
hanzei, line 186: @wiggin77 Is that correct? The docs in http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/7101/manage/logging.html#file-target-configuration-options state that it's `max_size`.

source/manage/logging.rst
```diff
@@ -185,7 +185,7 @@ Define advanced log output
...
                 },
-                    MaxQueueSize": 1000
+                "MaxQueueSize": 1000
```
hanzei, line 188: Fix JSON format

source/manage/logging.rst
wiggin77: The linked docs are correct:
```go
type FileOptions struct {
	// Filename is the file to write logs to.  Backup log files will be retained
	// in the same directory.  It uses <processname>-lumberjack.log in
	// os.TempDir() if empty.
	Filename string `json:"filename"`

	// MaxSize is the maximum size in megabytes of the log file before it gets
	// rotated. It defaults to 100 megabytes.
	MaxSize int `json:"max_size"`

	// MaxAge is the maximum number of days to retain old log files based on the
	// timestamp encoded in their filename.  Note that a day is defined as 24
	// hours and may not exactly correspond to calendar days due to daylight
	// savings, leap seconds, etc. The default is not to remove old log files
	// based on age.
	MaxAge int `json:"max_age"`

	// MaxBackups is the maximum number of old log files to retain.  The default
	// is to retain all old log files (though MaxAge may still cause them to get
	// deleted.)
	MaxBackups int `json:"max_backups"`

	// Compress determines if the rotated log files should be compressed
	// using gzip. The default is not to perform compression.
	Compress bool `json:"compress"`
}
```

wiggin77:
@cwarnermm I wish the docs were a bit more clear that the [Advanced logging section](http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/7101/manage/logging.html#advanced-logging) applies to the general [logging section](http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/7101/manage/logging.html#mattermost-logging)  and the [audit logging section](http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/7101/manage/logging.html#audit-logging-beta).  The only difference is the log levels.

wiggin77 (approved):

cwarnermm:
Agreed, @wiggin77. I'll work a clarification update into the v9.8 docs.

source/manage/logging.rst
hanzei: But the examples are still wrong, right? They are using `MaxSizeMB`, but should be using either `max_backups` or `MaxBackups`.

hanzei: Opened https://github.com/mattermost/docs/pull/7109

PR #7108, opened 2024-05-06
Clarify NGINX timeout values

hanzei:
#### Summary
Per https://nginx.org/en/docs/syntax.html
> A value without a suffix means seconds. It is recommended to always specify a suffix.

This PR clarifies NGINX timeout values as seconds.

#### Ticket Link
https://hub.mattermost.com/private-core/pl/rha7ce68xby3zyoeea34xmh9rr


mrckndt (approved):
Looks good to me

PR #7136, opened 2024-05-16
Update scale guidance

agnivade:
Some of the app level and nginx level config optimizations were missing/outdated. Revamping them as part of this PR.


cwarnermm (approved):
Thank you, @agnivade! Please feel free to merge this great update when reviews are complete.

source/scale/high-availability-cluster.rst
```diff
@@ -155,40 +191,89 @@ A sample configuration for NGINX is provided below. It assumes that you have two
...
+            proxy_cache_min_uses 2;
+            proxy_cache_use_stale timeout;
+            proxy_cache_lock on;
+            proxy_ignore_headers Cache-Control Expires;
+            proxy_cache_valid 200 24h;
+      }
+
+      location / {
+            proxy_set_header Connection "";
+            client_max_body_size 50M;
```
streamer45, line 255: Can we retain the change we made on https://github.com/mattermost/docs/pull/7021? I also realize that's a different document, not sure if we need both or if we should converge.

source/scale/high-availability-cluster.rst
```diff
@@ -155,40 +191,89 @@ A sample configuration for NGINX is provided below. It assumes that you have two
...
+            proxy_http_version 1.1;
+            proxy_pass http://backend;
+            proxy_cache mattermost_cache;
+            proxy_cache_revalidate on;
+            proxy_cache_min_uses 2;
+            proxy_cache_use_stale timeout;
+            proxy_cache_lock on;
+            proxy_ignore_headers Cache-Control Expires;
+            proxy_cache_valid 200 24h;
+      }
```
streamer45, line 251: If I remember correctly we made this change because our load-test didn't implement proper client-side caching of static assets which a real browser would do. Do we need them here?

agnivade: This might still be helpful for serving common content across multiple clients. But overall, not too much opinionated on this.

streamer45: What worries me is that we are completely ignoring both `Cache-Control` and `Expires` headers so we are essentially not giving the client any way to refresh stale data nor control caching as it would normally expect. I think there's a chance this could be causing some issues.

streamer45: I think we can remove the entire block. Requests will be handled by the root one which looks to be exactly the same once we remove `proxy_ignore_headers` and `proxy_cache_valid`.

nab-77 (commented):
any changes here that we need to consider for docs.mm.com @agnivade?

agnivade:
Sorry, didn't get you @nab-77. These changes _are_ for docs.mattermost.com.

agnivade:
Ah good point there. Perhaps we can link from there to this page, mentioning that "to reach high scales, consider the tunings in linked page".

streamer45 (approved):

agnivade:
@agarciamontoro - No rush, whenever you get some time.

agarciamontoro (approved):
This is gold! Thanks, Agniva :)

PR #7170, opened 2024-06-07
Update plugins-configuration-settings.rst

coltoneshaw:
#### Summary
When setting `defaultenabled` to false it sets `Test Mode` to `true` in the UI.
Screenshot 2024-06-07 at 3 03 10 PM


cwarnermm:
@streamer45 - Is this intentional product functionality?

streamer45:
@cwarnermm Yes, `defaultenabled` was effectively relabeled as "test mode" but it's the same setting underneath as we didn't want to cause a breaking change.

cwarnermm (approved):
Thanks, @coltoneshaw!

streamer45 (commented):
@coltoneshaw I don't think these changes are needed. The semantic is inverted due to the labeling, but the behavior is as documented. Test mode enabled (on) translates into a false value of defaultenabled.

streamer45:
> @streamer45 - as it stands now the docs are incorrect in guidance for setting this. Setting to true does not mean only sys admins can start calls like the docs say.

Okay, that makes sense, but the setting is called "Test Mode" here, which clearly refers to the admin UI so it's confusing. I wish we deprecated it more formally now. I am okay with merging the changes here but maybe we should change the name back to what it really is and add a note instead.

coltoneshaw:
@streamer45 i'd second more docs clarification. It took a few looks to make sure I submitted this right 😂 on being false tripped me along with naming.

PR #7192, opened 2024-06-11
Added note detailing what happens with deactivated user integrations

cwarnermm:


source/configure/user-management-configuration-settings.rst
```diff
@@ -105,11 +105,24 @@ To delete a user from your Mattermost deployment, you can deactivate the user's
...
 
+What happens to deactivated user integrations?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. important::
+
+  If you deactivate a Mattermost user who has integrations tied to their user account, consider the following consequences and recommendations based on the integration type:
+
+  - **Slash commands** will continue to work after user deactivation. You may want to consider deleting the existing slash command token and creating a new slash command associated with a different user account to decouple sensitive token data from the deactivated user account.
+  - **Outgoing webhooks** will continue to work after user deactivation; however, access to webhook tokens is limited to team admins and system admins only, so the risk of exposing sensitive token data is lower. You may want to consider deleting the existing outgoing webhook token and creating a new outgoing webhook associated with a different user account.
```
esarafianou, line 121: When the outgoing webhook is created, the to-be-deactivated user who creates the outgoing webhook gets to see the webhook token both as the response to the webhook creation and when they visit the the outgoing integrations. 

So this phrase is not accurate. The creator of the webhook can always see the webhook token:
>  however, access to webhook tokens is limited to team admins and system admins only, so the risk of exposing sensitive token data is lower.

We can suggest that the regenerate the webhook token and also that they check that the deactivated user no longer has access to the callback URLs, as having access would result in the deactivating user receiving the outgoing webhooks.

Screenshot 2024-06-12 at 12 11 18

source/configure/user-management-configuration-settings.rst
```diff
@@ -105,11 +105,24 @@ To delete a user from your Mattermost deployment, you can deactivate the user's
...
   :alt: Activate a user in Mattermost using the System Console.
 
+What happens to deactivated user integrations?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. important::
+
+  If you deactivate a Mattermost user who has integrations tied to their user account, consider the following consequences and recommendations based on the integration type:
+
+  - **Slash commands** will continue to work after user deactivation. You may want to consider deleting the existing slash command token and creating a new slash command associated with a different user account to decouple sensitive token data from the deactivated user account.
```
esarafianou, line 120: We can also suggest that they regenerate the token and that they check that they check that the deactivated user doesn't have access to the slash command `Request URL` as this is the callback URL to receive the HTTP POST or GET event request when the slash command is run.

source/configure/user-management-configuration-settings.rst
```diff
@@ -105,11 +105,24 @@ To delete a user from your Mattermost deployment, you can deactivate the user's
...
+  - **Outgoing webhooks** will continue to work after user deactivation; however, access to webhook tokens is limited to team admins and system admins only, so the risk of exposing sensitive token data is lower. You may want to consider deleting the existing outgoing webhook token and creating a new outgoing webhook associated with a different user account.
+  - **Incoming webhooks** will continue to work after user deactivation. Sensitive token data risks are low risk because incoming webhooks don't include a token returned to a user.
```
esarafianou, line 122: As we mention here: https://developers.mattermost.com/integrate/webhooks/incoming/#create-an-incoming-webhook in #3, the URL produced is of the format https://your-mattermost-server.com/hooks/**xxx-generatedkey-xxx**.  Anyone who has it will be able to post messages to the Mattermost instance.Unfortunately we don't offer a way to regenerate the generated key, so they'd need to remove the incoming webhook and recreate a new one.

esarafianou (changes requested):
@cwarnermm Added a few comments. Let me know what you think

source/configure/user-management-configuration-settings.rst
```diff
@@ -105,11 +105,24 @@ To delete a user from your Mattermost deployment, you can deactivate the user's
...
   :alt: Activate a user in Mattermost using the System Console.
 
+What happens to deactivated user integrations?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. important::
+
+  If you deactivate a Mattermost user who has integrations tied to their user account, consider the following consequences and recommendations based on the integration type:
+
+  - **Slash commands** will continue to work after user deactivation. Consider deleting the existing slash command token and creating a new slash command associated with a different user account to decouple sensitive token data from the deactivated user account. In addition, consider regenerating the token and check that the deactivated user doesn't have access to the slash command **Request URL** which is the callback URL to receive the HTTP POST or GET event request when the slash command is run.
```
esarafianou, line 120:
```suggestion
  - **Slash commands** will continue to work after user deactivation. Consider deleting the existing slash command and creating a new slash command associated with a different user account to decouple sensitive token data from the deactivated user account. Alternatively, consider regenerating the token of the existing slash command. Check that the deactivated user doesn't have access to the slash command **Request URL** which is the callback URL to receive the HTTP POST or GET event request when the slash command is run.
```

esarafianou (approved):
Minor suggestions for clarity.

source/configure/user-management-configuration-settings.rst
esarafianou: They cannot delete only the token, they need to either regenerate it or delete the slash command entirely and recreate it.

Issue #7230, opened 2024-06-23
Generic Linux install Desktop Client instructions don't work


mxa:
The 'generic Linux' [install instructions](https://docs.mattermost.com/collaborate/install-desktop-app.html) for the Mattermost Desktop Client don't work.

_A [sic] AppImage distribution (Beta) of a compressed tarball is available. Automatic app updates are supported and enabled. When a new version of the desktop app is released, your app updates automatically._

What is it? An AppImage or a binary in a compressed tarball? Confusing wording. I expected an AppImage, but get a tarball, not an AppImage. I think the sentence should read: "An AppImage distribution AND a compressed tabball of a binary are available." Both are downloadeble here: https://github.com/mattermost/desktop/releases/tag/v5.8.1 but in the docs the AppImage gets and awkward mention convoluted with the tarball and then only the tarball is linked as a download.

Following the instructions:

_2. Extract the archive to a convenient location, then execute mattermost-desktop located inside the extracted directory._

```
$ ./mattermost-desktop 
[19564:0623/180528.261713:FATAL:setuid_sandbox_host.cc(158)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /home/mxa/Applications/mattermost-desktop-5.8.1-linux-x64/chrome-sandbox is owned by root and has mode 4755.
Trace/breakpoint trap (core dumped)
```

mvitale1989:
Thanks for reporting! We have included those resolution steps in the [Desktop app's tarball installation instructions](https://docs.mattermost.com/collaborate/install-desktop-app.html), as on Ubuntu Noble it seems the only way to give `chrome-sandbox` the right ownership and permissions after extracting the tarball as a non-root user.

mxa:
Btw, I appreciate the speedy support and resolution here. That is a good user experience.

Issue #7229, opened 2024-06-23
Ubuntu/Debian Desktop Client install instructions do not work.


mxa:
Following the [installation guide](https://docs.mattermost.com/collaborate/install-desktop-app.html) for Ubuntu Debian on Kubuntu 24.04, I can't install Mattermost. The relevant terminal output here:

```
$ curl -fsS -o- https://deb.packages.mattermost.com/setup-repo.sh | sudo bash
Err:21 https://deb.packages.mattermost.com stable InRelease                                                                       
  The following signatures were invalid: EXPKEYSIG F8F2C31744774B28 Mattermost Build <dev-ops@mattermost.com>
$ sudo apt install mattermost-desktop
Package mattermost-desktop is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'mattermost-desktop' has no installation candidate
```

mxa:
Solution is to remove the old offending key like this:
`curl -fsS -o- https://deb.packages.mattermost.com/setup-repo.sh | sudo UPDATE_GPG_KEY=yes bash`
via: https://forum.mattermost.com/t/mattermost-desktop-repository-install-on-24-04-lts-noble/18693/9

PR #7050, opened 2024-04-08
Added note for Omnibus & Ubuntu admins to update GPG key

cwarnermm:
Documentation for: https://github.com/mattermost/desktop/issues/3000

PR #7231, opened 2024-06-24
Desktop: add note about rotated key, chrome-sandbox instructions

mvitale1989:
#### Summary

This PR tackles two issues of the Desktop app installation instructions:
- https://github.com/mattermost/docs/issues/7229: add a note on what to do if you have the expired GPG repo key on your system. We had updated the docs for server and omnibus for as part of [this PR](https://github.com/mattermost/docs/pull/7050), but didn't update them for Desktop.
- https://github.com/mattermost/docs/issues/7230: add instructions required to have the mattermost-desktop tarball run on Ubuntu Noble.

Regarding the second issue: this seems to be a widespread issue with electron-builder apps, e.g. see [here](https://youtrack.jetbrains.com/issue/IJPL-59368/IDE-crashes-due-to-chrome-sandbox-is-owned-by-root-and-has-mode-error-when-IDE-is-launching-the-JCEF-in-a-sandbox) or [here](https://github.com/NixOS/nixpkgs/issues/308128). The root cause is the following:
- `chrome-sandbox` needs to be owned by `root` and to have the SETUID bit enabled. If not, AppArmor will deny some operations on Ubuntu Noble (but not on previous Ubuntu releases).
- When electron-builder creates the archive, it gives `root` ownership to all the files in the archive (can be verified with `tar -tv`)
- If the electron-builder tar extraction is done by a non-root user, all of the file will belong to that user instead, including `chrome-sandbox`. This also drops the setuid bit in the process.

The above points mean that with our current installation instructions don't work on Ubuntu Noble, and we must explicitly address the `chrome-sandbox` file permissions when using the tarball.

Note that an alternative solution to what I documented is requesting to extract the whole tarball as root user, e.g. to `/opt/mattermost-desktop/`. cc @devinbinnie please let me know if you have any feedback or more insight on this, thanks!

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-7945
https://mattermost.atlassian.net/browse/CLD-7949

### DIFF
--- a/source/collaborate/install-desktop-app.rst
+++ b/source/collaborate/install-desktop-app.rst
@@ -87,7 +87,7 @@ You can download the `desktop app <https://mattermost.com/apps/>`_ directly from
 
   .. important::
 
-    The GPG public key has changed. If you had previously setup the repository on your system, you'll need to download the new key: you can set the ``UPDATE_GPG_KEY=yes`` environment variable when running the setup script, to configure it to overwrite the previous key on your system with the new one. E.g. the first step of the instructions would become ``curl -fsS -o- https://deb.packages.mattermost.com/setup-repo.sh | sudo UPDATE_GPG_KEY=yes bash``. Note that depending on your setup, additional steps may also be required, particularly for installations that didn't rely on the repository setup script.
+    The GPG public key has changed. If you had previously set up the repository on your system, you'll need to download the new key. You can set the ``UPDATE_GPG_KEY=yes`` environment variable when running the setup script to configure it to overwrite the previous key on your system with the new one. The first step of installation then becomes: ``curl -fsS -o- https://deb.packages.mattermost.com/setup-repo.sh | sudo UPDATE_GPG_KEY=yes bash``. Depending on your setup, additional steps may also be required, particularly for installations that don't rely on the repository setup script.
 
   1. At the command line, set up the Mattermost repository on your system: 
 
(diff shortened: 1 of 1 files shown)


source/collaborate/install-desktop-app.rst
```diff
@@ -85,6 +85,10 @@ You can download the `desktop app <https://mattermost.com/apps/>`_ directly from
 
   Both a ``.deb`` package (Beta), and an official APT repository is available for Debian 9 and for Ubuntu releases 20.04 LTS or later. Automatic app updates are supported and enabled. When a new version of the desktop app is released, your app updates automatically.
 
+  .. important::
+
+    The GPG public key has changed. If you had previously setup the repository on your system, you'll need to download the new key: you can set the ``UPDATE_GPG_KEY=yes`` environment variable when running the setup script, to configure it to overwrite the previous key on your system with the new one. E.g. the first step of the instructions would become ``curl -fsS -o- https://deb.packages.mattermost.com/setup-repo.sh | sudo UPDATE_GPG_KEY=yes bash``. Note that depending on your setup, additional steps may also be required, particularly for installations that didn't rely on the repository setup script.
```
cwarnermm, line 90:
```suggestion
    The GPG public key has changed. If you had previously set up the repository on your system, you'll need to download the new key. You can set the ``UPDATE_GPG_KEY=yes`` environment variable when running the setup script to configure it to overwrite the previous key on your system with the new one. The first step of installation then becomes: ``curl -fsS -o- https://deb.packages.mattermost.com/setup-repo.sh | sudo UPDATE_GPG_KEY=yes bash``. Depending on your setup, additional steps may also be required, particularly for installations that don't rely on the repository setup script.
```

cwarnermm (approved):
Thank you, @mvitale1989! I've proposed editorial updates inline. Approving for expediency.

phoinixgrr (approved):

source/collaborate/install-desktop-app.rst
mvitale1989: Thanks for the improvements Carrie :tada:

mvitale1989:
I'll merge this as is, if there's more feedback on this later we can revisit the instructions.

PR #7234, opened 2024-06-25
FY25 Packaging Updates

cwarnermm:
Packaging changes to align with Mattermost Pricing page


source/onboard/shared-channels.rst
```diff
@@ -1,40 +1,38 @@
-Shared channels (experimental)
-==============================
+Shared channels (Beta)
```
esethna, line 1: Shared channels is still going to stay "Experimental" until we release the UI being worked on right now

source/onboard/shared-channels.rst
```diff
@@ -159,10 +157,10 @@ Status details include:
 Frequently Asked Questions
 ---------------------------
 
-Why is this feature experimental?
+Why is this feature in beta?
```
esethna, line 160: experimental

esethna:
> do we want to note that video is being considered for a future Calls release?

@cwarnermm, I'd recommend we leave that out here.

PR #7235, opened 2024-06-27
MM-58742: Add a note for postgres upgrade

agnivade:
#### Summary

Add note for postgres upgrade

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58742

### DIFF
--- a/source/install/prepare-mattermost-database.rst
+++ b/source/install/prepare-mattermost-database.rst
@@ -207,3 +207,7 @@ To set up a PostgreSQL database for use by the Mattermost server:
 The PostgreSQL interactive terminal starts. To exit the PostgreSQL interactive terminal, type ``\q`` and press :kbd:`Enter` on Windows or Linux, or :kbd:`↵` on Mac.
 
 When the PostgreSQL database is installed, and the initial setup is complete, you can install the Mattermost server.
+
+### Important note while upgrading a Postgres instance
+
+If you are upgrading a major version of Postgres, it is essential that ``ANALYZE VERBOSE`` is run on the database post upgrade. This is necessary to re-populate the ``pg_statistics`` table used to generate optimal query plans. The database performance might suffer if this step is not done.


source/install/prepare-mattermost-database.rst
```diff
@@ -207,3 +207,7 @@ To set up a PostgreSQL database for use by the Mattermost server:
 The PostgreSQL interactive terminal starts. To exit the PostgreSQL interactive terminal, type ``\q`` and press :kbd:`Enter` on Windows or Linux, or :kbd:`↵` on Mac.
 
 When the PostgreSQL database is installed, and the initial setup is complete, you can install the Mattermost server.
+
+### Important note while upgrading a Postgres instance
+
+If you are upgrading a major version of Postgres, it is essential that ``ANALYZE VERBOSE`` is run on the database post upgrade. This is necessary to re-populate the ``pg_statistics`` table used to generate optimal query plans. The database performance might suffer if this step is not done.
```
streamer45, line 213: Is there a way to run it on the whole DB or should we perhaps recommend specific tables?

Should we note that this process may have an impact on performance?

agnivade: No we should run it on the whole DB. The current command does that. This does not do any vacuuming, so there are no write locks acquired. `ANALYZE` only acquires a read lock: https://www.postgresql.org/docs/current/sql-analyze.html#id-1.9.3.46.8. So I think it should be fine to skip it.

stafot (approved):
LGTM

source/install/prepare-mattermost-database.rst
streamer45: Understood, so this is sufficient. 

Side question, why don't we do this in our load test then but require vacuuming instead?

agnivade: Vacuuming is technically not needed actually. We just do this to keep things neat and tidy.

streamer45 (approved):

cwarnermm (approved):

PR #7264, opened 2024-07-11
SMTP improvements: formatting fixes & key callouts

cwarnermm:


source/configure/smtp-email.rst
```diff
@@ -6,126 +6,95 @@ SMTP email setup
...
-                select ``Send a Test Email`` to verify setup is working
-                correctly.
+    4. Choose a sender address like ``mattermost@domain.com`` and select ``Send a Test Email`` to verify setup is working correctly.
 	    
-        - By local ``sendmail`` can be used by using **Server Name** ``127.0.0.1`` with **Port** ``25``
-	- Or by using one of the services listed below in the **Sample SMTP Settings**
+    Alternatively, you can use one of the `services listed below <#sample-smtp-settings>`__, or can set up local ``sendmail`` by setting **Server Name** ``127.0.0.1`` with **Port** ``25``.
 
-2. **Configure SMTP settings**
+    If deploying Mattermost using :doc:`Docker </install/install-docker>`, the standard docker ``172.16.0.0/12`` IP range isn't used. Specify the IP range ``192.168.0.0/24`` to the email service to avoid relay access errors.
```
streamer45, line 28: It's not super clear that this refers to the postfix configuration below which ideally should only include the expected network where Mattermost is running on rather than multiple wildcards. 

In the case of our docker setup, `192.168.0.0/24` should work fine, but it could vary depending on how Mattermost is deployed.

cwarnermm: Thanks, @streamer45! So the postfix and the Docker notes are related, correct? I myself wasn't clear on that based on the issue logged.

streamer45: Correct @cwarnermm , the security exception on the network class is about allowing Mattermost to communicate with postfix so it relates to the latter's config.

PR #7290, opened 2024-07-18
Update sso-saml.rst to include missing de-provisioning warning

svelle:
Right now the fact that SAML does not support automatic de-provisioning is not very prominent in the SAML docs.

This PR adds a warning to make it more clear that this is the case and what can be done about it (use LDAP sync)


cwarnermm (approved):
Thank you, @svelle!

PR #7315, opened 2024-08-07
Improved code block rendering

lieut-data:
#### Summary
I encountered an issue rendering Yaml blocks (see https://community.mattermost.com/core/pl/wdnw6s81bprc8p4dh786irwnfy) and took at look as to why. This led me down a bunny trail of all sorts of code-block rendering inconsistencies that I'm hoping to have cleared up here:
* use `code-block` instead of triple backticks to get the free copy/paste behaviour
* remove the old `mm-code-copy` and associated styling in favour of the now universal copy button that appears for code blocks
* use the right language for the various instances of `::code-block`
* split up some cases where the command to run and the output expected were in the same code block, making it hard to copy/paste

There's more I would have liked to fixup -- various pages using lots of `run command` that don't make it easy to copy/paste -- but kept it smallish for now.

I've clicked "Next" through most of the pages in the docs to make sure things work as expected. Here's a handful of screenshots to illustrate the changes:

### Before
Specialized copy buttons in just a few places:

CleanShot 2024-08-09 at 12 41 17@2x

### After
Consolidated on the already existing copy code block paradigm used almost everywhere else

CleanShot 2024-08-09 at 12 41 20@2x

### Before
Combined code block made it hard to copy just the commands to run:

CleanShot 2024-08-09 at 12 48 04@2x

### After
Two code blocks to copy just the commands to run:

CleanShot 2024-08-09 at 12 48 08@2x

### Before
Missing YAML indentation

CleanShot 2024-08-09 at 12 53 00@2x

### After
Readable YAML indentation

CleanShot 2024-08-09 at 12 43 09@2x

#### Ticket Link
None


lieut-data:
For review purposes, https://github.com/mattermost/docs/pull/7315/commits/5b816e888c92fa805e22bbe164e2e30aa909f967 is perhaps the one commit with more judgement calls than the rest, many of which were global search and replaces or just removing the legacy copy/paste code.

cwarnermm:
Thank you so much, @lieut-data! Really appreciate your help cleaning up code blocks across the docs site. It's been a to-do item for quite some time.

One thing worth noting: the specialized copy buttons are intentional. They were introduced by Marketing to track engagement with specific installation docs content. CC @emdecr for visibility.

emdecr:
Thanks @cwarnermm - I'm meeting with AT today and can double check the status of the tracking on these buttons. Will circle back.

lieut-data:
> One thing worth noting: the specialized copy buttons are intentional. They were introduced by Marketing to track engagement with specific installation docs content. CC @emdecr for visibility.

Aha, thanks for that context @cwarnermm! @emdecr, if still in use, I'm curious if we're trying to track "engagement" with types of buttons, or just interest in a particular piece of text in a long list of things to be copied? Would be ideal to converge on decoupling style and tracking, if possible -- let me know how I can help.

lieut-data:
Thanks, @emdecr! 

> We don't use the conversion on Marketing's end, but I'm not sure if anybody in Product is still keeping tabs on this either. I honestly haven't heard anybody mention it for several months/longer.

I'll follow up with product to double check how/if we used this data, but I bet we'd get better metrics from just access logs on `https://deb.packages.mattermost.com/` directly.

cwarnermm:
@lieut-data - @nab-77 will be a great person to loop into this conversation.

nab-77:
I believe we are tracking all actions from the "Copy to clipboard" clicks. We did track them when they were introduced.

emdecr (approved):
As per [the discussion in MM](https://hub.mattermost.com/private-core/pl/wdpmk6puh3gkxbow69r63uiz1c) sounds like we're good to make these changes. Did a look over the updates and ran it locally: looks gravy. Thanks @lieut-data and @cwarnermm 🙏

cwarnermm (approved):
Thank you, @lieut-data!! This PR can be merged at your convenience.

PR #7321, opened 2024-08-12
Added metadata details for generated support packets

cwarnermm:
Documentation for: https://github.com/mattermost/mattermost/pull/27573

### DIFF
--- a/source/manage/common-support-packet.rst
+++ b/source/manage/common-support-packet.rst
@@ -59,7 +59,7 @@ Add the generated support packet to a Mattermost Support ticket, or share with w
 
 .. important::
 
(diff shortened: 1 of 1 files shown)


source/manage/common-support-packet.rst
```diff
@@ -5,15 +5,14 @@ Use the System Console or the :ref:`mmctl system supportpacket <manage/mmctl-com
 Contents of a support packet
 ----------------------------
 
-A Mattermost Support Packet can contain the following files:
+A Mattermost support packet can contain the following files:
```
hanzei, line 8: I think it makes sense to use capital letters for the term "Support Packet" as it's a special term used only in Mattermost. It also matches the casing in the UI and other documents.

hanzei (changes requested):
Thanks for taking care of the docs, Carrie!

PR #7332, opened 2024-08-14
Corrected feature flag & corrected Sys Console UI setting location/path

cwarnermm:
Based on internal feedback by @coltoneshaw


source/scale/deploy-prometheus-grafana-for-performance-monitoring.rst
```diff
@@ -113,7 +113,7 @@ To help you get started, you can download three sample dashboards shared in Graf
...
-- `Mattermost Notification Health metrics <https://grafana.com/grafana/dashboards/21305-mattermost-notification-health>`_, which can be used to track different types of notifications sent from Mattermost. Accessing Mattermost Notification Health requires the feature flag ``NotificationMetrics`` to be set to ``true``. System admins can :ref:`disable notification metrics data collection <configure/environment-configuration-settings:enable notification metrics>` through the System Console.
+- `Mattermost Notification Health Monitoring <https://grafana.com/grafana/dashboards/21305-mattermost-notification-health>`_, which can be used to track different types of notifications sent from Mattermost. Accessing and enabling Mattermost Notification Health Monitoring requires the feature flag ``NotificationMonitoring`` to be set to ``true``. System admins can :ref:`disable notification monitoring data collection <configure/site-configuration-settings:enable notification metrics>` through the System Console.
```
coltoneshaw, line 116: does it require the feature flag? or just the setting? I honestly don't know

cwarnermm: Well worth confirming with @marianunez and the team. @marianunez - can you confirm whether the notifications monitoring is still behind a feature flag, please?

marianunez: We still have a feature flag so the documentation is correct. Having said that, given that the feature is stable and also has a setting to disable, @devinbinnie we should be ready to remove the flag

devinbinnie: Yeah I'd be okay with removing the feature flag.

PR #7356, opened 2024-08-20
MM-53705: Update metrics docs

agarciamontoro:
#### Summary
I tried to automate generating a list with all metrics registered, but that ended up being quite cumbersome, so this is the result of a semi-automated generation plus some manual post-processing.

My idea was to use [the Gather function in the registry](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Gatherer) and write a script that simply called that. However, it's not as easy:
- The `go_sql_*` metrics need to be registered explicitly. This is doable, it's just those ones. 
- The `*Vec` metrics need to be observed at least once so that those metrics are reported back by `Gather`. This is error-prone, there are multiple ones, and doing this manually it's terrible

You can see the test I wrote in [this commit](https://github.com/mattermost/mattermost/commit/6238fce4e3835974e99c00d7f11e0ef6dc80586a). That prints the metrics in the same order we had in the docs, so that it was easy to compare if my test was missing some registered metrics (i.e., if there were some *Vec metrics whose Observe/Increment/Decrement functions I had not called). That happened quite a lot, so it was a very manual process, and we'd need to maintain that list of calls to the Observe/Increment/Decrement functions. Not good.

If we want to really automate this, I see two alternatives:
1. Use custom functions wrapping the `prometheus.New*` and `MustRegister`, so that we can collect the names and descriptions in a list we can later export.
2. Use `reflect` to inspect the calls to `MustRegister` and collect the names and descriptions of the passed arguments.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53705

### DIFF
--- a/source/scale/performance-monitoring-metrics.rst
+++ b/source/scale/performance-monitoring-metrics.rst
@@ -33,13 +33,16 @@ The above metrics can be used to calculate ETag and memory cache hit rates over 
 Cluster metrics
 ~~~~~~~~~~~~~~~
 
-- ``mattermost_cluster_cluster_request_duration_seconds``:  The total duration in seconds of the inter-node cluster requests.
+- ``mattermost_cluster_cluster_request_duration_seconds``: The total duration in seconds of the inter-node cluster requests.
+- ``mattermost_cluster_cluster_health_score``: A score that gives an idea of how well it is meeting the soft-real time requirements of the gossip protocol.
 - ``mattermost_cluster_cluster_requests_total``: The total number of inter-node requests.
-- ``mattermost_cluster_event_type_totals``: The total number of cluster requests sent for any type.
+- ``mattermost_cluster_cluster_event_type_totals``: The total number of cluster requests sent for any type.
 
 Database metrics
 ~~~~~~~~~~~~~~~~
 
+- ``mattermost_db_active_users``: The total number of active users.
+- ``mattermost_db_cache_time``: Time to execute the cache handler.
 - ``mattermost_db_master_connections_total``: The total number of connections to the master database.
 - ``mattermost_db_read_replica_connections_total``: The total number of connections to all the read replica databases.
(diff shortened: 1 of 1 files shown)


cwarnermm (approved):
Really looking forward to the say when we can automate our way out of these manual updates, @agarciamontoro! Thank you for taking a first pass at getting there, and for the great updates!

agnivade (approved):
This was much needed!

I am wondering if you could outline the steps that you performed? Mainly for others to repeat it when more metrics get added.

agarciamontoro:
My idea was to use [the Gather function in the registry](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Gatherer) and write a script that simply called that. However, it's not as easy:
- The `go_sql_*` metrics need to be registered explicitly. This is doable, it's just those ones. 
- The `*Vec` metrics need to be observed at least once so that those metrics are reported back by `Gather`. This is error-prone, there are multiple ones, and doing this manually it's terrible

You can see the test I wrote in [this commit](https://github.com/mattermost/mattermost/commit/6238fce4e3835974e99c00d7f11e0ef6dc80586a). That prints the metrics in the same order we had in the docs, so that it was easy to compare if my test was missing some registered metrics (i.e., if there were some *Vec metrics whose Observe/Increment/Decrement functions I had not called). That happened quite a lot, so it was a very manual process, and we'd need to maintain that list of calls to the Observe/Increment/Decrement functions. Not good.

If we want to really automate this, I see two alternatives:
1. Use custom functions wrapping the `prometheus.New*` and `MustRegister`, so that we can collect the names and descriptions in a list we can later export.
2. Use `reflect` to inspect the calls to `MustRegister` and collect the names and descriptions of the passed arguments.

agnivade:
Both solutions look fine to me. If you wanted to open an HW ticket on this, please feel free!

Issue #7361, opened 2024-08-22
Install Desktop App: Wrong path in Windows File Explorer troubleshooting


Movion:
In the troubleshooting guide for the desktop app, the windows paths are showing double backslashes `\\`

`%APPDATA%\\Mattermost`
Which is wrong - correct would be just one:

`%APPDATA%\Mattermost`

I found the issues on these two lines:
https://github.com/mattermost/docs/blob/70b50ccb6dbc7bb282034171a46f16077dba55d3/source/collaborate/install-desktop-app.rst?plain=1#L242

https://github.com/mattermost/docs/blob/70b50ccb6dbc7bb282034171a46f16077dba55d3/source/collaborate/install-desktop-app.rst?plain=1#L259

devinbinnie:
@cwarnermm Yeah they're should only be one backslash in those paths.

cwarnermm:
Perfect - thanks, @devinbinnie. @Movion - Are you open to creating a docs PR with the necessary changes?

Movion:
@cwarnermm sure, but I just signed the "Mattermost Contributor Agreement", idk how fast I'm allowed to make a PR

cwarnermm:
@Movion - you're most welcome to get started any time (even if you hadn't signed the contributor agreement yet! It would be needed by the time we're ready to merge your changes into the docs code base).

Let us know here if you have any questions about the docs changes -- happy to help!

Movion:
I hope I did everything right.. please check and let me know if I have to improve or change anything :s

PR #7362, opened 2024-08-22
Fix windows %appdata% path

Movion:
#### Summary

This pull fix the document "install-desktop-app.rst" where the windows path %appdata% is wrong with two backlashes where only should be one.

#### Ticket Link

Fixes https://github.com/mattermost/docs/issues/7361


cwarnermm (approved):
Thanks, @Movion!

Issue #7380, opened 2024-09-05
Help Wanted: Re-implement an approved docs PR


cwarnermm:
https://github.com/mattermost/docs/pull/7296 is approved and ready to merge but is blocked from merge because a signed CLA isn't in place.

The request is to create a new Docs PR with the same proposed changes to match https://github.com/mattermost/docs/pull/7296 and for the PR to pass the CLA check so that the docs change can be successfully merged.

PR #7381, opened 2024-09-06
[GH-7380] Bugfix/move maximum payload size bytes

### LINKED ISSUE #7292: Wrong documentation regarding "MaximumPayloadSizeBytes"
#### Summary
Wrong documentation in terms of config parameter "MaximumPayloadSizeBytes". The parameter is located under path "ServiceSettings". Documentation states that it is under path "FileSettings".

Also see [reality](https://github.com/mattermost/mattermost/blob/db138fd23a6984ef94201cc2635ce0cdcf191f50/e2e-tests/cypress/tests/support/api/cloud_default_config.json#L112) vs [documentation](https://docs.mattermost.com/configure/environment-configuration-settings.html#maximum-payload-size).

#### Expected behavior
Adjust the documentation so that it is clear that one has to put "MaximumPayloadSizeBytes" under "ServiceSettings".

ChinoUkaegbu:
#### Summary
It moves the config entry MaximumPayloadSizeBytes to its correct location. It has nothing to do with storage, but it's rather an API restriction. Also the JSON path is adjusted since it currently incorrect.

#### Ticket Link
Fixes #7292 and #7380


cwarnermm (approved):
Thank you, @ChinoUkaegbu!

PR #7414, opened 2024-09-19
Remove horizontal table scroll

cwarnermm:
Introduces CSS that enables table wrapping on simple RST tables via ``.. table::`` Sphinx directive. Concept taken from [here](https://github.com/readthedocs/sphinx_rtd_theme/issues/117#issuecomment-41506687)


emdecr:
@cwarnermm As per our call – the new table at the top looks good! I think the 100% max-width did the trick? It's weird because I can't see the class you used `.wy-table-responsive` in the page markup. We can keep chatting, but TL;DR - your changes seem to have done the trick.

hmhealey:
I don't see the issue on the new table, but the offending code block there has been manually wrapped to not have the issue. I don't think the CSS that you added is actually being used

cwarnermm:
Confirmed that CSS is unnecessary. Reformatting all SQL code blocks that exceed the center pane width successfully eliminates the horizontal scroll and cut off text display.

PR #7449, opened 2024-10-07
[MM-45894] Add documentation for Calls EnableDCSignaling setting

streamer45:
#### Summary

Adding documentation for the new experimental `EnableDCSignaling` setting introduced in https://github.com/mattermost/mattermost-plugin-calls/pull/865

To be merged before MM v10.2 release.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45894


cpoile (approved):
As an overall question about this feature -- when would it cease to be experimental? It seems like a pretty harmless feature with only benefits.

streamer45:
> As an overall question about this feature -- when would it cease to be experimental? It seems like a pretty harmless feature with only benefits.

@cpoile Given this can affect how media tracks are shared, a simple bug could potentially cause serious communication issues. Being cautious makes sense to me, at least for a few release cycles since this has not been thoroughly tested in production yet.

cpoile:
Yep, makes sense -- just want clarity on what the conditions will be. Thanks!

source/configure/plugins-configuration-settings.rst
```diff
@@ -980,6 +980,26 @@ Enable AV1 (Experimental)
 | :ref:`simulcast is enabled for screen sharing <configure/plugins-configuration-settings:enable simulcast for screen sharing (experimental)>`.                          |
 +--------------------------------------------------------------------------+---------------------------------------------------------------------------------------------+
 
+Enable DC signaling (Experimental)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. config:setting:: plugins-enabledcsignaling
```
cwarnermm, line 986: Search metadata must be positioned directly before a page section title.

cwarnermm, line 986: At some point on this page, search meta data started being positioned below the section heading, rather than directly above the section header. It's my understanding that search meta data needs to be directly above the section heading to return results correctly.

streamer45: Thanks, not sure either, but it should look better now.

cwarnermm (approved):

streamer45:
I will go ahead and merge this since we are releasing the plugin to the Marketplace, and the documentation should be up by then.

PR #7450, opened 2024-10-08
Fix typo in LDAP logs config

hanzei:
#### Summary
Thanks, @DHaussermann, for catching that one

#### Ticket Link
NONE

### DIFF
--- a/source/manage/logging.rst
+++ b/source/manage/logging.rst
@@ -579,7 +579,7 @@ Please use the following JSON configuration as as starting point to enable trace
                 }
             ],
             "options": {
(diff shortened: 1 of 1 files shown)


DHaussermann (changes requested):
@hanzei this solves the typo but there may still be on an issue here. Based on what I see locally you need the filed to be `filename` and **not** `file_name` for the server to start.

Unless there is another supporting code change in server repo for this, a change may still be needed.

hanzei:
Great catch @DHaussermann :+1:

DHaussermann (approved):
Thanks @hanzei 

LGTM!

Please merge.

PR #7472, opened 2024-10-11
feat: Create Azure Container Offer docs

toninis:
#### Summary

- Upgrade python and various components on our workflows 
- Create the Azure Container Offer documentation
- Changed a bit the orientation and the way it looks on the main site


toninis:
@cwarnermm Any idea why this happens ? 
image

source/install/install-kubernetes-aks.rst
```diff
@@ -0,0 +1,113 @@
...
+.. include:: ../_static/badges/allplans-selfhosted.rst
+  :start-after: :nosearch:
+
+You can use are supported `Azure Marketplace Container Offer <https://azuremarketplace.microsoft.com/en-us/marketplace/apps/mattermost.mattermost-operator>`__ to install Mattermost on your existing Azure Infrastructure.  
+
+
+Infrastructure Pre-requisites
+-----------------------------
+
+PostgreSQL v11.0+ Database
```
nab-77, line 13: Recommend this to be v13

nab-77 (changes requested):

cwarnermm:
@toninis - Yes, I can see some RST-specific syntax issues that are resulting in the docs output you're seeing. Are you open to me committing changes to this PR directly? If so, I'll pull your fork down to do so.

toninis:
> @toninis - Yes, I can see some RST-specific syntax issues that are resulting in the docs output you're seeing. Are you open to me committing changes to this PR directly? If so, I'll pull your fork down to do so.

Feel free of course . And take your time . Hopefully we will get this shipped next week . 
Do your thang 🚀

cwarnermm:
@toninis - I've applied editorial and syntax updates. Currently, the new Azure AKS deploy page is listed directly after the Kubernetes page in the nav pane. If you'd prefer to see a nested approach instead with parent/child pages, let me know!

source/install/install-kubernetes-aks.rst
cwarnermm: I've applied this feedback.

source/install/install-kubernetes-aks.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Mattermost relies on strong TLS certification in order to provide all the features to users. You need to have access to a DNS zone and be able to provide a valid TLS key and certificate for the Ingress Controller.
+
+Mattermost License and AKS Capacity  
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. include:: ../_static/badges/ent-only.rst
+  :start-after: :nosearch:
+
+If your deployment option is for more than ``100 users``, you must have more than 2 nodes on your AKS cluster to support High Availability, and you must provide a valid Mattermost License file.
```
saturninoabril, line 50: Though we don't have scaling info yet using Azure, I wonder if we should refer them (temporarily) to https://docs.mattermost.com/guides/scale-mattermost.html so they could have some guidance (by converting AWS resource to Azure).

source/install/install-kubernetes-aks.rst
saturninoabril: Should we bump it separately along with other instances from these docs that still reflect 11+?

saturninoabril (commented):
Thanks @toninis! Left couple comments but overall looks good to me.

source/install/install-kubernetes-aks.rst
toninis: It's not about scaling as such but giving the option to do so . We can limit the installation to some users if we want to but if someone purchases a license then they should be able to scale a bit.

saturninoabril (approved):
Great work @toninis!

toninis:
> @toninis - I've applied editorial and syntax updates. Currently, the new Azure AKS deploy page is listed directly after the Kubernetes page in the nav pane. If you'd prefer to see a nested approach instead with parent/child pages, let me know!

I am 3/5 but I think a nested approach is better meaning that we support Kubernetes installation but in the meantime we have cloud specific instructions on each cloud . 
WDYT ?

nab-77 (approved):
Great update. Agree that elevating Azure to same level as all deployments is the way to go (we might revisit this nesting when we have other K8s CSP deployments).

toninis:
> Great update. Agree that elevating Azure to same level as all deployments is the way to go (we might revisit this nesting when we have other K8s CSP deployments).

Perfect. We are good to merge then . @cwarnermm with your blessing as always 😄

source/install/install-kubernetes-aks.rst
cwarnermm: I've bumped other references to PG v11+ across the docs site to v12 as part of the v10.1 updates coming this week.

cwarnermm (approved):

PR #7506, opened 2024-10-21
Table checkmarks

cwarnermm:


matthewbirtch:
@cwarnermm are you able to center the check icons within the column so they better align with the headers?

cwarnermm:
@matthewbirtch - Unfortunately, no.

PR #7514, opened 2024-10-29
Desktop v5.10: Removed full screen mode on Linux

cwarnermm:
Documentation for: https://github.com/mattermost/desktop/pull/3151


devinbinnie (approved):
Just a thought - we might want to explain why this is missing in the docs, but if you don't feel it's necessary we can leave it at this.

cwarnermm:
@devinbinnie - I was thinking the same and that would fit well in our changelog notes. @amyblais - Are you open to drafting a changelog note that details why we're removing fullscreen mode for the Desktop App on Linux?

PR #7519, opened 2024-10-29
Pinned icon hidden when no messages pinned in channel

cwarnermm:
Documentation for: https://github.com/mattermost/mattermost/pull/28276


source/collaborate/save-pin-messages.rst
```diff
@@ -37,7 +37,7 @@ Save messages
 Pin messages
 ------------
 
-All members of a channel can pin important or useful messages to that channel. The list of pinned messages is visible to all channel members. There is no limit to the number of pinned posts in a channel.
+All members of a channel can pin important or useful messages to that channel. The list of pinned messages is visible to all channel members. There is no limit to the number of pinned posts in a channel. From Mattermost v10.2, the **Pinned messages** |pinned-messages| icon is hidden when no messages are pinned.
```
cwarnermm, line 40:
```suggestion
All members of a channel can pin important or useful messages to that channel. The list of pinned messages is visible to all channel members. There is no limit to the number of pinned posts in a channel. From Mattermost v10.2, the **Pinned** |pinned-messages| icon is hidden when no messages are pinned.
```

PR #7635, opened 2024-12-02
Scheduled messages

cwarnermm:
Documentation for: https://github.com/mattermost/mattermost/pull/28932

@harshilsharma63 - FYI This docs PR includes additional updates beyond scheduled messages.


source/collaborate/schedule-messages.rst
```diff
@@ -0,0 +1,43 @@
...
+- **Reschedule message**: Change when the message should be sent.
+- **Send now**: Confirm that you want to send the message immediately.
+
+.. image:: ../images/manage-scheduled-message.png
+    :alt: Once a message is scheduled, you can delete, edit, and reschedule it, copy the message text, or send it immediately.
+
+Troubleshoot scheduled messages
+-------------------------------
+
+If a scheduled message fails to send, you'll be alerted with a banner at the top of the Mattermost screen. Scheduled messages may fail for the following reasons:
```
harshilsharma63, line 38: > you'll be alerted with a banner at the top of the Mattermost screen

The alert banner is only visible in the scheduled tab (same place where you see the scheduled posts). 

Additionally, you also get a red badge for scheduled post count in the LHS if there are some scheduled posts with error. Note that the count of scheduled even in the red badge is a total, and not how many scheduled posts have error. Stating this as this can be a point of confusion.

Screenshot 2024-12-03 at 9 51 34 AM

source/collaborate/schedule-messages.rst
```diff
@@ -0,0 +1,43 @@
...
+
+- You're no longer a member of the channel.
+- The channel is archived.
+- The channel is read-only.
+- The Mattermost server didn't respond. Mattermost will try to send the message for up to 30 minutes.
```
harshilsharma63, line 43: This point isn't completely right. I propose updating this to something like "Mattermost server was down for more than 24 hours". We process scheduled posts up to 24 hours old, so if the server is down for a couple of hours, its fine, but more than 24 hours will cause some scheduled posts to be skipped.

source/collaborate/schedule-messages.rst
```diff
@@ -0,0 +1,43 @@
+Schedule messages
```
harshilsharma63, line 1: 1. Can we also highlight the fact that scheduled posts can have a post priority, attachments or basically anything that a regular post can contain.

2. Can we also highlight that scheduled posts also have to go through the following hooks, just like a regular post would- `MessageHasBeenPosted` and `MessageWillBePosted` on the server side at the scheduled time of posting, and  hooks registered via the `registerMessageWillBePostedHook` on webapp will be run at the tie of scheduling the post.

harshilsharma63 (approved):
Looks great other than a few enhancements!!!

source/collaborate/schedule-messages.rst
cwarnermm: @matthewbirtch - I didn't see the new badge in the [compass icons](https://github.com/mattermost/compass-icons) repository (though I may have missed it). Can you point me to these new icons, please?

source/collaborate/schedule-messages.rst
cwarnermm: @harshilsharma63 - Given that this docs PR focuses on end user documentation, this isn't the best place to cover things like hooks. Our dev docs may be a better place. Would you be open to drafting a dev docs update for scheduled messages?

source/collaborate/schedule-messages.rst
matthewbirtch: @cwarnermm The badge itself is not an icon that would be in compass icons. If you would like a PNG image here's one you can use:
image

matthewbirtch:
@cwarnermm here are a few images you can use. These might be a bit cleaner:

For 'manage scheduled messages'
image

For troubleshooting
image

These are @2x resolution, but should be set to 600 width

source/collaborate/schedule-messages.rst
```diff
@@ -0,0 +1,43 @@
...
+==================
+
+.. include:: ../_static/badges/ent-pro-cloud-selfhosted.rst
+  :start-after: :nosearch:
+
+There are times when you want to send a message, but you don't want it sent immediately. For example, it may be after working hours for the recipient. 
+From Mattermost v10.3, using Mattermost in a web browser or the desktop app, you can schedule messages to send in the future. 
+
+Compose a message, then select the right side of the **Send** icon |send-icon| to schedule when the message will be sent. You can choose from a fixed time or set a custom time.
```
matthewbirtch, line 10:
```suggestion
Compose a message, then select the right side of the **Send** button |send-icon| to schedule when the message will be sent. You can choose from a fixed time or set a custom time.
```

Should we consider something more like this image? 
image

matthewbirtch:
@asaadmahmood can you review this?

asaadmahmood:
Sure @matthewbirtch

cwarnermm:
Thanks for all the great feedback, @harshilsharma63! I've incorporated your feedback in my latest commit. 

@asaadmahmood - A new generated docs preview will be generated using my latest commit. Looking forward to any feedback you have.

cwarnermm:
Looking forward to your input, @asaadmahmood!

asaadmahmood (approved):
LGTM.
Not sure if we want to add any additional detail like how you cannot reschedule a failed message, or edit files on scheduled messages, etc.

PR #7673, opened 2024-12-26
Update mmctl export create docs with additional options

wetneb:
#### Summary

This adds a couple of new options to the `mmctl export create` command:
* `--no-roles-and-schemes` and `--include-profile-pictures` (which are already supported in Mattermost 10)
* `--team`, which will be introduced by https://github.com/mattermost/mattermost/pull/29406.

This PR should only be merged once https://github.com/mattermost/mattermost/pull/29406 is.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-61997

### DIFF
--- a/source/manage/mmctl-command-line-tool.rst
+++ b/source/manage/mmctl-command-line-tool.rst
@@ -2292,7 +2292,10 @@ Create an export file including message attachments.
 .. code-block:: sh
 
    --no-attachments              Omit file attachments in the export file.
+   --no-roles-and-schemes        Exclude roles and custom permission schemes from the export file.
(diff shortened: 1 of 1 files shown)


cwarnermm:
Thank you, @wetneb! Are you open to splitting this PR into two PRs? One PR would contain the missing options, and the other would contain the new option tied to https://github.com/mattermost/mattermost/pull/29406.

wetneb:
I only added the existing options in passing and splitting this in two PRs would likely generate a merge conflict later on since they touch the same area, so I am not so motivated to go down that route. I hope you understand.

PR #7687, opened 2025-01-09
Update manual-postgres-migration.rst

vish9812:
We need to copy the MM config from MySQL DB not postgres.
Thread: https://hub.mattermost.com/private-core/pl/u8mxhd9psf858y8j79bhdwq1ry


isacikgoz (approved):
Good catch, thanks @vish9812

PR #7693, opened 2025-01-10
Remove server push

otbutz:
#### Summary

HTTP/2 server push support has been removed from nginx since version 1.25.1.

Browser support is long gone:

* Chrome dropped support in 106
* Firefox dropped support in 132


cwarnermm (approved):
Thanks, @otbutz!

PR #7731, opened 2025-02-06
Deployment documentation updates

nickmisasi:
#### Summary
I took a pass of each of the main installation methods, and performed a dry run on a new EC2 instance for each. I have made some documentation updates. For most of these updates, it's primarily for clarity. I also made some adjustments to split out some things that are repeatedly written between the various deployment methods to be more DRY. 

The most major change was the overhaul of the Docker deployment. The current docs "for production use" have the reader deploy Mattermost and its database on the same instance. This isn't the best advice in production. I've split the documentation into 2 paths here - Quick Start (essentially the existing docs), and Production Use (adjustments to only deploy the Mattermost application container, suggesting external hosting for db and storage). These changes require 

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-8785
https://mattermost.atlassian.net/browse/CLD-8848
https://mattermost.atlassian.net/browse/CLD-8849
https://mattermost.atlassian.net/browse/CLD-8850
https://mattermost.atlassian.net/browse/CLD-8851
https://mattermost.atlassian.net/browse/CLD-8852
https://mattermost.atlassian.net/browse/CLD-8853


cwarnermm:
This is a great docs PR, @nickmisasi! I pulled your branch down and applied fixes to build errors and warnings where needed. Remind me: what's the goal of this docs update? What are we trying to achieve? That will help me formulate potential next step recommendations.

nickmisasi:
@cwarnermm High level, this is intended to make it easier to deploy Mattermost. The main focus was on clarity, as well as making sure that our recommendations are not only correct, but actually work (without prior knowledge of deploying Mattermost itself)

cwarnermm:
@nickmisasi - Thank you! From my perspective, the content of this docs PR is excellent. No concerns at all. I know our structure and navigation needs work, but that's beyond the scope of this lift. Does it make sense to add customer-facing stakeholders as additional technical reviewers at this time? Are there specific perspectives you're looking for prior to us publishing these details?

nickmisasi:
@cwarnermm I think it's a good idea to get some stakeholders to view this. Particularly I'd like eyes from @mrckndt, as he has some past work on Docker docs and is probably most familiar with RHEL. I'm open to others if you have anyone in mind. I'll also loop in @toninis as we discussed the Docker docs updates in Hub.

nickmisasi:
Adding @gabrieljackson as I've pushed an overhaul to the Kubernetes documentation

source/install/install-docker.rst
```diff
@@ -150,16 +141,106 @@ To deploy Mattermost on Docker:
       
 7. Create your first Mattermost system admin user, :doc:`invite more users </collaborate/manage-channel-members>`, and explore the Mattermost platform. 
 
+Deploy Mattermost on Docker (Production)
```
toninis, line 144: I do not think we should even consider this as an option for Production use . 
It's best to redirect them to omnibus documentation and clearly specify that this is a deprecated way to deploy Mattermost and unmaintained .

Below we have a note stating that the repo is deprecated , we should not provide this as a viable production solution.
```
The `mattermost-docker <https://github.com/mattermost/mattermost-docker>`__ GitHub repository is deprecated. 
```

toninis (approved):
Overall looks good . Once small consideration around docker for production ... 
cc: @spirosoik @stylianosrigas

source/install/install-docker.rst
nickmisasi: I've had some discussions around this as well, and I tend to agree with you. That being said, I'm wondering if we should keep it (but perhaps add a "this is no longer recommended" note) as there are likely a bunch of customers using this method right now?

spirosoik: Docker can be for production only in cases like ECS (AWS) and Azure Container Apps. Do we have any idea if there's any deployment like this?

cwarnermm: @spirosoik @toninis @svelle -- what's the right way forward here?

spirosoik: Also I think there's no option for HA in docker.

svelle: > Docker can be for production only in cases like ECS (AWS) and Azure Container Apps. Do we have any idea if there's any deployment like this?

I don't fully understand the point here, we have lots of customers running docker in production. It likely won't be big enterprises but I'm (don't have hard numbers) that especially of SMEs are using Docker on bare metal or ECS deployments. It's a low maintenance deployment method that requires little knowledge in addition to basic linux.

cwarnermm: @spirosoik - What's your recommended way forward here? A number of deployment docs changes are queued up behind this one, so we're looking to merge this as soon as possible.

cwarnermm: @sadohert - Taking @toninis and @spirosoik's input into account, can you help us understand the customer's preferences here with regards to Docker deployments? Is Docker an enterprise deploy path we want to continue to offer?

We need to align on a path forward re: Docker deployments and enterprise guidance.

sadohert: > Docker can be for production only in cases like ECS (AWS) and Azure Container Apps. Do we have any idea if there's any deployment like this?

We have one **VERY** big customer running on ECS/Docker (with High Availability)... but I only know of this one in the "Super Large" end of the spectrum.  Something to note on this, that should probably go in these Docker docs, is that we went through a very challenging upgrade exercise with them in 2023, and one of the key changes we had them make was to NOT use Elastic/Dynamic scaling features with the MM App Servers (and I believe for the DB as well).

Personally, there's very low chance I'd lead a customer down the path of Docker, but I'm in agreement with keeping the docs around.

spirosoik: Thanks @sadohert super useful. Let's remove it from production deployment completely and put it only for testing purposes.

cwarnermm: ^ @nickmisasi - Can you update this docs PR based on Spiros' direction, please? Once the updates are in, I'll add the last couple redirects for the Docker page, approve the PR and we can merge it.

source/install/install-kubernetes.rst
```diff
@@ -1,110 +1,100 @@
...
-    - See the :doc:`configuration settings </configure/configuration-settings>` documentation to learn more about customizing your production Mattermost deployment.
-    - Encountering issues with your deployment? See the :ref:`deployment troubleshooting <install/troubleshooting:deployment troubleshooting>` documentation for details.
+While this guide focuses on using external, managed services for your database and file storage, the Mattermost Operator *does* offer the flexibility to use other solutions. For example, you could choose to deploy a PostgreSQL database within your Kubernetes cluster using the CloudNative PG operator, or use a self-hosted MinIO instance for object storage. 
 
-Install Mattermost on Kubernetes
---------------------------------
+However, managing these components yourself adds complexity and is generally not recommended for production deployments unless you have specific operational requirements. This guide will primarily focus on the recommended approach of using S3-compatible storage and a managed database service. If you choose to use self-managed components, you will need to adapt the instructions accordingly.
```
gabrieljackson, line 21: I like how we are more opinionated of how to install and run Mattermost via the operator, but I think it would be good to walk back some of this wording a bit. In particular `generally not recommended for production deployments` should probably be clarified so our customers that want to own all of their data know they are using a supported setup.

nickmisasi: Updated

source/install/install-docker.rst
nickmisasi: I've removed the production route for docker and updated references to it. I've also closed the accompanying PR in mattermost/docker

nickmisasi:
@cwarnermm I think I've addressed everything here. Let me know if there's anything else you need from me!

source/install/install-docker.rst
spirosoik: adding @gabrieljackson  @coltoneshaw and @victorehernandez who may have some more feedback  for customers before we close this PR. What's your thoughts guys about the docker deployment in production?

From your experience with customers so far what's the number #1 install method from higher priority to lower priority?

source/install/install-docker.rst
```diff
@@ -4,48 +4,40 @@ Deploy Mattermost via Docker
 .. include:: ../_static/badges/allplans-selfhosted.rst
   :start-after: :nosearch:
 
+.. important::
+   This deployment method is not recommended for production environments. If you're looking to deploy a production Mattermost instance, we recommend using managed container services like AWS ECS, Azure Container Apps, or Google Cloud Run. These services provide better scalability, reliability, and maintenance features suited for production workloads.
```
coltoneshaw, line 8: I wonder if this would be better to suggest past a specific size. I can't see why a smaller organization shouldn't run Mattermost via Docker.

source/install/install-docker.rst
```diff
@@ -150,16 +129,37 @@ To deploy Mattermost on Docker:
       
 7. Create your first Mattermost system admin user, :doc:`invite more users </collaborate/manage-channel-members>`, and explore the Mattermost platform. 
 
+Configure SSO With GitLab (Optional)
```
coltoneshaw, line 132: Is this normal for the docker builds? Wondering if this needs to be moved to a new SSO page

source/install/install-docker.rst
coltoneshaw: Hey guys, I'd take the opposite approach here. Let's just suggest it for customers under a specific size like 200, since our [scaling for 200 docs](https://docs.mattermost.com/scale/scale-to-200-users.html) only suggest a single node _anyways_. In reality, it's a great quick way to spin up Mattermost when you're in the initial implementation phases of a small deployment. 

> From your experience with customers so far what's the number https://github.com/mattermost/docs/pull/1 install method from higher priority to lower priority?

- Tarball
- K8
- Docker

> I don't really understand why not to use the autoscale features.

Autoscale on Docker can take the nodes down in the middle of the day and cause some issues.

> Also I think there's no option for HA in docker.

You can do HA docker, either via a single compose or multiple. However, It's not useful here and I would steer away from it anyways.

source/install/install-docker.rst
nickmisasi: It was already there, so I left it. 0/5 on moving it out of this doc, but these instructions are Docker specific

source/install/install-docker.rst
cwarnermm: @spirosoik - Can we get a bit more specific about threshold guidance here?

source/install/install-docker.rst
cwarnermm: @spirosoik - How do we want to proceed based on this input? CC @nickmisasi

source/install/install-docker.rst
cwarnermm: Given that it's Docker-specific, I'm inclined to leave here and reconsider as part of the restructuring work in progress.

source/install/install-docker.rst
spirosoik: thanks @coltoneshaw @victorehernandez. Just to summarise, so we are all in agreement. Docker can be in production but with not a very straightforward easy path and that's the last option we offer and mostly for "trying" purposes.  Based on this, docker should not be used for production use but for trying/early phases/development. Do I get this correctly?

coltoneshaw: >  with not a very straightforward easy path

Fairly easy path with the work CRE has put into the docker containers

> Based on this, docker should not be used for production use but for trying/early phases/development.

Is a valid choice for testing/early phases or 200 seat deployments.

spirosoik: @coltoneshaw Let me rephrase it. it's straighforward to run, though based on the followings:

> autoscale on Docker can take the nodes down in the middle of the day and cause some issues."
> You can do HA docker, either via a single compose or multiple. However, It's not useful here and I would steer away from it anyways.

It's not for production use and mission-critical work. So we can make clear in the docs that Docker is for early stage POCs/testing with a sub-note, you can use it in prod with your own risk. @cwarnermm can frame it much better than me of course.

thoughts?

nickmisasi: @coltoneshaw I think the key here is that while the Docker deployment _can_ easily support 200 seat and smaller deployments, it doesn't support our goals of mission-critical communication, so we don't think it _should_.

In addition, I think it sets up the customer for a more difficult move down the road, should they decide to go for a proper highly available deployment (now you have to migrate your database on top of the application nodes, etc). Anecdotally I have heard of a few customers who were in the position of having followed this doc, ending up with a database in Docker, and then lacking the resources to move that database to RDS after the fact. 

We are looking to become more prescriptive in the way we recommend deployment methods, aiming to guide people to more managed or semi-managed gold standard deployments as these can drastically improve the user experience within Mattermost itself, as well as reduce the time to get up and running. 

There's no imminent plan to abandon support for Docker based deployments. But, we want to encourage users to choose other things. For example, if you plan to have under 200 seats, and are OK with your database being on the same instance as the application itself, you should probably just use Omnibus, as you'll be up and running quicker.

mrckndt (approved):

cwarnermm (approved):
Thank you, @nickmisasi, for this outstanding docs PR!

PR #7701, opened 2025-01-16
V10.5 Docs

amyblais:
Will add a release date to the changelog when ready.

PR #7737, opened 2025-02-11
Added support packet v2 details to docs

cwarnermm:
Documentation for: https://github.com/mattermost/mattermost/pull/29403

### DIFF
--- a/source/manage/generating-support-packet.rst
+++ b/source/manage/generating-support-packet.rst
@@ -15,18 +15,42 @@ Contents of a Support Packet
 
 A Mattermost Support Packet can contain the following files:
 
-- `metadata.yaml <#metadata>`__
-- ``mattermost.log``
-- ``plugins.json``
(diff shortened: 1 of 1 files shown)


hanzei (approved):
LGTM :+1:

hanzei:
This section is no longer accurate:

> Plugins are not sanitized during packet generation. Ensure you sanitize any additional confidential details in the plugin.json file before sharing it with Mattermost. Replace details with example strings that contain the same special characters if possible, as special characters are common causes of configuration errors.


Since https://github.com/mattermost/mattermost/pull/27986, plugins can mark fields as secret, which then don't show up in the Support Packet. Mind updating that as part of this PR?

cwarnermm:
Thanks, @hanzei! I've applied that update to [the](https://github.com/mattermost/docs/pull/7701) (the v10.5 branch).

PR #7750, opened 2025-02-18
Update installing-license-key.rst

svelle:
added clearer wording for license checks to better explain what checks the server makes when uploading a new license.


cwarnermm (approved):
Thanks, @svelle!

PR #7791, opened 2025-03-13
Update install-docker.rst

jlandells:
Removed `sudo` from the docker commands, as per this conversation: https://hub.mattermost.com/private-core/pl/oc8d1ctjjbrfum9w71dxden5dr

### DIFF
--- a/source/install/install-docker.rst
+++ b/source/install/install-docker.rst
@@ -103,7 +103,7 @@ To deploy Mattermost on Docker:
 
    .. code-block:: sh
   
-      sudo docker compose -f docker-compose.yml -f docker-compose.without-nginx.yml up -d
+      docker compose -f docker-compose.yml -f docker-compose.without-nginx.yml up -d
 
    To access your new Mattermost deployment, navigate to ``http://<YOUR_MM_DOMAIN>:8065/`` in your browser.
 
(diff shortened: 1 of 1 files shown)


source/install/install-docker.rst
```diff
@@ -24,6 +24,9 @@ If you don't have Docker installed, follow the instructions below based on your
 
   Follow the `Install Docker Engine on Ubuntu <https://docs.docker.com/engine/install/ubuntu/>`_ documentation.
 
+  .. important::
+     Please ensure that you follow the post-installation steps when installing Docker on Ubuntu!
```
cwarnermm, line 28: is there a link we can add to clarify what's needed of the admin here?

jlandells: It's already linked at the bottom of the install page.  Do you think we should add a link to this page (https://docs.docker.com/engine/install/linux-postinstall/) here as well?

cwarnermm, line 28:
```suggestion
     Please ensure that you follow the `optional post-installation steps <https://docs.docker.com/engine/install/linux-postinstall/>`_ when installing Docker on Ubuntu!
```

cwarnermm (approved):

PR #7792, opened 2025-03-13
Update system roles feature name in system console path

esethna:


cwarnermm (approved):
Great catch, @esethna! Thank you!

PR #7801, opened 2025-03-17
install-desktop-app.rst: Fix macOS App Store config path

lathiat:
#### Summary
The macOS App Store config path was incorrect, missing the .desktop suffix on Library/Containers/Mattermost.desktop/

Previously the documentation suggested ~/Library/Containers/Mattermost/

#### Review Notes
I am somewhat fascinated by this path, every other directory on my system in ~/Library/Containers has a reverse-domain name (e.g. would be com.mattermost.xxx), Mattermost is the sole exception for me. I would have assumed that to be enforced by Apple.

In any case, I can find multiple references to this path on Google and checked two different machines. This path appears to be correct, at least currently.

#### Ticket Link
None, I just noticed the issue when trying to use the documentation


cwarnermm:
@rahimrahman - Can I get your help to technically review/approve this PR, please?

rahimrahman:
@lathiat it's different on my local machine.  Perhaps I need to pull in another expert for this matter.  @devinbinnie -- can you assist?

Screenshot 2025-03-17 at 07 08 24

devinbinnie:
@rahimrahman The contributor is technically correct here, this is just Apple obfuscating folder names, I think they do that for folders as well when you hide file extensions:

image

@cwarnermm Not sure if we want to call it out in the docs (ie. it's `Mattermost` or `Mattermost.desktop` depending on where you look)

rahimrahman:
@devinbinnie it's not case-sensitive?  Your folder is `Mattermost.Desktop` while the change proposed is `Mattermost.desktop`.

devinbinnie:
> @devinbinnie it's not case-sensitive? Your folder is `Mattermost.Desktop` while the change proposed is `Mattermost.desktop`.

Doesn't seem to care on my machine, though for "correctness" we should probably capitalize it.

image

source/collaborate/install-desktop-app.rst
```diff
@@ -190,7 +190,7 @@ The location of the Mattermost desktop app configuration file depends on the pla
 
 - Windows: ``Users\<username>\AppData\Roaming\Mattermost``
 - macOS installer: ``/Users/<username>/Library/Application Support/Mattermost``
-- macOS App Store: ``/Users/<username>/Library/Containers/Mattermost/Data/Library/Application Support/Mattermost``
+- macOS App Store: ``/Users/<username>/Library/Containers/Mattermost.desktop/Data/Library/Application Support/Mattermost``
```
cwarnermm, line 193:
```suggestion
- macOS App Store: ``/Users/<username>/Library/Containers/Mattermost.Desktop/Data/Library/Application Support/Mattermost``
```

devinbinnie:
> @devinbinnie - Copilot says that while macOS sometimes "simplifies" visible file paths for user-friendliness, this is NOT the case for app configuration directories in `~/Library`, as these paths are not user-facing and won't hide the `.desktop` extension arbitrarily.
> 
> Instead, Copilot proposes that the path differs based on how it was installed via the macOS App Store versus downloaded directly.
> 
> * If installed via the macOS App Store, the configuration path is commonly named after the application bundle identifier or app display name, which could simply be Mattermost. This is because macOS recognizes the app as an officially signed and sandboxed application from Apple’s ecosystem.
> * If installed via a direct download (for example, from the Mattermost website), the app might use a path like `Mattermost.desktop`, reflecting its distribution as a standalone, unsigned, or differently packaged application.
> 
> If you agree, I'd like to add that level of granularity to the docs as updated guidance.

That somewhat contradicts how our application is installed though, which is done through Mac App Store entirely. We DO have a self-packaged version as well, but that does not create a containerized installation and just puts the file in your local library:

image

Looks like Copilot has it backwards :P

cwarnermm:
Awesome! Thanks, Copilot! @devinbinnie - can you summarize for me the conditions in which the path _may_ change? I'll add that as a note to the docs for clarity.

devinbinnie:
I don't believe it does "change" really, I think it's pretty static.

In short:
`~/Library/Containers/Mattermost.Desktop` for MAS, except in Finder it will read `Mattermost` as it hides the extension
`~/Library/Application Support/Mattermost` for the self-packed version

source/collaborate/install-desktop-app.rst
```diff
@@ -190,7 +190,7 @@ The location of the Mattermost desktop app configuration file depends on the pla
 
 - Windows: ``Users\<username>\AppData\Roaming\Mattermost``
 - macOS installer: ``/Users/<username>/Library/Application Support/Mattermost``
-- macOS App Store: ``/Users/<username>/Library/Containers/Mattermost/Data/Library/Application Support/Mattermost``
+- macOS App Store: ``/Users/<username>/Library/Containers/Mattermost.Desktop/Data/Library/Application Support/Mattermost``
```
cwarnermm, line 193:
```suggestion
- macOS App Store: ``/Users/<username>/Library/Containers/Mattermost.Desktop/Data/Library/Application Support/Mattermost`` (via Finder: ``~/Library/Application Support/Mattermost`` as the extension is hidden)
```

PR #7807, opened 2025-03-18
Updated config.json example blocks

ewwollesen:
I didn't like how the codeblocks were formatted with the `"Enable":` line being not indented with the rest of the code. I could be convinced to leave off the `"Office365Settings":` and `"GoogleSettings":` line since that is mentioned in the line above it as part of the instructions, but the first line not being indented I think looked off.


cwarnermm (approved):
Thanks, @ewwollesen!

PR #7826, opened 2025-03-25
Update installing-license-key.rst

cwarnermm:


source/upgrade/installing-license-key.rst
```diff
@@ -23,6 +23,7 @@ You can use the System Console or the mmctl tools to add or change a Mattermost
 
 .. note::
 
+    - Enterprise customers with the Premier Support add-on can access a staging license for testing.
```
svelle, line 26:
```suggestion
    - Enterprise customers with the Premier Support add-on can request a staging license for testing.
```

this should probably be "request" not "access" because we don't provide them outright.

svelle (changes requested):
minor nitpick

cwarnermm:
Very worthwhile clarification!

svelle (approved):
lgtm

PR #7831, opened 2025-03-26
fix sitemap url scheme in init.py

svelle:
the url scheme in the init config uses a language prefix that we actually don't use in the deployment of the real docs website, this causes all links to be invalid. 

this pr removes this prefix making the sitemap generation work again.


svelle:
See production sitemap here: https://docs.mattermost.com/sitemap.xml

And the fixed sitemap here: http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/7831/sitemap.xml

cwarnermm (approved):

PR #7832, opened 2025-03-26
Update performance-alerting.rst

jlandells:
Updated instructions to refer to the newer version of Grafana that no longer has Notification Channels


cwarnermm (approved):
Thanks, @jlandells!

PR #7845, opened 2025-04-04
Update default value for Enable Image Proxy setting

jadrales:
Changing the default value for the Enable Image Proxy setting from `true `to `false`.

This was updated in v5.9 but was never changed to be reflected in the docs: https://docs.mattermost.com/about/unsupported-legacy-releases.html#id660

### DIFF
--- a/source/configure/environment-configuration-settings.rst
+++ b/source/configure/environment-configuration-settings.rst
@@ -2434,8 +2434,8 @@ An image proxy is used by Mattermost apps to prevent them from connecting direct
   :configjson: .ImageProxySettings.Enable
   :environment: MM_IMAGEPROXYSETTINGS_ENABLE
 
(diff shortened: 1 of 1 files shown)


cwarnermm (commented):
Thanks, @jadrales!! Please also update the default in the table below. Config settings consist of 2 pieces of code: one is metadata to serve back the config setting in search results; the other is the docs content itself.

ewwollesen:
@jadrales Thanks for doing this.

jadrales:
> Thanks, @jadrales!! Please also update the default in the table below. Config settings consist of 2 pieces of code: one is metadata to serve back the config setting in search results; the other is the docs content itself.

Thanks for catching that, Carrie! I have updated the default value in the table as well.

ewwollesen (approved):
Checked changes in preview environment and everything looks correct.

cwarnermm (approved):

PR #7848, opened 2025-04-07
MM-63660: Added client-side filtering capability for upgrade notes and changelog

agnivade:
We add JS and CSS code to dynamically filter the rendered
HTML files in the upgrade notes and changelog pages.

I tried several approaches here before setting down on this one.

My initial approach was to convert the .rst file to a JSON file
during the build process and then dynamically load the JSON file.
But that had issues because the .rst contents had links and code
which needed to be formatted correctly.

The second approach was to see if it's possible to template
this on the server side. This would have been the right solution
here because the current solution requires parsing through HTML
which is brittle. But unfortunately, this would require changing
the architecture of the app from a static HTML app to being served
with Python. While that is possible, it is out of scope.

Therefore, we settle down on this approach where we include JS
to filter through the rendered HTML files and apply a bit of CSS
to polish the UI.

The result works fine. But it's likely a bit brittle because of
the lack of ability to control the rendered HTML.

Most of it assisted by Claude

https://mattermost.atlassian.net/browse/MM-63660


agnivade:
Perhaps I didn't get you. Everything is already done in this PR.

http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/7848/upgrade/important-upgrade-notes.html
http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/7848/about/mattermost-v10-changelog.html

Were you looking for something else?

cwarnermm:
@agnivade - WOW! I didn't realize that both the changelog and the important upgrade notes had the drop-downs incorporated! That's incredible!

If we changed how the content on these pages was formatted (like the table format on the important upgrade notes page), would these drop-downs continue to work? Or are the drop-downs tied to the page formatting in some key way?

agnivade:
Yes, that is the tradeoff here. If you make major changes, like change the table to something else, then yes it's liable to break. There is no other good solution here (as outlined in my PR message). The only proper solution is to change the app architecture to render server side and return the contents which would be a major change.

I'd say if you like this solution, we can go ahead with this. And if this kind of dynamic logic gains more popularity, we can look towards changing the app architecture.

cwarnermm:
One piece of feedback from our Success Team: "Maybe have a line to separate supported vs. EOL versions in the target?"

I'll share the preview in the PDE Meeting channel to invite more feedback. Let's aim to merge this next week.

agnivade:
> One piece of feedback from our Success Team: "Maybe have a line to separate supported vs. EOL versions in the target?"

This will need extra tracking within the rst page to differentiate what is supported and what is EOL. I'd say this is out of scope for what we are trying to do. And we probably don't even need JS for this. We should just add a static tag mentioning that release X is EOL.

agnivade:
@cwarnermm - PTAL

cwarnermm (approved):
Thank you, @agnivade!!!

agnivade:
@cwarnermm - Is this okay to be merged?

cwarnermm:
Yes! @agnivade, this is awesome! Thank you so much! Please merge when ready!

PR #7859, opened 2025-04-08
Removed specific Postgres version string

sadohert:
I noticed we mentioned "PostgreSQL v11.7" specifically, but we are deprecating (have deprecated?) support for that version.

For simplicity, removing any mention of version.


cwarnermm (approved):
Thanks, @sadohert

PR #7878, opened 2025-04-11
Adding documentation on Redis

agnivade:


cwarnermm (approved):
Thank you, @agnivade!!

source/scale/redis.rst
```diff
@@ -0,0 +1,109 @@
...
+- ``platform/services/cache/provider.go`` - Contains the library initialization code
+- ``platform/services/cache/redis.go`` - Contains the cache interface implementation
+
+Frequently asked questions (FAQ)
+--------------------------------
+
+Do I need to use Redis?
+~~~~~~~~~~~~~~~~~~~~~~~
+
+While Redis is not required for smaller Mattermost deployments, it is highly recommended for large-scale installations, especially those with more than 100,000 users. Redis significantly improves performance by reducing database load and optimizing common operations.
```
coltoneshaw, line 92: @agnivade - Are we saying there's no benefit to Redis sub 100k?

source/scale/redis.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Should I install Redis on the same machine as Mattermost Server?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For production deployments, we recommend running Redis on a separate machine from the Mattermost server. This separation allows for better resource allocation and improved system reliability.
+
+What Redis metrics should I monitor?
```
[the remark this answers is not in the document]
agnivade: Yes, last line is that.

source/scale/redis.rst
agnivade: Would you prefer to remove that line? Honestly speaking, there isn't really that much of an improvement at low scale.

source/scale/redis.rst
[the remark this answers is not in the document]
streamer45: Do you feel we should keep https://docs.mattermost.com/scale/performance-monitoring-metrics.html#custom-mattermost-metrics updated as well for visibility?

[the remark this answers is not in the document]
agnivade: Oh yes, there is a ticket for that. Can't find it atm.

source/scale/redis.rst
streamer45: Maybe my interpretation is biased, but I don't see `not required` as meaning it's not beneficial at all. I think that having an extra component in a deployment is always a tradeoff to consider so we probably wouldn't recommend it for smaller deployments.

cwarnermm: @agnivade - Is the next step here to clarify that Redis isn't recommended for smaller deployments under 100K?

cwarnermm: @agnivade - I recommend being more explicit and prescriptive to ensure that smaller customers understand whether Redis applies to their scenario.

agnivade: Understood. How about something like this:

```
- If your installation has more than 100K users, Redis is highly recommended.
- If your installation has less than 100K users, Redis is not recommended.
```
?

source/scale/redis.rst
```diff
@@ -0,0 +1,111 @@
...
+
+.. include:: ../_static/badges/ent-cloud-selfhosted.rst
+  :start-after: :nosearch:
+
+Redis is an open-source, in-memory data structure store used as a database, cache, message broker, and streaming engine. Mattermost uses Redis as an external cache to improve performance at scale. When properly configured, Redis can help support Mattermost installations with more than 100,000 users by providing improved performance through efficient caching.
+
+Deployment guide
+----------------
+
+Deploying Redis with Mattermost includes the following two steps: `setting up the Redis server <#set-up-a-redis-server>`__, and `configuring Redis in Mattermost <#configure-redis-in-mattermost>`__.
```
cwarnermm, line 12:
```suggestion
Deploying Redis with Mattermost includes the following 2 steps: `setting up the Redis server <#set-up-a-redis-server>`__, and `configuring Redis in Mattermost <#configure-redis-in-mattermost>`__.
```

cwarnermm (approved):
Thank you, @agnivade!

source/scale/redis.rst
streamer45: @agnivade I know we typically refer to "users" for simplicity, but I find it unrealistic for customers to think in such strict terms. For example, what if a customer has 80,000 users? Perhaps we could suggest that above 50k users (just a random example), it's recommended to start considering Redis, and above 100k users, it becomes highly recommended.

More broadly, do you think we could define a few self-diagnostic guidelines for when Redis might become necessary, for example, by identifying specific thresholds in Grafana?

agnivade: I believe Carrie was looking to have a more prescriptive guidance for the customers. If we introduce a third level into this (not needed, start thinking, absolutely needed), it might get complicated.

Re: specific thresholds - that's a great point. Yes, it's a correlation we are drawing with number of users with scale. Specifically, in this case, the number of nodes in your cluster, and rate of cache invalidations would be the threshold to monitor. I can come up with the exact prom query for that and update the PR.

streamer45: > If we introduce a third level into this (not needed, start thinking, absolutely needed), it might get complicated.

Yep, I completely understand. I just feel that making it a strict yes/no decision based on an arbitrary number (like 100k) doesn't add much value either. I would prefer to provide guidelines to help customers identify when it might be time to consider using Redis.

> Specifically, in this case, the number of nodes in your cluster, and rate of cache invalidations would be the threshold to monitor. I can come up with the exact prom query for that and update the PR.

That would be amazing :pray:

agnivade: PTAL @streamer45

cwarnermm:
@streamer45 @coltoneshaw - Looking forward to your input on this docs PR.

source/scale/redis.rst
```diff
@@ -0,0 +1,115 @@
...
+4. Monitor CPU utilization closely, as Redis is single-threaded and performance bottlenecks are typically CPU-bound rather than memory-bound.
+
+Performance considerations
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When using Redis with Mattermost, consider the following performance insights:
+
+- Current load tests indicate that memory usage of Redis is relatively low, with CPU being the primary bottleneck because Redis is single-threaded.
+- When planning your Redis deployment, use CPU utilization as the primary metric for scaling decisions.
+- For extremely large deployments, consider sharding your Redis deployment to distribute the workload.
```
streamer45, line 70: Did we use sharding during our 200k test? Mostly trying to understand whether this is something we tried internally before, or if there's an opportunity to explore this in future ceiling tests.

source/scale/redis.rst
```diff
@@ -0,0 +1,115 @@
...
+--------------------------------
+
+Do I need to use Redis?
+~~~~~~~~~~~~~~~~~~~~~~~
+
+While Redis is not required for smaller Mattermost deployments, it is highly recommended for large-scale installations.
+- If your installation has more than 100K users, Redis is highly recommended. Redis significantly improves performance by reducing database load and optimizing common operations.
+- If your installation has less than 100K users, Redis is not recommended.
+
+It is suggested to keep an eye on the `sum(rate(mattermost_cache_mem_invalidation_total[5m]))` metric which indicates the rate of cache invalidations happening. A single cache invalidation will send a message across your HA cluster to all nodes, leading to database calls from all of them. So the metric, combined with the number of nodes in your cluster should be used to determine when is the right time to use Redis.
```
streamer45, line 96: Thoughts on making `sum(rate(mattermost_cache_mem_invalidation_total[5m]))` code formatted rather than italic?

image

streamer45 (approved):
Thanks Agniva, looks great.

source/scale/redis.rst
agnivade: No we didn't. Let me actually remove this line until we have done it ourselves.

source/scale/redis.rst
agnivade: Done

PR #7910, opened 2025-05-01
Moved secure section to Other K8s

cwarnermm:
"Secure" section doesn't apply to Azure (redundant), so moved from K8s landing page to apply to only the "Other K8s" page. Will confirm with @DSchalla whether the "Secure" section applies to "Other K8s", or can be removed altogether.


wiersgallak (approved):

DSchalla:
I'd need input from @gabrieljackson / @spirosoik if the operator already takes care of these actions, if it does I'm supportive of removing and simplifying this. Thanks!

cwarnermm:
@gabrieljackson @spirosoik - Looking forward to your input on this docs PR.

gabrieljackson (commented):
Daniel was correct in his comment that the Operator manages and deploys the ingress resource for installations. As such this doesn't really apply so we should be okay to just remove it.

PR #7945, opened 2025-05-14
Clarify CLI docs

amyblais:
Removed references to an old v8.0.0 release.


cwarnermm (approved):
Thanks, @amyblais!

hanzei (commented):
Can we also drop the v8 notices around L108?

PR #7946, opened 2025-05-14
Add note to clarify Calls deployment in air-gapped environments

streamer45:
#### Summary

We are adding a note under network requirements to clarify that Calls can work in air-gapped environments, as long as all users (clients) can connect from within the internal/private network.

/cc @jasonblais in case you have comments


jasonblais:
Thanks Claudio! Only feedback is to have it further up in the document, it's a key differentiator in the market. I would move it right below the opening paragraph, before terminology.

streamer45:
@cwarnermm Perhaps we could add an additional note or a simple paragraph on top to incorporate Jason's feedback, which makes total sense. I still think the note I included fits as part of the network requirements since we mention more technical details (STUN, ICE Host Override).

PR #7949, opened 2025-05-15
Calls: Add full support for config environment overrides

streamer45:
#### Summary

With https://github.com/mattermost/mattermost-plugin-calls/pull/1004 in, we can finally update our docs to include full coverage of environment overrides for Calls plugin config settings.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62732


cpoile (approved):

cwarnermm (approved):
Thanks, @streamer45!

#radar: I noticed that the env vars for Calls diverge a bit from other env vars across the platform. The formula used takes the ``config.json`` setting, adds ``MM_`` to the beginning, converts all characters to uppercase, and replaces the ``.`` with ``_``. The config name itself becomes all one word with no underlines between words.

Given the established pattern, I wonder if this new pattern may result in calls env vars being slightly less guessable/discoverable.

streamer45:
> Thanks, @streamer45!
> 
> #radar: I noticed that the env vars for Calls diverge a bit from other env vars across the platform. The formula used takes the `config.json` setting, adds `MM_` to the beginning, converts all characters to uppercase, and replaces the `.` with `_`. The config name itself becomes all one word with no underlines between words.
> 
> Given the established pattern, I wonder if this new pattern may result in calls env vars being slightly less guessable/discoverable.

The concern is valid, but I made that choice because I find the new pattern superior in readability and far easier to debug as a consequence. I personally can't work well with something like `MM_SERVICESETTINGS_TLSSTRICTTRANSPORTMAXAGE` (an example to make the point).

cwarnermm:
That's an excellent point, @streamer45! Thank you for the context!

PR #7986, opened 2025-05-28
Removed Go mmctl install option docs

cwarnermm:


source/manage/mmctl-command-line-tool.rst
```diff
@@ -96,16 +96,6 @@ The mmctl tool comes bundled with Mattermost package. For customers that want to
...
-.. tab:: Go install 
-
-   Use this option on Linux, macOS, and Windows if you have a ``go`` environment configured.
-
-   To build and install the mmctl binary in your `$GOPATH`, run the following command:
-
-   .. code-block:: sh
-
-      go install github.com/mattermost/mattermost/server/v8/cmd/mmctl@master
-
```
hanzei, line 108: Building `mmctl` from a local git clone still works, right? Should we keep that as an alternative?

hanzei: The steps for that are:
```bash
git clone https://github.com/mattermost/mattermost.git
cd mattermost/server
go install ./cmd/mmctl
```

PR #8013, opened 2025-06-09
Add note on login rate for 200k users architecture

agarciamontoro:
#### Summary
This is a short note to reflect the results of the investigation on the initial load phase: https://mattermost.atlassian.net/wiki/spaces/XYZ/pages/edit-v2/3593928708

I'm not sure how to better word this. The data we have shows that we can login 150k users in under 1.5 hours, but after that, we don't have good data. I'm open to ideas to tackle this from another angle.

What we need is more testing, of course, but we should have something public with the data we already have.


source/scale/scale-to-200000-users.rst
```diff
@@ -14,6 +14,11 @@ This page describes the Mattermost reference architecture designed for the load
...
   - While the following Elasticsearch specifications may be more than sufficient for some use cases, we have not extensively tested configurations with lower resource allocations for this user scale. If cost optimization is a priority, admins may choose to experiment with smaller configurations, but we recommend starting with the tested specifications to ensure system stability and performance. Keep in mind that under-provisioning can lead to degraded user experience and additional troubleshooting effort.
 
+Ramp-up rate
+------------
+
+The tests for all architectures were designed to login 4 users per second; i.e., 14,400 users in an hour. For this architecture, we ran an additional test to better understand larger rates. The maximum rate we tested was 30 users per second, that was supported by this architecture for the first 150,000 users. That means that this architecture can login 150,000 users in under 1.5 hours. There is no hard data on larger rates after that 150,000 mark, apart from the base rate of 4 users per second.
+
```
cwarnermm, line 21:
```suggestion
User login scalability
-----------------------

Tests across all architectures were conducted at a rate of 4 user logins per second (14,400 users per hour). For this architecture, we performed additional testing at higher rates, reaching up to 30 user logins per second. 

Results show that this architecture supports logging in up to 150,000 users within 1.5 hours at the higher rate. Beyond this 150,000-user mark, the only available data is at the base rate of 4 logins per second, and no conclusive performance data exists at larger rates.

```

streamer45 (approved):
Discussed internally. This is a good interim improvement. Thanks!

PR #8057, opened 2025-06-16
Adding links to the Plans page

cwarnermm:


cwarnermm:
@claude - For each feature in the editions-and-offerings.rst file that includes a link, find the same feature in the plans.md file and add that feature link in HTML format.

PR #8062, opened 2025-06-17
remove locale from sitemap urls

svelle:
my previous PR caused the sitemap extension to add the translation string /en/ to every url. this needs to be removed to conform to the actual url.
    
correct: `https://docs.mattermost.com/agents/docs/sovereign_ai.html`
incorrect: `https://docs.mattermost.com/en/agents/docs/sovereign_ai.html`


jwilander:
@claude why did you run on this PR?

PR #8065, opened 2025-06-17
Remove the automatic Claude review workflow

jwilander:
#### Summary
Unlike on the monorepo, I included the auto-run Claude review job in the original PR for this repo. I'm proposing we remove it so Claude only runs when a user asks for it. Thoughts @cwarnermm ?

#### Ticket Link
n/a


cwarnermm:
Thanks, @jwilander, I agree. I'd prefer it to be on demand.

PR #8066, opened 2025-06-17
AI-Aided: Remove built-in grace period references from documentation

lieut-data:
## Summary
- Removed references to the built-in 5% grace period from the licensed user limit error documentation
- Updated error conditions to reflect current behavior without grace period

## Related PRs
- https://github.com/mattermost/mattermost/pull/31629

## Test plan
- [ ] Verify documentation accurately reflects current licensing behavior
- [ ] Review that all grace period references have been removed

🤖 Generated with [Claude Code](https://claude.ai/code)


cwarnermm (approved):

lieut-data:
@cwarnermm, no objections -- is this actually a pattern we should adopt going forward? I admit I was ignorant about branch specific updates until now. (Or possibly I'd just forgotten :))

Note too that I /think/ the previous version of this docs is already live, so not sure if we want to revert that or not.

cwarnermm:
@lieut-data - Documenting product functionality within the associated release branch helps Docs and Release Management manage the monthly maintenance updates and is the preferred approach.

In this specific case, given that the content is already published/live, I recommend publishing now rather than reverting and publishing in July.

PR #8199, opened 2025-07-18
Update minimum RHEL requirements to RHEL9 for Mattermost v10.10+

cwarnermm:
Updates the minimum RHEL requirements documentation to reflect RHEL9 minimum due to Go 1.24 upgrade in Mattermost v10.10.

## Changes
- Updated minimum RHEL version from 7+ to 9+
- Added important note about Linux kernel 3.2+ requirement
- Mentioned RHEL8 EUS end date (May 31, 2025)
- Referenced Go 1.24 requirement and kernel compatibility

Fixes #8198

Generated with [Claude Code](https://claude.ai/code)

PR #8246, opened 2025-07-22
added note on RHEL compatibility with 10.10

svelle:
* Added a note about **RHEL 7 and RHEL 8 Compatibility Issues**. Mattermost 10.10 and later require GLIBC 2.34, which is incompatible with RHEL 7 and RHEL 8. Users are advised to upgrade to RHEL 9 or use distributions with newer GLIBC versions before upgrading to Mattermost 10.10. This issue arises from the Go version upgrade (from 1.23 to 1.24) and consequential changes in the build environment.


cwarnermm:
closing in favour of https://github.com/mattermost/docs/pull/8199

PR #8261, opened 2025-07-31
Fix: Use port 443 instead of 9200 for AWS OpenSearch

rohithchandran-mattermost:
Updated Step 4 to replace the incorrect reference to port 9200 with port 443, which is the only supported port for AWS OpenSearch HTTPS access. Added a note clarifying that port 9200 applies to local/on-prem setups only.

Updated Step 4 in the AWS OpenSearch Console Setup section:
* Replaced the incorrect reference to port 9200 with port 443, which is the only supported port for AWS OpenSearch HTTPS access.
* Added a note to clarify that port 9200 is used only for local/on-premise OpenSearch setups, not for AWS-managed domains.

Reference - https://docs.aws.amazon.com/opensearch-service/latest/developerguide/vpc.html


source/scale/opensearch-setup.rst
```diff
@@ -136,7 +136,9 @@ We highly recommend that you set up an AWS OpenSearch server on a separate machi
     - master nodes: 2
     - storage: EBS gp3 (1536 GiB, 4608 IOPS, 250 MiB/s)
 
-  4. Specify the network for: VPC with 2 subnets, and a security group allowing Mattermost IPs on port ``9200``.
+  4. Specify the network for: VPC with 2 subnets, and a security group allowing Mattermost IPs on port ``443``.
+.. note::
+   Port 9200 is commonly used for local or on-premise OpenSearch. AWS OpenSearch domain only supports HTTPS over port 443.
```
cwarnermm, line 141:
```suggestion

.. note::
   Port 9200 is commonly used for local or on-premise OpenSearch. The AWS OpenSearch domain only supports HTTPS over port 443.
```

cwarnermm (approved):

PR #8419, opened 2025-09-21
Fix missing LHS TOC entries for pages with inline tabs and a custom toctree

neflyte:
#### Summary

This PR fixes the case where the left-hand side TOC is missing child document entries for documents with a custom `toctree` and inline tabs.

In sphinx-inline-tabs, custom `toctree` elements were not considered when updating a document's TOC with inline tab names and sub-sections. This PR ensures that custom `toctree` elements are preserved when the document's TOC is updated.

### DIFF
--- a/extensions/sphinx_inline_tabs/events.py
+++ b/extensions/sphinx_inline_tabs/events.py
@@ -1,7 +1,6 @@
 """
 Sphinx event handler implementation.
 """
-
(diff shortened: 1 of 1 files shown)


neflyte:
I've pushed a change that resolves the issue of the LHS not expanding to the expected location for pages with inline tabs.

PR #8488, opened 2025-10-22
Exclude redirect-only pages from sitemap

svelle:
this pr should blacklist all redirects from sitemap creation


svelle:
@cwarnermm would appreciate a review on this.

right now we're indexing all the sitemap urls that are redirects and I'm getting a ton of errors in zendesk for this.
this pr removes all the redirects from the sitemap since we don't need it in there anyway.

PR #8535, opened 2025-11-11
Fix dots in environment variables

domibarton:
#### Summary

Some environment variables in the docs were wrong (had a `.` instead of a `_` in it).  
I tested them on v11, and fixed them.


cwarnermm (approved):
Thanks, @domibarton!

domibarton:
> @domibarton - Are you open to signing the Mattermost CLA so that we can merge your proposed changes into the product documentation codebase?

yeah why not… where?

PR #8608, opened 2026-01-07
Remove MinIO recommendations and update file storage documentation

jwilander:
#### Summary
MinIO has gone into maintenance mode and as a result is no longer recommended as a file storage solution for Mattermost.

Updated documentation to recommend:
- AWS S3 or any S3-compatible service
- Local storage
- Network storage using NFS

My linter removed a bunch of trailing spaces so it might be easier to review with whitespace changes filtered out.

#### Ticket Link
n/a


DSchalla:
@jwilander I'm wondering if we should widen the definition a bit from NFS. It doesn't need to be specifically NFS(although that will be most common), but rather a network-mountable ReadWriteMany volume. Thoughts?

DSchalla:
Sounds good to me. Thanks Joram.

source/administration-guide/configure/environment-configuration-settings.rst
```diff
@@ -1925,7 +1925,7 @@ With self-hosted deployments, you can configure file storage settings by going t
 
 .. note::
 
-  Mattermost currently supports storing files on the local filesystem and Amazon S3 or S3-compatible containers. We have tested Mattermost with `MinIO <https://min.io/>`__ and `Digital Ocean Spaces <https://docs.digitalocean.com/products/spaces/>`__ products, but not all S3-compatible containers on the market. If you are looking to use other S3-compatible containers, we recommend completing your own testing.
+  Mattermost currently supports storing files on the local filesystem and Amazon S3 or S3-compatible containers. We have tested Mattermost with `Digital Ocean Spaces <https://docs.digitalocean.com/products/spaces/>`__, but not all S3-compatible containers on the market. If you are looking to use other S3-compatible containers, we recommend completing your own testing. You can also use local storage or a network drive using NFS.
```
lieut-data: Do we need to aggressively remove MinIO when we're keeping Digital Ocean Spaces? MinIO isn't going away, it's just not free anymore -- but Digital Ocean Spaces never was?

PR #8671, opened 2026-01-21
Docs: Update ABAC Channel Access Rules documentation for v11.3

### LINKED ISSUE #8666: v11.3 Docs: Tweak auto-add to make it consistent with child policies
@claude - Update the Mattermost Administration Guide section for ABAC Channel Access Rules to clarify:

- Each ABAC channel access policy has an explicit active state that determines whether rules are enforced and whether auto-sync applies.
- Channel-level (child) ABAC policies now behave independently and consistently, even when parent policies exist.
- Auto-sync behavior follows the channel policy’s active state, reducing unexpected inheritance-related behavior.
- Add a brief troubleshooting note advising admins to verify that a channel’s access control policy is active if auto-sync is not behaving as expected.

Do NOT:

- Re-document API endpoints or schemas already covered in engineering-owned docs
- Introduce speculative configuration options or defaults
- Add end-user guidance

Combs7th:
Updates the Mattermost Administration Guide section for ABAC Channel Access Rules to clarify:

- Each ABAC channel access policy has an explicit active state that determines whether rules are enforced and whether auto-sync applies
- Channel-level (child) ABAC policies now behave independently and consistently, even when parent policies exist
- Auto-sync behavior follows the channel policy's active state, reducing unexpected inheritance-related behavior
- Added troubleshooting note advising admins to verify that a channel's access control policy is active if auto-sync is not behaving as expected

Fixes #8666

Generated with [Claude Code](https://claude.ai/code)


isacikgoz (changes requested):
Thanks @Combs7th I think we might want to clarify the active state.

> **active state** that determines whether the policy's rules are enforced and whether automatic member synchronization applies to the channel.

Regardless of active state, policy's rules will be enforced to remove members if they don't conform to required attribute rules.

Combs7th:
@claude - Please update this doc text based on the feedback from @isacikgoz above.

Key corrections to reflect accurately:

- The policy being applied to a channel is always enforced (including removing members who no longer meet required attribute rules), regardless of the policy’s active state.
- The policy active state controls automatic member synchronization / auto-add behavior (i.e., whether users matching rules are automatically added/re-added), not whether enforcement happens at all.
- System-wide (parent) policies can be inherited at the channel level for rules, but inheritance is not the point for auto-add: auto-add/auto-sync is checked per individual channel policy.
- Reframe the “Auto-add members based on access rules” toggle explanation: it’s primarily about ensuring members are automatically added/re-added when they match (e.g., if attributes were temporarily lost and later restored), more like an alternative to LDAP group channels / a way to keep the channel consistently aligned to an attribute set—not about inheritance reducing surprises.

What I need:

- Rewrite the relevant doc sections so they’re accurate and clear, using the above points.
- Remove or revise any statement implying the active state determines whether rules are enforced.
- Remove or revise any statement framing auto-sync behavior as “inheritance-related,” and instead explain the admin control/use-case described above.
- Keep the edits concise and consistent with Mattermost docs tone.

isacikgoz (approved):
Looks good, only one suggestion on improving the sentence.

cwarnermm (approved):

PR #8702, opened 2026-01-30
Fix fragment scroll margin

cpoile:
#### Summary
- The scroll margin was not being set properly in our global css

#### Ticket Link
- none, this was bugging me


lieut-data (approved):

cwarnermm:
While the scroll issue is indeed fixed (thanks, @cpoile!), the right-hand page-level navigation pane isn't accurate:
Screenshot 2026-01-30 at 1 56 18 PM

I've historically thought of the scroll and nav issue as one and the same root cause (Furo bug). Fixing it via Furo rather than forking Furo is preferred.

cwarnermm (dismissed):
See comments

Combs7th:
@asaadmahmood - Would you be able to help with a review on this one? Thanks so much in advance!

Combs7th (dismissed):

asaadmahmood (approved):

Combs7th:
@esethna - This one should be good to go.

esethna:
@cpoile just circling back on this. It has been a while since this was opened, can you help A) summarize the issues this is solving, and B) help us understand what state this PR is in (based on the last comment it seems there was still more work to do), and C) if it's ready to merge, can you help identify what areas of the docs we should pay spacial attention to when testing these changes to ensure there are no regressions? Thanks!

cc// @Combs7th

cpoile:
@esethna The current state is:
1. this code should be refactored -- the header and notification layout needs to be correctly designed, right now its full of hacks and magic numbers.
1. it's split between CSS and SCSS files that are kept manually up to date -- the css should be generated from the SCSS in CI.

But it's been that way for awhile, so pushing this won't make things worse (it improves things a bit). 

Last time I checked it worked on the few pages I tried, but I would feel more confident if someone who know what it should look like went through and checked (with the banners we use on and off, and across desktop sizes and mobile). 

The changes affect 1) the notification and headers layout, 2) the location we scroll to when we click an anchor (hash) link from outside and land on the page (it should scroll right to the section, but not past it, which is what we fixed here, 3) the sidebar table of contents should correctly highlight the section you're on (that was a bit broken so that it only showed your section /after/ the  section header had gone above your viewport off screen).

Hope that helps!

esethna:
Thanks for the info! @Combs7th can you help do some testing in the preview environment https://github.com/mattermost/docs/pull/8702#issuecomment-4344003121) around these areas identified above to make sure things are functioning as expected and there's nothing clearly broken?

Combs7th:
I spent an hour today testing the latest PR preview across regular desktop, and narrower desktop/mobile widths.

Checklist of things I tested:

* Clicking around on various right-side TOC links
* Copy/pasting direct hash/anchor links in a fresh tab
* Checking to make sure banners behave correctly
* And then I checked them all again on narrow desktop & mobile window views

Everything looks good from my preview pass. Anchor links land with the section headings where they should, and banners all behave like normal.

@esethna - Assuming the remaining review/check state is clean, I’m comfortable with this moving forward. There are a few notes from Coderabbitt still, but if I'm being honest I don't understand a lick of it.

Combs7th (approved):

rafaumeu:
Nice approach @cpoile! Reviewed the diff and this is the right fix — moving the notification bar inside `<header>` so Furo's scrollspy picks up the full height natively is much cleaner than maintaining manual scroll-margin-top overrides or bumping the entire theme version. I had opened #9074 to upgrade Furo but closed it in favor of this PR.

One thing worth double-checking: on mobile (max-width: 770px) the removed `header { top: 45px }` rule was previously offsetting the sticky header to account for the notification bar. Since the bar is now inside the header this should be fine, but worth confirming there's no double-offset or gap at the top on mobile.

esethna:
Thanks @rafaumeu for reviewing and @Combs7th for the testing. I will move forward with merging this and keep an eye out for any issues in the live docs once it's pushed to prod

PR #8741, opened 2026-02-13
update slack migration docs

vish9812:
- Updated Slack export terminology from "Regular export"/"Corporate export" to current Slack terms ("Public channels export"/"All channels and conversations export") and corrected the UI navigation path
- Removed the slack-advanced-exporter section and all references — Slack exports already include emails and file download links, making this tool unnecessary
- Removed the Additional Tools subsection (mm-emoji, mmimportjs, slack-migrate-pinned-posts) as they are archived, broken, or no longer maintained
- Added mmetl check slack validation step before running the transform
- Added documentation for useful mmetl transform flags: --skip-attachments, --allow-download, --default-email-domain, --skip-empty-emails, --discard-invalid-props
- Added mmctl import validate step before processing the import
- Fixed incorrect command mmctl import list process to mmctl import job list
- Fixed typos and minor errors throughout the document

PS: I've not tested it with the Enterprise export. So, no changes around that.


wiggin77 (approved):
LGTM 👍   @fmartingr should also have a look before merging.

vish9812:
> slack-advanced-importer built in into mmetl

@fmartingr  From what I've observed, it's not built-in into mmetl but it's actually not needed. Since, Slack regular export, at least now, provides the required data instead of relying on this extra tool to fetch that.

PR #8866, opened 2026-04-08
Split OTel Collector config into separate Mattermost and PostgreSQL files

sadohert:
## Summary

- Splits the combined `otel-collector-config.yaml` into two separate files: `otel-collector-config-mattermost.yaml` and `otel-collector-config-postgres.yaml`
- Step 3 (Install OTel Collector) now references the Mattermost-specific config as the primary download
- The Optional PostgreSQL section now has a distinct callout with its own download link for the PostgreSQL config
- Removes the PostgreSQL group permission command from Step 3 (now belongs in the optional section)

**Motivation:** The combined config caused confusion during customer engagements — users setting up Mattermost log collection were confused by Postgres configuration they didn't need, and vice versa.

## Test plan

- [ ] Verify both YAML download links resolve correctly on the rendered docs page
- [ ] Confirm Step 3 flow reads cleanly for a Mattermost-only deployment
- [ ] Confirm the Optional PostgreSQL callout is clearly separated and self-contained


sadohert:
@kwiersgalla  can you help with adding reviewers to this PR?  This was feedback that came out of a customer PS engagement.

esethna (approved):
Thanks @sadohert this looks file from an editor perspective. Do we need dev review here or have these config files been validated?

sadohert:
I've validated in my own lab setups.  I think someone else might need a bit of setup time to validate this.  I would say risks if there are errors is low so I'd merge.

PR #8879, opened 2026-04-12
Rework Calls Deployment Guide

esethna:
## Summary
- Reworks the Calls deployment doc from a reference-style overview into a step-by-step deployment guide for system administrators.
- Introduces a phased deployment flow covering preparation and networking, integrated setup, RTCD setup, recording setup, pilot rollout, and production rollout.
- Adds clearer decision guidance for deployment architecture, including when to use Integrated mode, RTCD, recording, STUN, and TURN.
- Adds explicit port tables, network validation checks, verification gates, rollback guidance, troubleshooting, and rollout communication templates.


esethna:
@bgardner8008 @sadohert looping you guys in for review on this PR now that we've done a few iterations on it.

There are a lot of file changes (mainly for navigation stuff), but the major changes are in this file:
`source/administration-guide/configure/calls-deployment-guide.md`

bgardner8008 (changes requested):
Overall this is great work! I'm going to summarize my comments here, rather than use inline comments on the PR. Also I know Ian wanted to redo the integrated/RTCD choices, I think this doc works fine as is, and not worth a lot of effort as this choice is going away with Calls v2.

I think the biggest problem is the UDP testing, which needs some changes, see below.

Also, I've recently added support for using private/self-signed certificates with the offloader/recorder/transcriber for air-gapped deployments, but it hasn't made it to the doc yet. Do we want to add it now, or follow up with another PR? I have a markdown doc I will forward to you.

My comments follow:


## Phase 1: Preparation and Networking

1.1 Prerequisites section seems repetitive here, as it was just discussed in previous section, but maybe this is to follow the exact programmatic nature of the doc. If so, then OK.

1.3.2 TURN Server

These conditions should state that connections can't be made *directly* from client to RTCD or calls plugin.

In practice, I'm not aware of any Calls deplyoment that uses TURN, so not worth tons of detail.

"note that NGINX cannot forward UDP traffic". NGINX can forward UDP, but it shouldn't be used because it will break WebRTC peerconnections negotiated using ICE protocol. So better to say "NGINX should not be setup to forward UDP traffic" or something like that.

## Section 1.6

The intent is right, but as written the UDP checks (1.6.1, 1.6.3, 1.6.7) have near-zero diagnostic value in cloud deployments. Pre-install there's no service listening, so the legend's pass states reduce to `closed` for both UDP and TCP. For UDP, `closed` requires the kernel's ICMP port-unreachable to come back to the scanner — and AWS/GCP/Azure Security Groups silently drop disallowed traffic without ICMP, while Linux rate-limits ICMP unreachables by default. Result: a correctly-opened firewall and a misconfigured one both produce `open|filtered`, which the legend doesn't even cover.

TCP checks aren't affected the same way (RST rides the tracked TCP flow, no ICMP dependency), but the ncat-listener pattern below makes them uniform with UDP and additionally validates that a process can bind the port.

Per the section's own framing — "test firewall rules and network reachability only" — the cleanest fix is to spin up a temporary listener on the target host before each scan, so the test is decoupled from whether the real service is installed yet.

**For UDP rows (1.6.1, 1.6.3, 1.6.7):**

On the server:
```
sudo ncat -u -l -k -p 8443 -c '/bin/cat'
```

On the client:
```
sudo nmap -sU -p 8443 TARGET_IP
```

**For TCP rows (1.6.2, 1.6.4, 1.6.5, 1.6.6, 1.6.8, 1.6.9, 1.6.11):**

On the server (substitute the relevant port):
```
sudo ncat -l -k -p <PORT>
```

On the client:
```
nmap -p <PORT> TARGET_IP
```

**For TURN rows (1.6.10, 1.6.13):** these hit coturn, which does respond to anonymous STUN binding requests once installed. Either keep them as-is and reorder them after Phase 2 TURN install, or use the ncat-echo pattern if they need to stay pre-install.

Pass criterion in all cases: **`open`**. Anything else is fail. Stop the listener (Ctrl-C) before installing the real service so it can bind the same port.

## RTCD setup, section 2, test UDP.

I notice this uses nc (different than nmap/ncat), probably OK. But below it should say to enter message on client first:

On the client, type hello and hit Enter. Success: the word hello appears on the server's terminal within ~1 second. To verify the return path, type a message on the server and hit Enter — it should appear on the client.

sadohert:
I had a read-through from a structure standpoint.  I like the numbered execution steps with clear instructions/command references.  I didn't reason about/validate all the commands, but they look sane (not-wisthstanding what Bill said above).

bgardner8008 (approved):
Hi Eric, the changes look great, thanks. Plus I never mentioned that I really like the fact that the Calls deployment has been moved out into a separate section in Deployment guide (it was hidden before).

esethna:
@Combs7th can you help prioritize an editor review on this PR? Specifically, you'll want to review the `source/administration-guide/configure/calls-deployment-guide.md` file, which is live in the preview environment here: http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/8879/administration-guide/configure/calls-deployment-guide.html

Combs7th (approved):

PR #8913, opened 2026-04-20
[MM-67856] docs: add /mobile-logs slash command

Willyfrog:
#### Summary
Document the new /mobile-logs command introduced in mattermost/mattermost#35658, which toggles the attach_app_logs preference to enable attaching mobile app logs as a file in the Mattermost mobile client (v2.38+) for debugging.

#### Ticket Link

[MM-67856]

[MM-67856]: https://mattermost.atlassian.net/browse/MM-67856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/source/integrations-guide/built-in-slash-commands.rst
+++ b/source/integrations-guide/built-in-slash-commands.rst
@@ -56,6 +56,24 @@ Manage channels
 - Edit the channel header using ``/header {text}`` or the channel purpose using ``/purpose {text}``.
 - Rename a channel using ``/rename {text}``.
 
+Manage mobile app log attachments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Use ``/mobile-logs`` to enable an attachment option in the Mattermost mobile client that lets users attach their mobile app logs as a file to the message being sent. This helps administrators and support engineers debug mobile-specific issues by providing device-side context that isn't available from server logs alone. The command updates the ``attach_app_logs`` advanced preference and always responds with an ephemeral message visible only to the invoking user.
+
+.. note::
+
+    This command requires Mattermost mobile app v2.38 or later. On earlier versions the preference can still be set, but the attachment option won't appear in the mobile client.
+
+- Enable log attachment for yourself using ``/mobile-logs on``.
+- Disable log attachment for yourself using ``/mobile-logs off``.
+- Check your current setting using ``/mobile-logs status``.
+- System admins can manage the setting for another user by appending a username, such as ``/mobile-logs on @username``, ``/mobile-logs off @username``, or ``/mobile-logs status @username``.
+
+.. note::
+
+    Non-admin users can only manage their own preference. Attempts to target another account return a neutral *Unable to change mobile log settings for that user* message to avoid username enumeration. Preference changes made through this command are recorded in the audit log.
+
 More useful slash commands
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


BenCookie95 (approved):

source/integrations-guide/built-in-slash-commands.rst
```diff
@@ -56,6 +56,24 @@ Manage channels
 - Edit the channel header using ``/header {text}`` or the channel purpose using ``/purpose {text}``.
 - Rename a channel using ``/rename {text}``.
 
+Manage mobile app log attachments
```
esethna: @Combs7th have you reviewed this? If so are you clear on how it works? 

As written I'm personally unclear on how it works. Let's think about this through the lens of Novice Nate. See below:

esethna: Is this the right place for this whole addition? If we think about the flow when a user would hit this or where this might be useful to an end user, it seems like it should belong in the mobile app troubleshooting section https://docs.mattermost.com/deployment-guide/mobile/mobile-troubleshooting.html, rather than a generic slash commands doc in the integration guide.

source/integrations-guide/built-in-slash-commands.rst
```diff
@@ -56,6 +56,24 @@ Manage channels
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Use ``/mobile-logs`` to enable an attachment option in the Mattermost mobile client that lets users attach their mobile app logs as a file to any message they send. This helps administrators and support engineers debug mobile-specific issues by providing device-side context that isn't available from server logs alone. The command updates the ``attach_app_logs`` advanced preference and always responds with an ephemeral message visible only to the invoking user.
+
+.. note::
+
+    This command requires Mattermost mobile app v2.38 or later. On earlier versions the preference can still be set, but the attachment option won't appear in the mobile client.
```
esethna: Why is this relevant to an end user? If it can't generate the logs then what's the purpose of changing the preference? Seems like we can just remove this and start the whole block with "On Mattermst mobile app v2.38 and later, ..."

source/integrations-guide/built-in-slash-commands.rst
```diff
@@ -56,6 +56,24 @@ Manage channels
...
+
+- Enable log attachment for yourself using ``/mobile-logs on``.
```
esethna: Do you need to use the "/mobile-logs on" command before you can output logs as an attachment? What is the expected output?  Why would/should a user run "/mobile-logs status"?

esethna (changes requested):

source/integrations-guide/built-in-slash-commands.rst
Willyfrog: either the slash command or the toggle in the report an error page needs to be turned on, otherwise the option doesn't show.

The slash command is a shortcut QoL kind of thing. the setting can be turned on and off from the report a problem screen in the mobile app. But navigating there takes a bit and moves you out of the thread. With the slash command we can afffect the setting without leaving the troubleshooting conversation. either by the user or the admin.

So a user can toggle it by either using the slash command or the toggle in that screen.

status was just a way to tell the state without having to navigate, again a simple shortcut to prevent leaving the conversation.

source/integrations-guide/built-in-slash-commands.rst
Willyfrog: I can simplify into "requires v2.38"

source/integrations-guide/built-in-slash-commands.rst
Willyfrog: moved

Willyfrog:
I tried to improve the wording as well as move it to the suggested place @esethna

Willyfrog:
@esethna gentle reminder about this

esethna (approved):

esethna:
Needs to be rebased to 11.8 when it's cut in the next week or so

PR #8914, opened 2026-04-20
Enhance SAML configuration instructions for Mattermost

svelle:
Updated instructions for configuring SAML attributes and claims in Entra for Mattermost integration, including detailed explanations of required and additional claims.


esethna (changes requested):
@svelle for readability of the document, wondering if we should move this content to an appendix, then just add a prerequisit about generating the key and cert, with reference link to the appendix?

esethna:
@Combs7th can you please give this a first pass review?

Combs7th (changes requested):
Thanks, Eric! I finally spent some time giving this a first-pass review.

@svelle, the Entra updates look useful overall. I agree the full certificate script should be removed in favor of linking to the existing `gencert.sh` docs.

When you have time, please also address the remaining security comments, confirm the UI labels, and very the technical guidance and screenshots. Then re-request review and I can take another look.

svelle:
@esethna @Combs7th removed the certificate guidance and addressed coderabbit review. good to merge now.

Combs7th:
You rock, Sven! It's looking much better after that round of updates.

From a Novice Nate perspective, my only lingering concern is whether the cert generation section assumes a bit too much prior OpenSSL/Linux knowledge, but that's also coming from someone who's not super familiar with generating certs myself.

Otherwise, it looks good to me. I'll approve from my side unless @esethna still has any concerns.

svelle:
Thanks @Combs7th 
Yeah I think those cert instructions are due for a different PR. 
In general they're fairly simple to follow for a somewhat knowledgeable sysadmin.

Combs7th (approved):

PR #8916, opened 2026-04-23
Fix changelog automation workflow

amyblais:
Fixed the changelog automation workflow by creating two dedicated GitHub Apps. Amy is reaching out to IT to help with remaining variables/secrets.

References:
 - Spec doc https://mattermost.atlassian.net/wiki/spaces/Security/pages/4449992717/Automation+-+Changelog+Creation

### DIFF
--- a/.github/workflows/generate-changelog.yml
+++ b/.github/workflows/generate-changelog.yml
@@ -29,31 +29,36 @@ jobs:
       pull-requests: write
  
     steps:
-      # App 1: Read-only access to mattermost/mattermost and mattermost/enterprise.
-      # Used by the Python script to fetch merged PRs and their release notes.
-      # Credentials needed in Settings → Secrets and variables → Actions:
-      #   CHANGELOG_READ_APP_ID      (variable)  — App ID from the mattermost+enterprise GitHub App
-      #   CHANGELOG_READ_PRIVATE_KEY (secret)    — Private key from the mattermost+enterprise GitHub App
+      # Single GitHub App with access to mattermost/mattermost, mattermost/enterprise, and mattermost/docs.
+      # Two tokens are generated with different repo scopes to preserve least privilege.
+      # Credentials are stored as org-level secrets/variables in the mattermost organization,
+      # with access explicitly granted to mattermost/docs.
+      #
+      # Required org-level secrets/variables (Organization Settings → Secrets and variables → Actions):
+      #   CHANGELOG_APP_ID      (org variable) — Client ID of the GitHub App
+      #   CHANGELOG_PRIVATE_KEY (org secret)   — Private key of the GitHub App
+ 
+      # Read-only token scoped to the source repos only.
+      # Used by the Python script to fetch merged PRs and release notes.
       - name: Get read token (source repos)
-        uses: actions/create-github-app-token@v1
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
         id: readToken
         with:
-          app-id: ${{ vars.CHANGELOG_READ_APP_ID }}
-          private-key: ${{ secrets.CHANGELOG_READ_PRIVATE_KEY }}
+          app-id: ${{ vars.CHANGELOG_APP_ID }}
+          private-key: ${{ secrets.CHANGELOG_PRIVATE_KEY }}
           owner: mattermost
+          repositories: mattermost,enterprise
  
-      # App 2: Read/write access to mattermost/docs.
+      # Write token scoped to the docs repo only.
       # Used to check out this repo, push the changelog branch, and open the PR.
-      # Credentials needed in Settings → Secrets and variables → Actions:
-      #   CHANGELOG_WRITE_APP_ID      (variable) — App ID from the docs GitHub App
-      #   CHANGELOG_WRITE_PRIVATE_KEY (secret)   — Private key from the docs GitHub App
       - name: Get write token (docs repo)
-        uses: actions/create-github-app-token@v1
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
         id: writeToken
         with:
-          app-id: ${{ vars.CHANGELOG_WRITE_APP_ID }}
-          private-key: ${{ secrets.CHANGELOG_WRITE_PRIVATE_KEY }}
+          app-id: ${{ vars.CHANGELOG_APP_ID }}
+          private-key: ${{ secrets.CHANGELOG_PRIVATE_KEY }}
           owner: mattermost
+          repositories: docs
  
       - name: Checkout docs repo
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-changelog.yml
```diff
@@ -29,21 +29,38 @@ jobs:
...
+      - name: Get write token (docs repo)
+        uses: mattermost/github-app-installation-token-action@4d4c8c09a49df5b54e1eafd372d1a1e44fdcab13
+        id: writeToken
+        with:
+          appId: "${{ vars.CHANGELOG_WRITE_APP_ID }}"
+          installationId: "${{ vars.CHANGELOG_WRITE_INSTALL_ID }}"
+          privateKey: ${{ secrets.CHANGELOG_WRITE_PRIVATE_KEY }}
  
       - name: Checkout docs repo
         uses: actions/checkout@v4
```
NARSimoes: Unrelated with this pull-request but we should pin this to commit sha, can be addressed in follow up.

NARSimoes (approved):
Thanks @amyblais!

.github/workflows/generate-changelog.yml
amyblais: Updated!

.github/workflows/generate-changelog.yml
```diff
@@ -29,24 +29,41 @@ jobs:
       pull-requests: write
  
     steps:
-      # A GitHub App token is required so the workflow can read PRs from the
-      # server repo AND push/create PRs in this docs repo.
-      # The default GITHUB_TOKEN only works within the repo the workflow runs in.
-      - name: Get GitHub App token
-        uses: mattermost/github-app-installation-token-action@4d4c8c09a49df5b54e1eafd372d1a1e44fdcab13
-        id: appToken
```
esarafianou, line 37: Why do we prefer creating 2 GH apps and move away from using the github-app-installation-token action?

.github/workflows/generate-changelog.yml
```diff
@@ -29,24 +29,41 @@ jobs:
...
+      # App 1: Read-only access to mattermost/mattermost and mattermost/enterprise.
+      # Used by the Python script to fetch merged PRs and their release notes.
+      # Credentials needed in Settings → Secrets and variables → Actions:
+      #   CHANGELOG_READ_CLIENT_ID   (variable)  — Client ID from the mattermost+enterprise GitHub App
+      #   CHANGELOG_READ_PRIVATE_KEY (secret)    — Private key from the mattermost+enterprise GitHub App
+      - name: Get read token (source repos)
+        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2.1.0
```
esarafianou: Similar for here and line 52: https://github.com/tibdex/github-app-token is marked as archived/deprecated. If we'd like to move away from the `mattermost/github-app-installation-token-action`, we can look at https://github.com/actions/create-github-app-token

esarafianou (changes requested):

.github/workflows/generate-changelog.yml
amyblais: The workflow now uses ``actions/create-github-app-token@v1``, which is the official GitHub-maintained action and not archived. It also auto-discovers the installation by org (owner: mattermost), so no Installation ID is needed.

.github/workflows/generate-changelog.yml
amyblais: The first test run failed: https://github.com/mattermost/docs/actions/runs/24129744941. We are using ``UNIFIED_CI_PRIVATE_KEY`` in public / private repos and Nuno wasn't sure if we should keep spreading that. I worked with Claude to come up with another solution.

(1) The original ``mattermost/github-app-installation-token-action`` requires an ``installationId``, which we don't have readily available — IT only provided the Client ID and private key when setting up the apps. 
(2) Separate apps were created intentionally to follow least-privilege: one is read-only on ``mattermost/mattermost`` and ``mattermost/enterprise``, and the other is read/write on ``mattermost/docs`` only. 

Happy to revisit if there's a preferred pattern for cross-repo workflows in this org.

esarafianou: I'm in favor of the least privilege approach. My only worry with the new approach is that it doesn't lend itself well to scaling for many workflows. Nothing to worry about here; more as a general discussion.

.github/workflows/generate-changelog.yml
```diff
@@ -29,24 +29,39 @@ jobs:
...
-      - name: Get GitHub App token
-        uses: mattermost/github-app-installation-token-action@4d4c8c09a49df5b54e1eafd372d1a1e44fdcab13
-        id: appToken
+      # App 1: Read-only access to mattermost/mattermost and mattermost/enterprise.
+      # Used by the Python script to fetch merged PRs and their release notes.
+      # Credentials needed in Settings → Secrets and variables → Actions:
+      #   CHANGELOG_READ_APP_ID      (variable)  — App ID from the mattermost+enterprise GitHub App
+      #   CHANGELOG_READ_PRIVATE_KEY (secret)    — Private key from the mattermost+enterprise GitHub App
+      - name: Get read token (source repos)
+        uses: actions/create-github-app-token@v1
```
esarafianou: could we ping to a specific hash? (similarly below)

```suggestion
        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3  # v3.1.1
```

esarafianou (changes requested):

.github/workflows/generate-changelog.yml
amyblais: @esarafianou Would the below proposal be a cleaner approach to address "My only worry with the new approach is that it doesn't lend itself well to scaling for many workflows", or do you have any other ideas?

Proposed new approach:
1. One GitHub App with read access to mattermost/mattermost and mattermost/enterprise, and read/write to mattermost/docs.
2. CHANGELOG_APP_ID stored as an org-level variable, accessible to mattermost/docs.
3. CHANGELOG_PRIVATE_KEY stored as an org-level secret, accessible to mattermost/docs.
4. Workflow generates two scoped tokens as shown above.

.github/workflows/generate-changelog.yml
amyblais: Addressed 👍

.github/workflows/generate-changelog.yml
esarafianou: > Proposed new approach

This is the approach already implemented in the PR, correct? 

That works well for this case. We can think in the future, how to structure GH Apps for least privilege. An idea could be 1 GH App for read access to X repos and one 1 GH App for read/write access to Y repos (what the proposed new approach is), or make it further fine grained to differentiate between private and public repos. 

cc. @NARSimoes

amyblais: The new approach hasn't been implemented here yet (I accidentally implemented it but I've reverted it for now since the discussion is ongoing).

I will let Nuno take another look at the current approach with the two GitHub apps, and let him share feedback if we'd like to use another approach.

NARSimoes: Yes, the idea in this case was to use the least privilege using these new apps. Please find more details in https://mattermost.atlassian.net/browse/SEC-10047

> My only worry with the new approach is that it doesn't lend itself well to scaling for many workflows. Nothing to worry about here; more as a general discussion.
> We can think in the future, how to structure GH Apps for least privilege.

💯

Yes, we need to define / discuss how we'll scale this (goal of the ticket mentioned above).

esarafianou (approved):

NARSimoes (approved):

PR #8919, opened 2026-04-27
Add PostgreSQL HA cluster guide

sadohert:
## Summary

- Adds `source/administration-guide/scale/postgres-ha-cluster.rst` — a new guide for deploying a 3-node PostgreSQL HA cluster using repmgr, HAProxy, and Keepalived
- Updates `scaling-for-enterprise.rst` toctree and adds a prose entry
- Updates `high-availability-cluster-based-deployment.rst` to cross-link to the new page

## Why

The existing HA doc covers Mattermost app-layer clustering but explicitly does not cover database HA. Many self-hosted customers on bare-metal or VMs need a database-level HA guide. This fills that gap.

## Page structure

The guide follows Mattermost documentation conventions:
- Requirements consolidated upfront (Before you begin section)
- Decision guidance table to help admins choose the right architecture
- 5 numbered phases with explicit pass/fail checkpoint commands
- Day-2 operations and troubleshooting sections

## Validation

All setup steps and checkpoint commands have been validated on Ubuntu 24.04 LTS, PostgreSQL 17, repmgr 5.5, HAProxy 2.8.

## Related

- Companion Multi-DC disaster recovery guide planned as follow-up PR once DC2 failover testing is complete


sadohert:
@coderabbitai review

sadohert:
@esethna @wiersgallak - Would one of you mind adding the `preview-environment` label to this?  I'd like to share wiht a customer for early preview.

esethna:
Done @sadohert

Combs7th:
Heya @sadohert - Is there you a dev you'd recommend we tag on this for the technical review?

sadohert:
Hey @Combs7th -  You could raise with Jesse, Alejandro, or Doug?

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+routing, and `Keepalived <https://keepalived.org/>`__ for Virtual IP (VIP)
+management.
+
+This is infrastructure-level HA that operates independently of your Mattermost
+edition. It is compatible with any self-hosted Mattermost deployment.
+
+.. note::
+
+   This guide has been validated on: **Ubuntu 24.04 LTS**, **PostgreSQL 17**,
+   **repmgr 5.5**, **HAProxy 2.8**, **Keepalived**.
```
agarciamontoro: Are we missing the version in Keepalived?

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+
+Architecture overview
+---------------------
+
+A PostgreSQL HA cluster for Mattermost consists of three nodes running in
+parallel. Each node runs the full stack: PostgreSQL, repmgr daemon (repmgrd),
+HAProxy, Keepalived, and a health-check service. A Virtual IP (VIP) floats
+across nodes and always points to the current primary.
```
agarciamontoro, line 26: Does it *need* to be exactly three?

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+     - TCP load balancer. Routes write traffic to the primary and read traffic
+       to standbys via two ports.
+   * - Keepalived
+     - —
+     - Manages the VIP using VRRP. Moves the VIP to the new primary after
+       failover.
+   * - pgchk.py
+     - —
+     - HTTP health-check endpoint (port 8008). HAProxy queries this to
+       determine which node is the current primary.
```
agarciamontoro, line 80: There are no versions here either, not sure if intended, raising just in case.

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+
+**Step 1.1 — Configure /etc/hosts**
+
+On each node, append to ``/etc/hosts``:
+
+.. code-block:: text
+
+   <PG1_IP>  pg1
+   <PG2_IP>  pg2
+   <PG3_IP>  pg3
```
agarciamontoro, line 237: Maybe a dumb question: should each node use `localhost` for its own entry?

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+   host    replication repmgr      <SUBNET>/24     scram-sha-256
+   host    replication repmgr      127.0.0.1/32    scram-sha-256
+
+Create a ``.pgpass`` file on each node so repmgr can authenticate without
+an interactive password prompt:
+
+.. code-block:: bash
+
+   echo "*:*:repmgr:repmgr:<YOUR_REPMGR_PASSWORD>" >> ~/.pgpass
+   chmod 600 ~/.pgpass
```
agarciamontoro, line 322: This reads a bit weird, because it seems we have already come up with a string for `<YOUR_REPMGR_PASSWORD>` before. However, we're deciding the password at this point, and the actual setting of the password happens later when we run `ALTER USER repmgr PASSWORD '<YOUR_REPMGR_PASSWORD>'`. It took me a couple of seconds to understand this, so maybe this warrants some short explanation here? Or maybe configuring this after we run the `ALTER USER`? Not sure.

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+
+   sudo -u postgres createuser --superuser repmgr
+   sudo -u postgres createdb --owner=repmgr repmgr
+   sudo -u postgres psql -c "ALTER USER repmgr SET search_path TO repmgr, public;"
+   sudo -u postgres psql -c "ALTER USER repmgr PASSWORD '<YOUR_REPMGR_PASSWORD>';"
+
+**Step 3.2 — Create /etc/repmgr.conf (all nodes)**
+
+Create ``/etc/repmgr.conf`` on each node. Adjust ``node_id``, ``node_name``,
+and ``host`` for each node:
```
agarciamontoro: I first looked for `host` setting in the file, but it's inside the `conninfo` setting, so maybe we can say something like
```suggestion
Create ``/etc/repmgr.conf`` on each node. Adjust ``node_id``, ``node_name``,
and ``host`` (in ``conninfo``) for each node:
```

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+   [Install]
+   WantedBy=multi-user.target
+
+.. code-block:: bash
+
+   sudo systemctl daemon-reload
+   sudo systemctl enable repmgrd
+   sudo systemctl start repmgrd
+
+✅ **Phase 3 checkpoint** — run on any node:
```
agarciamontoro, line 452: This may be slightly more clear?
```suggestion
✅ **Phase 3 checkpoint** — run on every node:
```

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+       server pg1 <PG1_IP>:5432 check port 8008
+       server pg2 <PG2_IP>:5432 check port 8008 backup
+       server pg3 <PG3_IP>:5432 check port 8008 backup
+
+   backend pg_replicas
+       balance roundrobin
+       option tcp-check
+       server pg2 <PG2_IP>:5432 check port 8008
+       server pg3 <PG3_IP>:5432 check port 8008
+       server pg1 <PG1_IP>:5432 check port 8008 backup
```
agarciamontoro: I guess the order in each of these blocks is important. Should we make a note about that? Or would that be too much?

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+           httpd.serve_forever()
+       except KeyboardInterrupt:
+           pass
+       httpd.server_close()
+
+   if __name__ == '__main__':
+       parser = argparse.ArgumentParser(description='PostgreSQL Health Check for HAProxy')
+       parser.add_argument('--port', type=int, default=DEFAULT_PORT)
+       args = parser.parse_args()
+       run(port=args.port)
```
agarciamontoro, line 618: Should we host this in a repo or in a gist so that we can track history here? Just an idea.

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+                   self.safe_write(b"OK - Replica\n")
+               else:
+                   self.send_response(503); self.end_headers()
+                   self.safe_write(b"Service Unavailable - Not Replica\n")
+           else:
+               self.send_response(404); self.end_headers()
+               self.safe_write(b"Not Found\n")
+
+       def log_message(self, format, *args):
+           pass
```
agarciamontoro, line 603: This function is not used, we should remove it.

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+           <CLUSTER_VIP>/24
+       }
+   }
+
+.. code-block:: bash
+
+   sudo systemctl enable keepalived
+   sudo systemctl start keepalived
+
+✅ **Phase 4 checkpoint** — run on any node:
```
agarciamontoro, line 687: every again? Maybe this is just me, sorry 😂 
```suggestion
✅ **Phase 4 checkpoint** — run on every node:
```

source/administration-guide/scale/postgres-ha-cluster.rst
```diff
@@ -0,0 +1,902 @@
...
+   .. code-block:: bash
+
+      sudo systemctl stop postgresql
+      sudo -u postgres repmgr -h <PRIMARY_IP> -U repmgr -d repmgr \
+          -f /etc/repmgr.conf standby clone --delete-existing-pgdata
+      sudo systemctl start postgresql
+      sudo -u postgres repmgr -f /etc/repmgr.conf standby register
+
+4. Add the new node to ``/etc/haproxy/haproxy.cfg`` on all existing nodes and
+   reload HAProxy: ``sudo systemctl reload haproxy``.
```
agarciamontoro, line 834: Do we need to change the `/etc/hosts` file of the pre-existing nodes to add the new `pg4` with its IP?

agarciamontoro (changes requested):
Amazing guide! The only thing that I though "huh, weird", is the need for the Python script. Isn't there a way to automate that so that we can avoid using that mini-server? If not, it's ok, but I'm curious.

My comments are mostly nits, so treat them like that, but I wanted to make sure I understood everything :)

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Kept as "any node" intentionally — `repmgr cluster show` reports the whole cluster's state from whichever node you run it on, so running it once on any single node is enough for this checkpoint. (The Phase 1 and 2 checkpoints say "every node" because those verify node-local service state.)

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Same reasoning as the Phase 3 checkpoint — these commands report cluster-wide/VIP-level state, so running them once on any node is sufficient. Kept "any node" here; the per-node service checks in Phases 1–2 use "every node".

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Good catch — this pointed at a real issue. The config was using `option tcp-check`, which only confirms port 8008 is reachable. Since `pgchk.py` listens on 8008 on every node regardless of role, that check passed everywhere, so routing depended on the static server order rather than on which node is actually the primary — and it would not follow the primary after a repmgr failover.

I've switched both backends to `option httpchk GET /master` / `GET /replica` with `http-check expect status 200`, so HAProxy now routes on the health-check result (only the primary answers 200 on `/master`, only standbys on `/replica`). Order is no longer significant, and the now-redundant `backup` markers are removed. Fixed in 8d86f14b2.

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: We're keeping `pgchk.py` embedded inline on purpose — it keeps the script versioned alongside the docs with no external dependency to fetch or trust at deploy time. (This also resolved a separate review note about an external repo link that no longer existed.) Happy to revisit hosting it separately if the script grows.

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Good catch — added **Keepalived 2.2.8** here (the version that ships with Ubuntu 24.04 LTS). Fixed in 910be719d.

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Filled in the Keepalived version (2.2.8) in the table. I left pgchk.py as `—` since it's the small script bundled in this guide rather than a packaged component with its own version. Fixed in 910be719d.

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Added a note clarifying this: three is the recommended *minimum* (so a primary + at least one standby survive a single-node failure), you can scale out with more standbys, and an odd node count is preferred. Fixed in 910be719d.

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Not a dumb question — added a note to make it explicit: use the real IP for all three entries on every node, including each node's own entry. `127.0.0.1`/`localhost` shouldn't be substituted because repmgr and streaming replication advertise these hostnames to the other nodes and need them to resolve to routable addresses. Fixed in 910be719d.

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Agreed it reads ahead of itself — added a note here explaining that you choose the password now and reuse the same value for every `<YOUR_REPMGR_PASSWORD>`, but it's only actually set on the database later in Step 3.1 via `ALTER USER repmgr PASSWORD`. Writing it to `.pgpass` here just pre-stages the credential. Fixed in 910be719d.

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Applied your suggestion — now reads `host` (in `conninfo`). Fixed in 910be719d.

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: This one's actually load-bearing: `log_message` is an override of `BaseHTTPRequestHandler.log_message`, which the framework calls on every request to write a log line to stderr. Overriding it with `pass` suppresses that — otherwise each HAProxy health check (every few seconds, per node) would spam the journal. I kept it and added a comment explaining why, so it doesn't read as dead code. Done in 910be719d.

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Yes — added an explicit step for it: when adding a standby, add the new node's hostname/IP to `/etc/hosts` on the new node *and on every existing node* so all nodes can resolve the new hostname. Fixed in 910be719d.

sadohert:
@wiersgallak would you mind adding the preview label here so I can review before re-requesting a review from Alejandro?

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Took this further per your point — added a note to the checkpoint: if `repmgr cluster show` lists **fewer than three nodes**, run it on **every** node and compare (the views can diverge when a node is isolated), and collect all three outputs when raising a support request. So "any node" stays the normal path, but there's now an explicit per-node step when something looks off. Added in aaf9e04da.

source/administration-guide/scale/postgres-ha-cluster.rst
sadohert: Hi @agarciamontoro - Its me :).  That above answer was not.  I will resolve this another way.

esethna:
@sadohert what are next steps on this PR? Can we help move it along at all?

agarciamontoro:
Self-requesting my review! I had forgotten about this PR. Will try to go through it tomorrow.

Combs7th:
@agarciamontoro - Just a gentle reminder that this doc still needs a review whenever you're able. Lemme know if there's anything I can do to help!

agarciamontoro:
@Combs7th: I talked to @sadohert offline and he wanted to give this another pass before my review, sorry I didn't share the update here!

agarciamontoro (changes requested):
Marking this again as Request changes to remove it from my queue until I get @sadohert's ping. Sorry for the noise!

esethna:
@sadohert friendly ping, are you still working on this PR? Thanks!

sadohert:
@esethna  Apologies for the delay.  I'll look into closing this out next week (on vacation this week)

Combs7th:
@sadohert - Just sharing another friendly on this one. :) Thanks in advance!

esethna:
@sadohert please resubmit this against the [/mattermost repository](https://github.com/mattermost/mattermost/tree/master/docs) since docs have now migrated. Thanks!

cc// @Combs7th

PR #8925, opened 2026-04-30
Create release docs automation

amyblais:
Full spec doc: https://mattermost.atlassian.net/wiki/spaces/Security/pages/4518772760/Automation+-+Release+Docs+Update+GitHub+Actions+Workflow.

Adds a GitHub Actions workflow that automates docs updates for Mattermost releases. Instead of manually opening a GitHub issue, tagging claude, and pasting file URLs each time, you go to Actions → Update Docs → Run workflow and fill in four fields:

 - Component — Server, Mobile, or Desktop
 - Release type — ESR, Feature Release, Security Release, Patch/Dot Release, or Other
 - Version — e.g. 11.7, 2.40, 6.2
 - Release date — e.g. May 15, 2026
 - Additional instructions (optional) — any extra context like an ESR end-of-support date

The workflow then automatically selects the right set of files for that component and release type, sends each one to Claude via the Anthropic API with the release context, and opens a PR with all the updated files — no issue, no copy-pasting URLs, no manual file list.


amyblais:
@coderabbitai full review

amyblais:
@coderabbitai resolve

Combs7th (approved):

Combs7th:
@NARSimoes @esarafianou - Are y'all able to help give this a dev review?

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,100 @@
...
+jobs:
+  update-docs:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+ 
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
```
NARSimoes: Should we use a recent version? This commits refers to `v4` and the latest version is now `v6.0.2` 

The comment applies to other actions.

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Normalize version for branch naming
+        id: vars
+        shell: bash
+        run: |
+          safe_version="$(printf '%s' "${{ inputs.version }}" | tr ' /' '--' | tr -cd '[:alnum:]._-')"
+          test -n "$safe_version"
+          echo "safe_version=$safe_version" >> "$GITHUB_OUTPUT"
+ 
+      - name: Create pull request
+        uses: peter-evans/create-pull-request@f230e4b8d3f1f2f2c8f2a9a0d6c1f3f3b7e3d1a0 # v6
```
NARSimoes: Should we use `gh api` like we are doing in other workflows for consistency and avoiding additional third party actions. In case we need the Github App to get we already have that too.

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,100 @@
...
+          - Other
+      version:
+        description: 'Version number (e.g., 11.7, 2.40, 6.2, 5.13.6)'
+        required: true
+        type: string
+      release_date:
+        description: 'Release date (e.g., May 15, 2026)'
+        required: true
+        type: string
+      instructions:
```
NARSimoes: What additional instruction are we planning to add? I wonder if we should strict the input instructions since current approach is a free form and inject anything to the model prompt.

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,100 @@
+name: Update Docs
+ 
+on:
+  workflow_dispatch:
```
NARSimoes, line 7: Before merging double check who can trigger this workflow.

.github/scripts/update_docs.py
```diff
@@ -0,0 +1,237 @@
...
+    print(f"  Version:      {VERSION}")
+    print(f"  Release date: {RELEASE_DATE}")
+    if INSTRUCTIONS:
+        print(f"  Instructions: {INSTRUCTIONS}")
+    print(f"  Files ({len(files)}):")
+    for f in files:
+        print(f"    {f}")
+    print()
+ 
+    client = anthropic.Anthropic(api_key=os.environ["ANTHROPIC_API_KEY"])
```
NARSimoes: Should we add a `timeout` to avoid hung the run in Anthropic calls? https://platform.claude.com/docs/en/api/sdks/python#timeouts

NARSimoes: I think the retries default is `max_retries=2` so I wonder if we should make it explicit.
https://platform.claude.com/docs/en/api/sdks/python#retries

.github/scripts/update_docs.py
```diff
@@ -0,0 +1,237 @@
...
+    if not response.content or response.content[0].type != "text":
+        raise RuntimeError(
+            f"Unexpected API response structure for {filepath}: "
+            f"content={response.content!r}"
+        )
+    if response.stop_reason == "max_tokens":
+        raise RuntimeError(
+            f"Claude hit max_tokens ({max_tokens}) for {filepath}; output is truncated. "
+            "Increase max_tokens or split the file."
+        )
```
NARSimoes: These two conditions seem duplicated with the previous ones.

.github/scripts/update_docs.py
```diff
@@ -0,0 +1,237 @@
...
+        with open(filepath, "r", encoding="utf-8") as f:
+            original = f.read()
+    except FileNotFoundError:
+        print(f"  WARNING: {filepath} not found — skipping.")
+        return
+ 
+    print(f"  Sending to Claude...")
+    response = client.messages.create(
+        model="claude-sonnet-4-6",
+        max_tokens=16000,
```
NARSimoes: Why max tokens = 16000 ? Have you checked If this enough for exiting files / prompts?

.github/scripts/update_docs.py
```diff
@@ -0,0 +1,237 @@
...
+#   - ESR: releases page + changelog + both install guides
+#   - Patch/Dot: releases page + changelog only
+# ---------------------------------------------------------------------------
+ 
+SERVER_FILES = [
+    "source/product-overview/mattermost-v11-changelog.md",
+    "source/product-overview/mattermost-server-releases.md",
+    "source/deployment-guide/server/linux/deploy-rhel.rst",
+    "source/deployment-guide/server/linux/deploy-tar.rst",
+    "source/product-overview/mattermost-desktop-releases.md",
```
NARSimoes, line 59: Any reason to keep this file under `SERVER_FILES`?

NARSimoes (changes requested):
Thanks @amyblais, added a few comments / clarifications.

.github/workflows/update-docs.yml
amyblais: Addressed! Bumped to v6.0.2 and v6.2.0 respectively, both pinned to their exact commit SHAs.

.github/workflows/update-docs.yml
amyblais: Addressed! Replaced entirely with a two-step git commit + push followed by ``gh pr create --body-file``, consistent with how other Mattermost workflows handle PRs. Also added a no-op guard: if the script made no changes, the commit step sets changed=false and the PR step is skipped entirely.

.github/workflows/update-docs.yml
amyblais: Addressed! Updated the description to "Keep concise — appended to the Claude prompt as-is" so it's clear what it does and nudges users toward short, specific input rather than arbitrary text.

.github/workflows/update-docs.yml
amyblais: Addressed! Added a comment at the top of the workflow explaining that ``workflow_dispatch`` is restricted to write-access members by default, with a pointer to environments for tighter control.

.github/scripts/update_docs.py
amyblais: Addressed! Added ``timeout=120.0`` to the Anthropic client.

amyblais: Addressed! Made explicit as ``max_retries=2`` with a comment noting this matches the default.

.github/scripts/update_docs.py
amyblais: Addressed! Added section comments clearly distinguishing the two layers: the API integrity checks (which raise and mark the file as failed) vs. the content quality guards (which return with a warning and leave the file unchanged).

.github/scripts/update_docs.py
amyblais: Addressed! Raised to ``MAX_TOKENS = 32000`` with a comment explaining the reasoning (most files are a few thousand tokens; 32k gives headroom without approaching the 64k model limit).

.github/scripts/update_docs.py
amyblais: Addressed!

amyblais: Addressed! Added an inline comment explaining it was intentionally included because server ESR releases reference compatible desktop versions.

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,132 @@
...
+            echo "- Component: ${COMPONENT}"
+            echo "- Version: v${VERSION}"
+            echo "- Release type: ${RELEASE_TYPE}"
+            echo "- Release date: ${RELEASE_DATE}"
+            [ -n "${INSTRUCTIONS}" ] && echo "- Notes: ${INSTRUCTIONS}"
+            echo ""
+            echo "Please review all changes carefully before merging."
+          } > pr_body.md
+ 
+          gh pr create \
```
NARSimoes: Consider adding `--base`  / `--head` branch. For example in `generate-changelog` we use
```
            --base master \
            --head "changelog/${{ inputs.version }}"
```

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,132 @@
...
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git checkout -b "docs/${{ inputs.component }}-v${{ steps.vars.outputs.safe_version }}"
+          git add source/
+          if git diff --cached --quiet; then
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+            echo "No files were modified — skipping PR creation."
+          else
+            git commit -m "docs: update ${{ inputs.component }} files for v${{ inputs.version }} ${{ inputs.release_type }}"
+            git push origin "docs/${{ inputs.component }}-v${{ steps.vars.outputs.safe_version }}"
```
NARSimoes: We might have some collisions here if we are planning to re-run this for same component / version. For example if the pull requests already exists or content already pushed. One possible option could be `--force-with-release` but would need to double check.
 ```
# Create new branch e.g. docs/server-v17
git checkout -b "docs/${{ inputs.component }}-v${{ steps.vars.outputs.safe_version }}"
...
# Push to e.g. docs/server-v17
git push origin "docs/${{ inputs.component }}-v${{ steps.vars.outputs.safe_version }}"
```

File: .github/scripts/update_docs.py [hunk 11]
```diff
@@ -0,0 +1,243 @@
...
+            errors.append((filepath, f"{type(e).__name__}: {e}"))
+        print()
+ 
+    if errors:
+        print(f"\n{len(errors)} file(s) failed:")
+        for fp, err in errors:
+            print(f"  {fp}: {err}")
+        sys.exit(1)
+    else:
+        print("All files processed successfully.")
```
NARSimoes: Clarification: there are files being skipped just printing messages with `Warnings / Errors` and here it's printing `All files processed successfully.`. Is this expected?

.github/scripts/update_docs.py
```diff
@@ -0,0 +1,243 @@
...
+# ---------------------------------------------------------------------------
+# File lists per component / release type
+#
+# Desktop ESR and Desktop Patch/Dot releases touch different files:
+#   - ESR: releases page + changelog + both install guides
+#   - Patch/Dot: releases page + changelog only
+# ---------------------------------------------------------------------------
+ 
+SERVER_FILES = [
+    "source/product-overview/mattermost-v11-changelog.md",
```
NARSimoes: I wonder if this file is too heavy. If we are only interested in the ~ < 50 % of the file we could find some strategy to reduce the size of (e.g. just send part of the file, etc). Open for discussion.

.github/workflows/update-docs.yml
NARSimoes: Thank you @amyblais, What additional instruction are we planning to add? Just trying to understand if we really need this or we can think about an alternative to avoid having a free form that allow injecting anything to the models.

NARSimoes (changes requested):
Thanks for addressing the review, a few more clarifications.

.github/workflows/update-docs.yml
amyblais: Replaced free-form instructions with a structured ``esr_end_date`` input (covers the known use case without the injection risk).

.github/workflows/update-docs.yml
amyblais: Added ``--base/--head`` to ``gh pr create``.

.github/workflows/update-docs.yml
amyblais: Addressed - using ``git push --force-with-lease`` and check if a PR already exists before creating.

.github/scripts/update_docs.py
amyblais: Refactored ``update_file()`` to return a status, and print an accurate summary that distinguishes ``updated/unchanged/skipped/not-found``.

.github/scripts/update_docs.py
amyblais: Removed from the list as we have a separate automation for the server changelog.

.github/scripts/update_docs.py
```diff
@@ -0,0 +1,288 @@
...
+        with open(filepath, "r", encoding="utf-8") as f:
+            original = f.read()
+    except FileNotFoundError:
+        print(f"  WARNING: {filepath} not found — skipping.")
+        return "not_found"
+ 
+    # Large changelogs only need recent context; new entries go near the top.
+    # Send only the head and reconstruct the full file afterward.
+    truncated = len(original) > MAX_SEND_CHARS
+    send_content = original[:MAX_SEND_CHARS] if truncated else original
```
NARSimoes, line 167: When `truncated=true` we do partial input and merge / append the tail in the code. So, I wonder if we should tell Claude explicitly e.g. “you only see the start; return the full updated start only; we append the rest” to avoid hallucinations or rewrites the missing part. This only happens when truncated is true so we can approach this is in a follow up.

NARSimoes (approved):
Thanks @amyblais, just added a comment with a minor follow up.

PR #8988, opened 2026-05-19
Changelog automation fixes

amyblais:
Minor improvements to the changelog generation automation.

 - "UI" → "user interface" — added as a dedicated Terminology rule in the system prompt.
 - Feature vs ESR releases — new release_type dropdown input (feature/esr). Feature releases generate (release-v11.7)= / ## Release v11.7; ESR releases keep the Extended Support Release link. Defaults to feature.
 - No blank lines between headers and bullets — explicitly stated in the rules.
 - Indented bullets — system prompt now specifies   - item (2-space indent) for all bullet points.
 - "Fixed a bug" → "Fixed an issue" — added to sentence patterns and rules.
 - Go Version section always included — new go_version workflow input. Script builds the section from the input (Updated to ``go1.22.5``.) or falls back to "Go version is the same as in the previous release.". The section is injected by replacing the AI's ### Go Version placeholder, or appended if the AI omitted it.
 - Open Source Components format — system prompt now specifies the exact format.
 - Fixed some cases where release notes may not get pulled from dev PRs: 
     - Heading variation: Added a secondary scan of the entire PR body for ``release-note`` blocks. If no ``#### Release Note`` section is found, the script now picks up any ``release-note`` blocks wherever they appear, matching the format you showed.
    - Null ``merged_at``: For any PR where merged_at is unexpectedly null, the script now makes a direct call to the Pulls API (``/repos/{owner}/{repo}/pulls/{number``}) to verify the merged flag before skipping it.
    - Rate limiting: Replaced all ``requests.get()`` calls with a new ``make_github_request()`` helper that reads the ``X-RateLimit-Reset`` header and sleeps until the window resets before retrying, up to 3 retries.


amyblais:
@coderabbitai full review

.github/scripts/generate_changelog.py
```diff
@@ -35,11 +39,37 @@
 # Prevents the workflow from hanging indefinitely if the API stalls.
 API_TIMEOUT = 30
  
+ 
+def make_github_request(url: str, params: dict = None) -> requests.Response:
```
NARSimoes: I wonder if we should use standard requests / retry logic for retries. I think we can keep similar goals but please double check. For example

```
# Simple example just to explain the idea: must be refined
from requests.adapters import HTTPAdapter
from requests.packages.urllib3.util.retry import Retry

# Retry transient failures and rate limits via urllib3 (bundled with requests).
# See: https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry
GITHUB_RETRY = Retry(
    total=5,
    connect=5,
    read=5,
    backoff_factor=1,
    status_forcelist=(429, 500, 502, 503, 504),
    allowed_methods=frozenset(["GET"]),
    respect_retry_after_header=True,
)
GITHUB_ADAPTER = HTTPAdapter(max_retries=GITHUB_RETRY)

# for every function we can define / close a session before iterating.
# we can also define the session in the top level since this is a script just to run once / called ci.
def get_milestone_number(repo: str, title: str) -> int | None:
    url = f"https://api.github.com/repos/{repo}/milestones"
    page = 1
    recent_titles = []
    with requests.Session() as session:
        session.headers.update(HEADERS)
        session.mount("https://", GITHUB_ADAPTER)
        while True:
            resp = session.get(url, params=params, timeout=API_TIMEOUT)
            resp.raise_for_status()
            ...
```

NARSimoes (changes requested):
Thanks @amyblais , just one minor point to clarifications.

.github/scripts/generate_changelog.py
amyblais: Addressed!

 - Removed the custom ``make_github_request`` helper (which had a manual ``time.sleep`` retry loop and leftover time module usage).
 - Added ``GITHUB_RETRY`` + ``GITHUB_ADAPTER`` + ``SESSION`` module-level constants using the standard ``requests.Session + HTTPAdapter`` + ``urllib3.Retry`` pattern.
 - Updated all 3 call sites in ``get_milestone_number`` and ``get_merged_prs`` to use ``SESSION.get(..., timeout=API_TIMEOUT)`` followed by ``resp.raise_for_status()``.

.github/scripts/generate_changelog.py
```diff
@@ -20,6 +23,8 @@
 import sys
```
NARSimoes: I think we aren't using `sys` but can be addressed as follow up.

NARSimoes (approved):
Thanks @amyblais

.github/scripts/generate_changelog.py
amyblais: Addressed! Removed it.

PR #9001, opened 2026-05-28
Fix release docs automation

amyblais:
Fixes based on a request from IT:

 - Added ``actions/create-github-app-token@v1``, which generates a short-lived installation token from ``DOCS_APP_ID and DOCS_APP_PRIVATE_KEY``. That token is then passed to both the checkout step and ``GH_TOKEN`` in the PR step. The note on the permissions block updated to reflect this.
 - Also added a minor fix for "source/product-overview/mattermost-v11-changelog.md" file.

### DIFF
--- a/.github/scripts/update_docs.py
+++ b/.github/scripts/update_docs.py
@@ -50,7 +50,8 @@ MAX_SEND_CHARS = 50_000
 #   - Patch/Dot: releases page + changelog only
 # ---------------------------------------------------------------------------
  
-SERVER_FILES = [
+# Files updated for every server release type.
+SERVER_BASE_FILES = [
     "source/product-overview/mattermost-server-releases.md",
     "source/deployment-guide/server/linux/deploy-rhel.rst",
     "source/deployment-guide/server/linux/deploy-tar.rst",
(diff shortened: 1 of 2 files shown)


amyblais:
@coderabbitai full review

amyblais:
Updated; ready for re-review.

.github/workflows/update-docs.yml
```diff
@@ -45,13 +45,35 @@ on:
...
+      # Credentials are already present in Settings -> Secrets and variables -> Actions:
+      #   CHANGELOG_WRITE_CLIENT_ID   (variable) -- Client ID from the docs GitHub App
+      #   CHANGELOG_WRITE_PRIVATE_KEY (secret)   -- Private key from the docs GitHub App
+      - name: Get write token (docs repo)
+        id: app-token
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
+        with:
+          client-id: ${{ vars.CHANGELOG_WRITE_CLIENT_ID }}
+          private-key: ${{ secrets.CHANGELOG_WRITE_PRIVATE_KEY }}
+          owner: mattermost
```
NARSimoes, line 68: nit: `create-github-app-token` allows passing `repositories` so we can narrow the installation token only for the required repositories.

.github/scripts/update_docs.py
```diff
@@ -1,3 +1,4 @@
+
```
NARSimoes: nit: adding a extra line here.

NARSimoes (changes requested):
Thank you @amyblais , just two nit comments otherwise looks good.

.github/workflows/update-docs.yml
amyblais: Addressed! Added ``repositories: docs`` to the ``create-github-app-token`` step, scoping the installation token to only the ``docs`` repo rather than all repos the app is installed on.

.github/scripts/update_docs.py
amyblais: Fixed!

NARSimoes (approved):

PR #9003, opened 2026-06-01
docs: clarify Entra ID SAML attribute namespace matching

ewwollesen:
## Summary

Mattermost matches SAML assertion attributes by their **full** `Name` (or `FriendlyName`), including any namespace, character-for-character. Microsoft Entra ID emits its built-in claims under the `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/` namespace by default, so an admin who enters the short name shown in the Entra UI (e.g. `email`, `name`) into the Mattermost attribute fields gets a silent mismatch and login fails with:

```
SAML login was unsuccessful because one of the attributes is incorrect. Please contact your System Administrator., <attribute> attribute is missing
```

The current Entra ID guide mentions editing "claim names and namespaces" but doesn't explain that the value entered in Mattermost must be the fully-qualified claim name (or that the namespace must be cleared in Entra). The SAML FAQ has no entry for this error at all. This is a recurring support issue.

## Changes

- **`sso-saml-entraid.rst`**: add an `.. important::` note under the *Attributes & Claims* step explaining exact full-name matching, the two ways to reconcile it (clear the namespace in Entra, or use the fully-qualified name in Mattermost), and how to confirm the real claim names with SAML-tracer. Cross-referenced from the Mattermost-side attribute step.
- **`sso-saml-faq.rst`**: add a provider-agnostic troubleshooting entry for the `"one of the attributes is incorrect" / "<attribute> attribute is missing"` failure. This FAQ is `include`-ed on all SAML provider pages, so it surfaces for Okta/OneLogin/ADFS/Keycloak/Entra alike.

## Source

Derived from a support investigation (ticket 51591) where a captured Entra assertion confirmed the IdP sent `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/{email,name,emailaddress}` while the config used short names.

### DIFF
--- a/source/administration-guide/onboard/sso-saml-entraid.rst
+++ b/source/administration-guide/onboard/sso-saml-entraid.rst
@@ -40,7 +40,20 @@ Set up an enterprise app for Mattermost SSO in Entra ID
 9. Select **Edit** in the **Attributes & Claims** section then set the below attributes:
 
   a. Set the the **Unique User Identifier (Name ID)** required claim **Name identifier format** and **Source attribute** values as required for your environment. Setting the **Source attribute** to an immutable value such as ``user.objectid`` is recommended.
-  b. Edit claim names and namespaces under **Additional claims** to match SAML attribute settings you wish to set in Mattermost. Configurable settings are Email, Username, Id, Guest, Admin, First Name, Last Name, Nickname, Position, and Preferred Language.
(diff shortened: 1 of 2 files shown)


ewwollesen:
@hanzei can you please look over this when have a chance? I added some info about a namespace mismatch that comes up quite a bit for customers using Entra. If you don't have time, or you think someone else would be more appropriate, please let me know.

hanzei (approved):
LGTM 👍

wiersgallak (approved):

PR #9006, opened 2026-06-02
Fix release docs automation

amyblais:
Two fixes:

1. Updated the script to support a comma-separated version list.
2. Fixed timeout issue at https://github.com/mattermost/docs/actions/runs/26812807336/job/79046907907.

Updated spec doc: https://mattermost.atlassian.net/wiki/spaces/Security/pages/4518772760/Automation+-+Release+Docs+Update+GitHub+Actions+Workflow.

### DIFF
--- a/.github/scripts/update_docs.py
+++ b/.github/scripts/update_docs.py
@@ -1,3 +1,4 @@
+
 """
 update_docs.py
  
@@ -141,7 +142,7 @@ Rules:
 - Never remove or alter existing content unless it is explicitly outdated by this release.
 - Do not add placeholder text, commentary, or notes -- only real documentation content.
 - If a file does not need changes for this release type, return it exactly as-is.
-- Return ONLY the complete file content. No explanations, no markdown code fences, no preamble.
+- Return ONLY the content you were given (the full file, or the provided portion if the user prompt says the file is truncated). No explanations, no markdown code fences, no preamble.
 """
  
  
@@ -186,7 +187,9 @@ def update_file(client: anthropic.Anthropic, filepath: str) -> str:
     This keeps each Claude call small and well-defined, and ensures changelog
     entries accumulate correctly across versions.
  
-    Returns one of: "updated", "unchanged", "skipped", "not_found".
+    Returns one of: "updated", "unchanged", "not_found".
+    Version-level quality failures (empty response, too-short response) are logged
+    and skipped via continue -- they do not surface as a file-level "skipped" status.
     Raises on hard failures (I/O errors, API errors) so the caller can track them.
     """
     print(f"  Reading {filepath}...")
@@ -236,6 +239,16 @@ def update_file(client: anthropic.Anthropic, filepath: str) -> str:
             )
  
         updated = response.content[0].text
+ 
+        # For truncated files, Claude should return only the head portion, but it may
+        # occasionally return slightly more. Cap the result at MAX_SEND_CHARS so that
+        # subsequent version passes don't receive an ever-growing prompt.
+        if truncated and len(updated) > MAX_SEND_CHARS:
+            print(
+                f"  WARNING: Claude returned {len(updated):,} chars for {filepath} "
+                f"(version {version}); truncating to first {MAX_SEND_CHARS:,} chars."
+            )
+            updated = updated[:MAX_SEND_CHARS]
  
         # --- Content quality guards (warn and skip this version, not the whole file) ---
         if not updated.strip():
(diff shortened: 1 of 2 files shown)


.github/scripts/update_docs.py
```diff
@@ -27,10 +31,18 @@
...
 RELEASE_TYPE = os.environ["RELEASE_TYPE"]    # ESR | Feature Release | etc.
-VERSION = os.environ["VERSION"]
 RELEASE_DATE = os.environ["RELEASE_DATE"]
 ESR_END_DATE = os.environ.get("ESR_END_DATE", "").strip()
  
+# Parse VERSION as a comma-separated list so the workflow can be triggered once
+# for multi-version releases (e.g. security patches across several branches).
+# Each version is applied to every file in sequence; the output of one version
+# becomes the input for the next, so entries accumulate correctly in changelogs.
+VERSIONS: list[str] = [v.strip() for v in os.environ["VERSION"].split(",") if v.strip()]
```
NARSimoes, line 41: The "versions" approach isn't aligned with the respective workflow `update-docs.yaml` (e.g. expect version, we do normalization there, etc).

Based on the current error I'd do
- Open a new pull-request just bumping the timeout -> merge -> run pipeline with one version to ensure is green
- Keep updating this pull-request to support multiple versions since it'll require more logic.

NARSimoes (changes requested):
A clarification:

.github/scripts/update_docs.py
amyblais: Thank you for catching, the multi-version approach has now been applied to both files.

.github/scripts/update_docs.py
NARSimoes, line 96: Sanity check: I think we don't have logic that supports multiple versions and here we have only `v11`

.github/scripts/update_docs.py
```diff
@@ -169,80 +181,106 @@ def build_user_prompt(filepath: str, content: str, truncated: bool = False) -> s
...
+        # occasionally return slightly more. Cap the result at MAX_SEND_CHARS so that
+        # subsequent version passes don't receive an ever-growing prompt.
+        if truncated and len(updated) > MAX_SEND_CHARS:
+            print(
+                f"  WARNING: Claude returned {len(updated):,} chars for {filepath} "
+                f"(version {version}); truncating to first {MAX_SEND_CHARS:,} chars."
+            )
+            updated = updated[:MAX_SEND_CHARS]
+ 
+        # --- Content quality guards (warn and skip this version, not the whole file) ---
```
NARSimoes: Opening for discussion: if I'm reading it correctly now that we've multiple versions the workflow can exit success and open pull-request even when not every version was applied to every files. Is this expected? Should we consider adding some hardening here? (e.g. if any version was skipped return partial status, if any version skipped fails the workflow)

.github/scripts/update_docs.py
NARSimoes, line 41: I'd consider adding a hard limit e.g. `MAX_VERSIONS`. Every version multiplies costs and time (e.g. number api calls) so, if we introduce a typo by mistake we can do unnecessary API calls.

Just to explain the rationale
```
MAX_VERSIONS = 5

if len(VERSIONS) > MAX_VERSIONS:
    print(f"ERROR: {len(VERSIONS)} versions given; maximum is {MAX_VERSIONS}.")
    print("Split into multiple workflow runs or fix the version list.")
    sys.exit(1)
```

NARSimoes (changes requested):
A few clarifications

.github/scripts/update_docs.py
amyblais: Thank you for catching it! ``mattermost-v10-changelog.md`` added above v11 in ``SERVER_CHANGELOG_FILES``.

.github/scripts/update_docs.py
amyblais: ``MAX_VERSIONS = 5`` — added immediately after parsing ``VERSIONS``. If more than 5 versions are given, the script exits early with a clear message before making any API calls.

.github/scripts/update_docs.py
amyblais: Partial version application — the two content quality guards (``continue`` on empty response, ``continue`` on too-short response) now raise ``RuntimeError`` instead. Since ``RuntimeError`` is already caught in ``main()`` and marks the file as an error with exit code 1, a partial application now fails the run visibly rather than silently opening a PR with only some versions applied.

NARSimoes (approved):

PR #9012, opened 2026-06-05
docs: fix broken code, command, and link examples

emmayusufu:
While reading through the admin and deployment guides I found several examples that don't run or render as written. Each is a small, self-contained fix:

- **comply/electronic-discovery**: the `curl` example's closing single quote was after the URL, so the URL ended up inside the `Authorization` header value and no request URL was passed. Moved the quote to right after the token.
- **configure/CJK search**: missing semicolon after `CREATE EXTENSION zhparser` (the next statement on the following line has one).
- **manage/health-check-probes**: the Go import used a `github.com/.../tree/master/...` web URL with an unbalanced quote; replaced with the real import path `"github.com/mattermost/mattermost/server/public/model"`.
- **onboard/sso-entraid**: the *Microsoft Entra ID* hyperlink was missing its opening backtick, so it rendered as literal text instead of a link.
- **scale/enterprise-search**: `opensearch` used triple backticks; reStructuredText inline literals are double backticks (the adjacent `elasticsearch` uses double).
- **upgrade/admin-onboarding-tasks**: the cross-reference display text read `mmctl user deletall`; the command is `deleteall` (the ref target itself spells it that way).
- **upgrade/upgrading-postgres**: the PostgreSQL catalog is `pg_statistic`, not `pg_statistics`.
- **desktop/linux-install**: the `rpm` upgrade flag is `-U`, not `-u` (`-u` isn't a valid rpm option).
- **server/linux/deploy-tar**: `rm - rf` had a stray space inside the flag.

9 files, all one-line fixes. I'll sign the CLA so this can be merged.


emmayusufu:
Gentle nudge on this one. CI is green and the CLA check passed. It would be great to get a Core Committer review when someone has a moment, and I'm happy to rebase if anything has drifted since I opened it.

Combs7th:
Thanks so much for your contribution, @emmayusufu! Is there a particular dev who you'd recommend we tag here for the technical review? Once that's in, I can help with the final steps for editor review.

emmayusufu:
Thanks @Combs7th. I'm an outside contributor, so I don't know who owns these pages internally. I'll leave picking the right reviewer to your team. Each change is a small, self-contained correction that's easy to verify on its own. For example, the `curl` snippet had the `-H` quote closing after the URL so the header swallowed the path, `rpm -u` should be `-U` for an upgrade, `pg_statistics` should be `pg_statistic`, and the Entra ID page was missing the opening backtick on an RST link. None of them change meaning. They just fix syntax that doesn't run or render as written. I can adjust anything a reviewer flags.

esethna (approved):
These are great, thanks for catching these @emmayusufu

PR #9061, opened 2026-06-23
Fix incorrect JSON example for mmctl user migrate-auth (email → SAML)

ewwollesen:
## Summary

The **mmctl user migrate-auth** docs page contained two conflicting JSON examples for the SAML users file. This removes the inaccurate one.

The **"user.json Example"** block showed an array of objects with `auth_data` / `idp_id` / `saml_user_id` keys. **mmctl does not understand that format.** In `server/cmd/mmctl/commands/user.go`, `migrateAuthToSamlCmdF` reads the users file straight into a Go `map[string]string`:

```go
matches := map[string]string{}
if !auto {
    matchesFile := userArgs[2]
    file, err := os.ReadFile(matchesFile)
    ...
    if err := json.Unmarshal(file, &matches); err != nil {
        return fmt.Errorf("invalid json: %w", err)
    }
}
```

Feeding it the array form fails with `invalid json` (cannot unmarshal an array into a map). The keys `auth_data` / `idp_id` / `saml_user_id` appear nowhere in the migrate-auth code path.

The **correct** format — a flat object mapping each Mattermost email to its SAML username — is already documented immediately above, under **migration-options (saml) → users_file**:

```json
{
   "usr1@email.com": "usr.one",
   "usr2@email.com": "usr.two"
}
```

So this PR simply deletes the misleading second example, leaving the accurate one in place.

## Test plan

- [x] Confirmed against mmctl source (`server/cmd/mmctl/commands/user.go`, `migrateAuthToSamlCmdF`) that the users file is unmarshalled into `map[string]string`.
- [x] Verified `auth_data` / `idp_id` / `saml_user_id` are not referenced anywhere in the mmctl migrate-auth path.
- [x] `gmake clean html` builds successfully (exit 0). Build reports 26 warnings, all in unrelated files (Azure deploy guide, upgrade notes, autotranslation, generated agents docs, etc.) and pre-existing; **zero** reference `mmctl-command-line-tool.rst`. This change is a pure deletion and introduces no new warnings.


ewwollesen:
@hanzei Would you mind taking a look at this, please? It's just a simple docs change, but I want to make sure that I (and Claude) have read the code correctly that block isn't actually needed.

hanzei (approved):
Good catch. The changes look good to me 👍

BenCookie95 (approved):

esethna (approved):

PR #9066, opened 2026-06-24
Update community-chat.rst

nivaas219:


esethna:
Thanks for this submission but I think the grammar is more correct in the original. I will close this PR, but would appreciate your help on future improvements.

PR #9067, opened 2026-06-25
Fix blank redirect stubs for unmapped fragments

hanzei:
## Why

Visiting a redirect-stub URL with an anchor that wasn't explicitly mapped returns a **blank page**. For example, http://docs.mattermost.com/configure/environment-configuration-settings.html#database renders nothing — the old `configure/…` path is a redirect stub, and `#database` was never added to its fragment map, so the JS logged `unknown redirect: database` and left the user staring at an empty page.

Redirect stubs (generated from `redirects.py` via `source/_templates/redirect.html`) only redirected fragments **explicitly listed** in `redirects.py`. Any other fragment hit the `else` branch and blanked out. The `// TODO: redirect the user to a 404 page` was also still unimplemented.

## Changes

**`source/_templates/redirect.html`:**

1. **Broader fallback for unmapped fragments.** When a fragment isn't explicitly mapped, redirect to the page's base destination (the `-` default entry) and **carry the original fragment over** — anchor names are generally preserved across a page move, so `…#database` / `…#mobile-security` now land on the real section instead of a blank stub. Fixes the whole class of unmapped-anchor links.
2. **Resolved the `// TODO`.** When no redirect can be resolved at all (no matching fragment and no base entry), the user is sent to `/404.html` instead of a blank page.
3. **Own-property lookup.** `fragment_redirects` is a plain object, so `fragment in …` also matched inherited members (`#toString`, `#constructor`, …), passing a non-string to `redirectTo()`. Switched to `Object.prototype.hasOwnProperty.call`.
4. Extracted the leading-`/` → absolute-origin handling into a small `redirectTo()` helper, and log the fallback at `console.debug` (not `warn`) so routine redirects stay quiet.

**`source/redirects.py`:**

5. **Removed 233 now-redundant fragment redirects.** With the fallback in place, any entry whose target is exactly `<base-target>#<same-fragment>` is redundant — the fallback produces the identical destination. Kept untouched: fragments on pages with no base entry (the fallback can't reach them → they'd 404), fragments pointing to a different page, and fragments whose anchor was renamed at the destination. (2166 → 1933 entries.)

## Verification

- Verified in a browser against the PR preview: both direct navigation and the redirect path land `#database` exactly 150px below the fixed header (matching `section[id] { scroll-margin-top }`), stable across the full page load. Works in Chrome and Firefox. (There's a pre-existing Edge-only quirk where a long page's late reflow leaves the anchor slightly high — unrelated to this change and out of scope.)
- `redirects.py` re-imports cleanly; 0 redundant entries remain; the 34 no-base fragment entries are preserved.

### DIFF
--- a/source/redirects.py
+++ b/source/redirects.py
@@ -322,8 +322,6 @@ redirects_map = {
         "https://docs.mattermost.com/administration-guide/manage/mmctl-command-line-tool.html#mmctl-permissions-reset",
 "administration/command-line-tools.html#mattermost-permissions-export":
         "https://api.mattermost.com/#tag/roles/operation/GetAllRoles",
-"administration/command-line-tools.html#mattermost-permissions-import":
-        "https://docs.mattermost.com/administration-guide/manage/command-line-tools.html#mattermost-permissions-import",
 "administration/command-line-tools.html#mattermost-group-team-list":
         "https://docs.mattermost.com/administration-guide/manage/mmctl-command-line-tool.html#mmctl-group-team-list",
 "administration/command-line-tools.html#mattermost-group-team-enable":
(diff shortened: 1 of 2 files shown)


source/redirects.py
hanzei, line 1: These are not needed any longer. The new code handles them automatically.

esethna (changes requested):
Thanks @hanzei. Can you confirm that it's intentional you have removed so many redirects? Spot checking a few seem to be causing missed subsection redirects, for example: `/administration/ediscovery.html#mattermost-restful-api`

Also I noticed that mistyped or invalid URLs now are leading to a 404 page vs current behaviour of at least landing them at the home docs.mattermost.com. For example `https://docs.mattermost.com/admin` (page does not exist)

Is there a safer way we can make this change to avoid blank pages without potential regressions?

esethna:
Friendly ping @hanzei ^

Combs7th:
@hanzei - Just another friendly ping on this one.^

PR #9073, opened 2026-06-26
Add consolidated "Install and manage plugins" documentation

ewwollesen:
## Summary

Adds a consolidated, plugin-agnostic admin guide for the full plugin lifecycle. Previously this guidance was scattered across per-plugin install pages (Boards, Legal Hold, Channel Export, Agents, etc.) and configuration references, with no single "how do I install/update/remove a plugin" page.

## What's included

New page **`administration-guide/configure/manage-plugins.rst`** ("Install and manage plugins") covering:

- **Before you begin** — prerequisite settings (Enable plugins, Enable Marketplace / remote Marketplace, Upload Plugin, Require plugin signature), with links to the config reference.
- **Install a plugin** — from the Marketplace (recommended), by uploading a bundle (custom/3rd-party/air-gapped), and via `mmctl`.
- **Enable or disable**, **Configure**, **Update** (auto pre-packaged, Marketplace update, re-upload same ID, `mmctl plugin add -f`), and **Remove**.
- **Air-gapped and restricted environments** note.
- **Related documentation** cross-links.

The page links out to the existing reference pages (plugins configuration settings, mmctl, plugins overview, popular integrations) rather than duplicating them.

## Wiring

- Registered in the `server-configuration.rst` hub (toctree + bullet list), next to the Boards entry.
- Added a discovery pointer from `integrations-guide/plugins.rst`.

## Build

Incremental `gmake html` succeeds with no new warnings; the page renders to `manage-plugins.html`.

## Notes for reviewers

- **No plan/edition badge** was added, since the plugin framework is available across all editions (matching `integrations-guide/plugins.rst`). Happy to add `all-commercial` if preferred.
- Placed under `administration-guide/configure/` alongside `install-boards.rst`; open to relocating if `integrations-guide/` is a better home.


source/administration-guide/configure/manage-plugins.rst
```diff
@@ -0,0 +1,177 @@
...
+.. tip::
+
+  The `Mattermost Marketplace <https://mattermost.com/marketplace/>`__ website offers an expanded selection of community-supported integrations.
+
+By uploading a plugin bundle
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Upload a plugin bundle when you're installing a custom or third-party plugin that isn't in the Marketplace, or when your server runs in an air-gapped or restricted environment. Uploading requires the **Upload Plugin** setting to be enabled and the **Require plugin signature** setting to be disabled.
+
+1. Download the plugin bundle. Mattermost plugins are distributed as ``.tar.gz`` files, typically from the plugin's GitHub releases page.
```
jadrales: I think it would be valuable to add something that mentions to download the package for their specific OS, since that can be a common issue for the plugin upload not working. In general, I might rephrase to something like:
> Download the plugin bundle `.tar.gz` for your specific OS from the plugin's GitHub releases page.

source/administration-guide/configure/manage-plugins.rst
```diff
@@ -0,0 +1,177 @@
...
+3. Go to **Plugins > Plugin Management**.
+4. In the **Upload Plugin** section, select **Choose File**, select the plugin bundle you downloaded, and then select **Upload**.
+5. The plugin appears in the **Installed Plugins** list. Select **Enable** to turn it on.
+
+Using mmctl
+~~~~~~~~~~~
+
+The :doc:`mmctl command line tool </administration-guide/manage/mmctl-command-line-tool>` is useful for scripted or remote plugin management. Plugins must be enabled in the server configuration first.
+
+- Install the latest version from the Marketplace by plugin ID:
```
jadrales, line 67: Wondering if we should provide some guidance on how to grab the plugin ID? For most plugins it's usually just the name of the plugin (e.g. `jira`), but others have less of a straightforward ID (e.g. Calls has an ID of `com.mattermost.calls` rather than just `calls`). We could add a quick one-liner here about checking the plugin's `plugin.json` for this, unless there's an easier/better way to grab that info

source/administration-guide/configure/manage-plugins.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Installing a plugin doesn't turn it on automatically. Enable a plugin to make it available to users, and disable it to remove it from the user interface without uninstalling it.
+
+To enable or disable a plugin in the System Console:
+
+1. Go to **Plugins > Plugin Management**.
+2. In the **Installed Plugins** list, locate the plugin.
+3. Select **Enable** or **Disable**.
+
+To enable or disable a plugin with mmctl:
```
jadrales, line 102: It might be helpful to have some subheaders here, so you can directly link others to specific sections on how to enable/disable plugins, similar to how "Install a Plugin" is set up. For example, we could have a subheader for From the System Console and one for Using mmctl for a structure like:
> Enable or disable a plugin
>   > From the System Console
>   > Using mmctl

I'm like 1/5 on this though so we can leave as-is if we think that the extra headers won't provide much value

source/administration-guide/configure/manage-plugins.rst
```diff
@@ -0,0 +1,177 @@
...
+------------------
+
+Most plugins have their own settings page. After a plugin is installed, go to **Plugins** in the System Console and select the plugin by name to configure it, then select **Save**. Refer to the documentation for each plugin for details on its available settings.
+
+Update a plugin
+---------------
+
+We recommend keeping plugins up to date as new versions are released. Updates are generally seamless and don't interrupt the user experience, but you should review each plugin's release notes for compatibility considerations, and test updates in a staging environment before applying them in production.
+
+You can update a plugin using any of the following methods:
```
jadrales, line 123: Similar feedback as before, I wonder if having subheaders/sections would be beneficial here for easier linking. Still like 1/5 on it though.

source/administration-guide/configure/manage-plugins.rst
```diff
@@ -0,0 +1,177 @@
...
+.. tip::
+
+  To confirm the version currently installed, go to **Plugins > Plugin Management** and review the **Installed Plugins** list, or run ``mmctl plugin list``.
+
+Remove a plugin
+---------------
+
+Removing a plugin disables it and uninstalls it from the server.
+
+To remove a plugin in the System Console:
```
jadrales, line 148: Same feedback as before regarding subsections.

source/administration-guide/configure/manage-plugins.rst
```diff
@@ -0,0 +1,177 @@
...
+
+To remove a plugin with mmctl:
+
+.. code-block:: sh
+
+   mmctl plugin delete <id>
+
+.. note::
+
+  Removing a plugin uninstalls it from the server and stops it from running, but Mattermost preserves the plugin's associated data:
```
jadrales, line 162: Is there any guidance on how this data can be fully cleared/deleted that we can add here in the event that that removal is a want or requirement for an instance?

jadrales (commented):
Formatting LGTM. I left a few comments regarding some small changes we can consider for the Manage Plugins page, but overall I think this will be a great addition to the admin docs.

source/administration-guide/configure/manage-plugins.rst
```diff
@@ -0,0 +1,177 @@
...
+Plugin availability and the actions you can take depend on a few configuration settings found in the System Console under **Plugins > Plugin Management**. Review these settings before installing or updating plugins:
+
+- **Enable plugins** must be set to **true** (the default) for the plugin system to work. When disabled, all plugins are turned off.
+- **Enable Marketplace** must be set to **true** (the default) to install plugins from the in-product Marketplace.
+- **Enable remote Marketplace** must be set to **true** (the default) for the Marketplace to connect to the remote endpoint and list community and Mattermost-provided plugins. Set this to **false** for servers that can't reach the internet; the Marketplace then shows only pre-packaged and installed plugins.
+- **Upload Plugin** (``PluginSettings.EnableUploads``) must be set to **true** to upload plugin bundles from your local computer. This setting applies to self-hosted deployments only.
+- **Require plugin signature**, when enabled, validates plugin signatures and **disables plugin file uploads** in the System Console. With this setting enabled, install plugins from the Marketplace or as pre-packaged plugins instead.
```
esethna: This last one re plugin signature maybe not needed for this section?

source/administration-guide/configure/manage-plugins.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Before you begin
+----------------
+
+Plugin availability and the actions you can take depend on a few configuration settings found in the System Console under **Plugins > Plugin Management**. Review these settings before installing or updating plugins:
+
+- **Enable plugins** must be set to **true** (the default) for the plugin system to work. When disabled, all plugins are turned off.
+- **Enable Marketplace** must be set to **true** (the default) to install plugins from the in-product Marketplace.
+- **Enable remote Marketplace** must be set to **true** (the default) for the Marketplace to connect to the remote endpoint and list community and Mattermost-provided plugins. Set this to **false** for servers that can't reach the internet; the Marketplace then shows only pre-packaged and installed plugins.
+- **Upload Plugin** (``PluginSettings.EnableUploads``) must be set to **true** to upload plugin bundles from your local computer. This setting applies to self-hosted deployments only.
```
esethna, line 22: Can we be consistent about including the plugin variable like this with all the settings mentioned in this doc (whenever it is mentioned for the first time)?

source/administration-guide/configure/manage-plugins.rst
```diff
@@ -0,0 +1,177 @@
...
+- **Require plugin signature**, when enabled, validates plugin signatures and **disables plugin file uploads** in the System Console. With this setting enabled, install plugins from the Marketplace or as pre-packaged plugins instead.
+
+.. note::
+
+  If you receive a ``Received invalid response from the server`` error when uploading a plugin, the :ref:`maximum file size <administration-guide/configure/environment-configuration-settings:maximum file size>` configuration setting is likely too small for the plugin bundle. Increase it, and update any proxy settings as needed.
+
+See the :doc:`plugins configuration settings </administration-guide/configure/plugins-configuration-settings>` documentation for details on each of these settings, including their ``config.json`` paths and environment variables.
+
+Install a plugin
```
esethna: Propose: "Plugin installation methods"

Then each nested section can be "App Marketplace" "Plugin Upload" "mmctl"

source/administration-guide/configure/manage-plugins.rst
esethna: +1

source/administration-guide/configure/manage-plugins.rst
esethna: I think it's fine as is given the content isn't very long, so it's easy enough to scan for the method the reader is looking for.

source/administration-guide/configure/manage-plugins.rst
esethna: Same as above: I think it's fine as is given the content isn't very long, so it's easy enough to scan for the method the reader is looking for.

esethna (changes requested):

esethna:
Thanks @ewwollesen, nice doc, some feedback provided ^

source/administration-guide/configure/manage-plugins.rst
ewwollesen: Good call — done in 648cd726. Each plugin setting now shows its `PluginSettings.*` key on first mention: `Enable`, `EnableMarketplace`, `EnableRemoteMarketplace`, `EnableUploads`, `AutomaticPrepackagedPlugins`, and `RequirePluginSignature` (now introduced in the Plugin upload section, since the prerequisite bullet was removed per your other comment).

source/administration-guide/configure/manage-plugins.rst
ewwollesen: Agreed — removed it from *Before you begin* in 648cd726. It's still covered in the Plugin upload section, where the signature/upload interaction is directly relevant.

source/administration-guide/configure/manage-plugins.rst
ewwollesen: Done in 648cd726 — the section is now *Plugin installation methods* with *App Marketplace* / *Plugin upload* / *mmctl* subsections. I kept *App Marketplace* capitalized since it's the literal menu name, and used sentence case for *Plugin upload* to match our heading style guide. Happy to switch it to *Plugin Upload* if you'd rather all three read as title case.

source/administration-guide/configure/manage-plugins.rst
ewwollesen: Added download guidance here in 648cd726, with one tweak worth flagging: looking at the plugin build system, standard Mattermost bundles package the server binaries for *all* supported platforms in a single `.tar.gz` (the manifest's `server.executables` map), and the server auto-selects the matching one at load. So "download for your OS" isn't quite accurate as blanket advice. I worded it so the per-platform caveat applies only when a plugin publishes separate per-OS assets. If the upload failures you've seen trace to something more specific, let me know and I'll capture that instead.

source/administration-guide/configure/manage-plugins.rst
ewwollesen: Added in 648cd726 — the mmctl section now explains that a plugin's ID is the `id` value in its `plugin.json`, often the plugin name (`github`, `playbooks`) but sometimes a reverse-DNS identifier like `com.mattermost.calls` for Calls, and that `mmctl plugin list` shows the IDs of installed plugins. Thanks — and thanks to @esethna for the +1.

source/administration-guide/configure/manage-plugins.rst
ewwollesen: Added a note in 648cd726. Confirmed against the server source: removing a plugin disables it and deletes the bundle, but intentionally retains its config, KV store data, and any bot accounts — and there's no supported System Console or mmctl command to fully purge that data. Clearing it requires direct database changes, so the note points admins to Mattermost support rather than documenting raw table edits. If you'd prefer we spell out the specific tables, I can expand it.

source/administration-guide/configure/manage-plugins.rst
ewwollesen: Correction to my note above — you were right. I checked the actual release assets (Calls v1.12.2, plus GitHub/GitLab/Jira) and they do publish per-platform bundles alongside the universal one, with the per-platform ones much smaller (Calls: 31 MB `linux-amd64` vs 92 MB universal). My earlier check only looked at the default build target and missed the released per-platform assets, so your original point stands. Reworded step 1 in 4e2316ba to recommend the bundle matching the server's OS/architecture (a mismatched one won't start), with the universal bundle as the auto-selecting fallback.

jadrales:
LGTM

jadrales (approved):

esethna (approved):

PR #9081, opened 2026-07-02
Fix Bash "safe mode" in upgrade_mattermost.sh

eric-j-ason:
This fixes two aspects of the changed line:
- Raises an error upon using an uninitialized variable. (The already existing `nounset` did not do what was intended.)
- Raises an error if any command in a pipeline fails.

The behavior now conforms to what the comment says it does:

https://github.com/mattermost/docs/blob/3c001a72caaeb0380c1266f2a01cf0c5afa362f3/source/administration-guide/upgrade/upgrade_mattermost.sh#L3-L5


eric-j-ason:
Nevermind, I see this requires signing some agreement. I've made it into an issue instead, so that someone else can commit it.

#9082

PR #9094, opened 2026-07-14
Update docs automation prompt

amyblais:
Fixes based on a test PR: https://github.com/mattermost/docs/pull/9093.

Server PR: https://github.com/mattermost/mattermost/pull/37495


esethna:
@amyblais given the work Eva's team is doing to migrate the docs repo to the main code repo, planned for this month afaik, I'm wondering if we should hold off on any of the docs automation work since it will need to be rebuilt anyway? Perhaps we just disable the function for now so it's not opening new docs PRs automatically?

amyblais:
> given the work Eva's team is doing to migrate the docs repo to the main code repo, planned for this month afaik, I'm wondering if we should hold off on any of the docs automation work since it will need to be rebuilt anyway? Perhaps we just disable the function for now so it's not opening new docs PRs automatically?

Sounds good 👍

PR #9107, opened 2026-07-20
Document per-node, per-data-source scope of MaxOpenConns

mrckndt:
#### Summary

Clarifies that `MaxOpenConns` is applied independently on each Mattermost server node, for the master database and for each configured read/search replica - not as cluster-wide limits. The docs previously said these settings apply "per data source" but never mentioned the additional multiplication by node count, which could lead to undersizing `max_connections` on the database.


roberson-io (approved):

vish9812 (approved):
Great addition. Thanks @mrckndt

PR #9116, opened 2026-07-28
Document that Elasticsearch/OpenSearch user search does not match email address

ewwollesen:
Elasticsearch and AWS OpenSearch-backed user search matches on username, nickname, and first and
last name only — email address is not indexed and cannot be matched. Database-backed user search
does match on email, so enabling enterprise search silently removes email matching from user
search.

This was previously undocumented. The `Enterprise search limitations` list did not mention it, and
the admin onboarding guide's "Enable Elasticsearch" step stated that Elasticsearch "solve[s] many
known issues with full text database search, such as dots, dashes, and email addresses returning
unexpected results" without scoping that to message search — which reads as though email handling
improves.

Changes:
- Add a fourth item to `Enterprise search limitations` describing the behavior, including which
  user searches are still database-backed and still match email, and the `Show email address`
  dependency for non-admins.
- Scope the onboarding-guide sentence to message search and link to the limitations list.
- Drive-by grammar fix in the same list: item 1 read "Elasticsearch and AWS OpenSearch uses a
  standard selection of stop words"; the subject is plural, so this is now "use". Committed
  separately (`9236ddce`) for easy review. This is the only occurrence in `source/`.

Reported by an Enterprise customer who diagnosed it from `GET users/_mapping` after migrating to
AWS OpenSearch.


source/administration-guide/scale/common-configure-mattermost-for-enterprise-search.rst
```diff
@@ -50,3 +50,4 @@ Enterprise search limitations
...
+4. User search and autocomplete served by Elasticsearch or AWS OpenSearch match on username, nickname, and first and last name only. Email addresses aren't indexed, so users can't be found by email address. User searches served from the database continue to match on email address, including the **System Console > User Management > Users** list, and channel and team member searches. For users who aren't system admins, email address matching also requires :ref:`Show email address <administration-guide/configure/site-configuration-settings:show email address>` to be enabled.
```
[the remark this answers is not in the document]
ewwollesen: I verified against current code rather than taking the suggestion at face value, and the reviewer's proposed wording would have introduced a factual error.

What the code actually does — server/channels/api4/user.go:1230:

if c.App.SessionHasPermissionTo(*c.AppContext.Session(), model.PermissionManageSystem) {
    options.AllowEmails = true
    options.AllowFullNames = true
} else {
    options.AllowEmails = *c.App.Config().PrivacySettings.ShowEmailAddress
    options.AllowFullNames = *c.App.Config().PrivacySettings.ShowFullName
}

The gate is the manage_system permission, which is granted to the system_admin role.

Why I didn't use the suggested phrasing. The reviewer asked me to say the setting applies to users who "lack permission to view email addresses." No such permission exists. Grepping server/public/model/permission.go for email-related permissions returns only sysconsole_read_authentication_email and sysconsole_write_authentication_email (the Email authentication settings page — unrelated to viewing user emails), plus invalidate_email_invite and test_email. Writing "permission to view email addresses" would send readers hunting for a permission toggle that isn't there.

What I did instead. Took the valid part — the system-admin behavior was implied by a negative carve-out rather than stated:

▎ For users who aren't system admins, email address matching also requires Show email address to be enabled.
▎ System admins can always match on email address. For all other users, email address matching also requires Show email address to be enabled.

Search-scope details and the surrounding sentences are untouched, per the reviewer's request to preserve them.

I kept "system admins" rather than naming the manage_system permission. It's accurate for any default deployment, and matches the register used elsewhere in these docs. The one gap: under Enterprise granular/custom admin roles, manage_system could in principle be assigned to a non-system-admin role, which the docs wording wouldn't cover — an edge case I judged not worth the precision cost in a limitations list, but easy to add if you'd rather be exact.

source/administration-guide/scale/common-configure-mattermost-for-enterprise-search.rst
```diff
@@ -47,6 +47,7 @@ Once the configuration is saved, new posts made to the database are automaticall
...
+4. User search and autocomplete served by Elasticsearch or AWS OpenSearch match on username, nickname, and first and last name only. Email addresses aren't indexed, so users can't be found by email address. User searches served from the database continue to match on email address, including the **System Console > User Management > Users** list, and channel and team member searches. System admins can always match on email address. For all other users, email address matching also requires :ref:`Show email address <administration-guide/configure/site-configuration-settings:show email address>` to be enabled.
```
jadrales: Only question here is whether sysadmins can be matched on email address using Elasticsearch/Opensearch. If they can't, then I think this line should be adjusted: 
> System admins can always match on email address.

Since it leaves some room for ambiguity. Users might read this and think that sysadmins are an exception to the previous statements.

jadrales (approved):
Overall LGTM. Left one comment regarding clarification for system admin users, but everything else checks out.

source/administration-guide/scale/common-configure-mattermost-for-enterprise-search.rst
ewwollesen: Good catch — you're right, and the answer to your question is no: system admins can't be matched on email address via Elasticsearch/OpenSearch either. Fixed in 36a547d.

Two independent reasons, both verified against the server code:

1. **There's no field to match against.** The indexed user document has exactly seven fields — `id`, `suggestions_with_fullname`, `suggestions_without_fullname`, `delete_at`, `roles`, `team_id`, `channel_id` (`ESUser` in `server/enterprise/elasticsearch/common/common.go`), and the index template in `common/templates.go` maps no `email` property. No permission level can match a field that isn't in the index.

2. **The admin exemption is structurally database-only.** The `manage_system` check at `server/channels/api4/user.go:1230` sets `options.AllowEmails = true`, but `AllowEmails` is never referenced anywhere in `server/enterprise/elasticsearch/` — it's consumed in exactly one place, `server/channels/store/sqlstore/user_store.go:1715`, inside the SQL query builder. So it can't reach the ES/OS prefix query even in principle.

So the sentence was wrong as written. Sitting right after the ES/OS statements, "always" read as spanning all backends, which is exactly the exception-to-the-previous-statements misreading you flagged.

The line now scopes each claim to the backend that serves it:

> 4. User search and autocomplete served by Elasticsearch or AWS OpenSearch match on username, nickname, and first and last name only. Email addresses aren't indexed, so no user can be found by email address, including system admins. User searches served from the database continue to match on email address. For channel and team member searches, system admins always match on email address, and all other users also require **Show email address** to be enabled. The **System Console > User Management > Users** list always matches on email address, independently of that setting.

While scoping this I found a second problem in the same sentence, which your comment led me to. The previous version grouped the System Console user list together with channel and team member searches and then applied the **Show email address** dependency to both. That's wrong for the System Console: it's served by the reporting API (`server/channels/api4/report.go`), gated on `sysconsole_read_user_management_users`, and that path never reads `ShowEmailAddress`. As written it would have told admins that disabling **Show email address** breaks email lookup in the System Console. It doesn't, and the new wording carves it out explicitly.

Combs7th:
Thanks so much, @ewwollesen! This looks good from my "Novice Nate" perspective - clear and easy to follow. I'm cool with approving!

Combs7th (approved):

PR #9129, opened 2026-08-06
Move Mattermost systemd unit file from /lib/systemd/system to /etc/systemd/system

mrckndt:
## Summary
- `/lib/systemd/system` (aliased to `/usr/lib/systemd/system` on merged-usr systems) is reserved for unit files installed and owned by a package manager. Our tarball and RHEL install guides have the admin create the Mattermost unit file by hand, so `/etc/systemd/system` is the correct location.
- Updates the unit file creation step in the tarball and RHEL install guides, with a note that earlier versions of the guide used `/lib/systemd/system`.
- Updates every downstream reference to the unit file path (troubleshooting guide, transport encryption guide, MySQL/database prep guides) for consistency. Two other pages already referenced `/etc/systemd/system` for this file, so this also removes an existing inconsistency rather than introducing one.

### DIFF
--- a/source/administration-guide/configure/configuration-in-your-database.rst
+++ b/source/administration-guide/configure/configuration-in-your-database.rst
@@ -100,7 +100,7 @@ The second line of output will have the location of the running ``mattermost.ser
 
 .. code-block:: text
 
-      Loaded: loaded (/lib/systemd/system/mattermost.service; enabled; vendor preset: enabled)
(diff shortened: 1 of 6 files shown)


shiftZZnet:
Reviewed changes, looks great. Thanks for implementing.

shiftZZnet (approved):
All looks fine, thanks for implementing, changes approved.

PR #9136, opened 2026-08-12
Document how to upgrade the rtcd service

hanzei:
## Review page

http://mattermost-docs-preview-pulls.s3-website-us-east-1.amazonaws.com/9136/administration-guide/configure/calls-rtcd-setup.html#upgrading-rtcd

## AI  Summary

We had no documentation on how to upgrade a running RTCD deployment. The Calls pages covered deploying and scaling the service, and the upstream [rtcd repo docs](https://github.com/mattermost/rtcd/tree/master/docs) cover install and config, but neither described the upgrade procedure. The only mention of upgrade ordering anywhere was a v7.7-specific row in [Important Upgrade Notes](https://docs.mattermost.com/administration-guide/upgrade/important-upgrade-notes.html) stating that rtcd must be upgraded to >= 0.8.0 first.

This adds an `Upgrading RTCD` section to `calls-rtcd-setup.md`, placed after `Horizontal Scaling` (whose DNS-based load balancing the rolling procedure depends on) and before `Integration with Mattermost`.

- **The plugin's dependency on the service.** The Calls plugin opens its connection when it starts and fails to start if RTCD isn't reachable, so RTCD has to be up first. Startup and runtime failures are then distinguished, because they need different responses: an activation failure leaves Calls deactivated with no automatic retry (the server's health check only monitors plugins that activated successfully), so the plugin has to be restarted once RTCD is back; whereas an established connection reconnects on its own, up to 8 attempts per host, after which a host still advertised in DNS is re-discovered by the 10 second host check.
- **Preserving the data store.** RTCD persists a Bitcask store at `[store] data_source` (default `/tmp/rtcd_db`) holding registered client IDs and bcrypt hashes of their auth keys. Covers why it has to survive a binary or image replacement, the bare-metal vs Docker difference, backup, and what happens when it's lost: recovery is automatic when `allow_self_registration` is on and impossible when it's off (the default), since registration itself requires authenticating first. Includes a warning that the setting permits unauthenticated registration and belongs only on a private, access-controlled network.
- **Version compatibility.** The plugin enforces a minimum RTCD version, so RTCD should be upgraded before the server. Documents both failure modes: a below-minimum host fails the check at plugin activation and prevents Calls from starting at all (if *any* resolved host fails, not just all of them), whereas a host discovered later via DNS is just logged and skipped. The practical upshot is that a stale host left in the DNS record can block Calls the next time the server restarts, even while calls are working.
- **Drain-on-shutdown behavior.** On `SIGTERM`/`SIGINT`, RTCD waits for all active sessions to end, with no drain timeout. Two consequences are called out: the HTTP/WS listeners stay open during the drain, so the server must leave DNS *before* it's signalled or it keeps getting new calls; and any supervisor that force-kills after a timeout cuts off calls in progress, including `systemctl stop` under systemd's 90 second `DefaultTimeoutStopSec`.
- **Single-server upgrades.** The two options (wait out the drain, or notify and force down at a set time) and their trade-offs. Placed before the rolling upgrade so the simpler case reads first.
- **Rolling upgrade.** Step-by-step for a horizontally scaled fleet: pull from DNS, wait for the plugin to flag the host (10s resolution interval), wait for `rtcd_rtc_sessions_total` to reach zero, stop, install, verify via `/version`, return to DNS.
- **Kubernetes.** Documents that `configuration.terminationGracePeriod` maps to the pod's `terminationGracePeriodSeconds` and warns against lowering the 5 hour default, since that turns a graceful drain into a `SIGKILL` that drops calls. Also notes that the chart sets no `maxSurge`, so a Deployment rollout can run old and new pods concurrently for the length of the drains, and covers how the data store is handled across pod replacements.


source/administration-guide/configure/calls-rtcd-setup.md
```diff
@@ -363,6 +363,108 @@ To scale RTCD horizontally:
...
+
+   Remove its IP address from the DNS record that the **RTCD Service URL** resolves to.
+
+2. **Wait for the plugin to pick up the change**:
+
+   The plugin re-resolves the hostname every 10 seconds and flags servers that are no longer advertised. A flagged server is excluded from new calls, while the calls already running on it continue uninterrupted.
+
+3. **Wait for the server to go idle**:
+
+   The `rtcd_rtc_sessions_total` metric reports the number of active RTC sessions (see [RTCD Metrics](calls-metrics-monitoring.md#rtcd-metrics)). The server can be restarted safely once this reaches zero.
```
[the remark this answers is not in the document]
hanzei: won't do

Combs7th:
@hanzei - Thank you for your submission! As a next step, can you help add a dev who can help give this a dev review for us? And can you also update the appropriate server release milestone if applicable?

source/administration-guide/configure/calls-rtcd-setup.md
```diff
@@ -363,6 +363,136 @@ To scale RTCD horizontally:
...
+
+Failures at startup and failures later on behave differently:
+
+- **During plugin activation**, an initial connection failure leaves Calls deactivated, and the server's plugin health check doesn't retry it, because it only monitors plugins that activated successfully. Once RTCD is reachable, restart the Calls plugin to bring it back.
+- **After a connection is established**, later disconnections are retried automatically, so a brief RTCD restart needs no action on the Mattermost side. The plugin makes up to 8 reconnection attempts for a host before dropping it, and a dropped host that's still advertised in DNS is picked up again by the 10 second host check.
+```
+
+### Version Compatibility
+
+The Calls plugin enforces a minimum RTCD version and won't use a server running an older one. All Calls versions from v1.5.0 onwards require RTCD v0.17.0 or later.
```
bgardner8008: I think it would be better to drop the specific versions here, as they are ancient, and simply refer the user to check for the minimum RTCD version required by calls plugin. This is currently found in the plugin manifest as props.min_rtcd_version.

source/administration-guide/configure/calls-rtcd-setup.md
```diff
@@ -363,6 +363,136 @@ To scale RTCD horizontally:
...
+
+   Remove its IP address from the DNS record that the **RTCD Service URL** resolves to.
+
+2. **Wait for the plugin to pick up the change**:
+
+   The plugin re-resolves the hostname every 10 seconds and flags servers that are no longer advertised. A flagged server is excluded from new calls, while the calls already running on it continue uninterrupted.
+
+3. **Wait for the server to go idle**:
+
+   The `rtcd_rtc_sessions_total` metric reports the number of active RTC sessions (see [RTCD Metrics](calls-metrics-monitoring.md#rtcd-metrics)). The server can be restarted safely once this reaches zero.
```
bgardner8008: I think you need to sum RTC sessions over all calls:

The rtcd_rtc_sessions_total metric reports the number of active RTC sessions per call group (see  RTCD Metrics (calls-metrics-monitoring.md#rtcd-metrics)). The server can be restarted safely once  the sum across all groups reaches zero.

bgardner8008 (changes requested):
This is a great improvement, thanks for doing this! I have a few suggestions.

source/administration-guide/configure/calls-rtcd-setup.md
hanzei: Yeah, referring to an ancient version is not helpful. I just dropped exact version suggestion from the PR.  referring to ` props.min_rtcd_version` is a bit too technical, so I didn't add that.

source/administration-guide/configure/calls-rtcd-setup.md
hanzei: Good catch, fixed :+1:

bgardner8008 (approved):
Thanks, looks good. I note you didn't add any description of the admin key. But I'm happy to approve anyway, as this is still a big improvement. Also we're trying not to spend lots of effort on Calls v1, an elusive goal.

hanzei:
Sorry Bill, I might be missing context here. What is the admin key you are talking about?

Combs7th:
Thanks for adding this! The upgrade guidance is clear, thorough, and gives admins useful coverage for single-server, rolling, and Kubernetes deployments. Looks good from my novice nate perspective.

Combs7th (approved):

PR #9160, opened 2026-08-21
Scope ClusterSettings.ReadOnlyConfig guidance to file-backed config

roberson-io:
## Summary

The HA cluster-based deployment guide currently tells admins to set `ClusterSettings.ReadOnlyConfig true` unconditionally in step 7 of the deployment procedure, and the High Availability settings section states that "the System Console is set to read-only" as a universal HA behavior. Both framings are misleading when configuration is stored in the database, which the same guide recommends immediately above.

The [Store configuration in your database](https://docs.mattermost.com/administration-guide/configure/configuration-in-your-database.html) page already documents that `ClusterSettings.ReadOnlyConfig` is ignored when DB-backed configuration is enabled, because all nodes read the same row and there is no local `config.json` to diverge.

This PR scopes both mentions to the file-backed case:

- In [high-availability-cluster-based-deployment.rst](https://github.com/mattermost/docs/blob/master/source/administration-guide/scale/high-availability-cluster-based-deployment.rst), pulls `ReadOnlyConfig` out of the unconditional `mmctl` command list in step 7 and adds a conditional block explaining that it applies only when using `config.json`, with a cross-link to the DB configuration page.
- In [environment-configuration-settings.rst](https://github.com/mattermost/docs/blob/master/source/administration-guide/configure/environment-configuration-settings.rst), rewords the High Availability section preamble so the read-only System Console behavior is scoped to `config.json` deployments and notes that DB-backed deployments keep the System Console fully editable.

## Test plan

- [ ] Sphinx builds without warnings on the changed files
- [ ] Cross-reference to `configuration-in-your-database` resolves in both new locations
- [ ] Step 7 in the HA deployment procedure still reads as a coherent, ordered sequence when `ReadOnlyConfig` is unset (DB-backed path) and when it is set (file-backed path)


shiftZZnet (approved):
Approved, thanks for the change @roberson-io

